PR #28, opened 2014-12-18
Run dfu-convert.py with Python 2

dbrgn:
On Arch Linux, Python 3 is default.

An even better solution would be to support both Python 2 and 3 with the script. Maye I'll port it sometime (although there are probably quite some other scripts that are also Py2 only).


tobbeanton:
Thanks for the fix!

PR #37, opened 2015-01-09
HeadFree mode added

capriele:
I've implemented CareFree mode into de firmware... this is the code (stabilizer.c)

### DIFF
--- a/drivers/interface/motors.h
+++ b/drivers/interface/motors.h
@@ -39,36 +39,10 @@
 
 /******** Defines ********/
 
-/**
- * *VARNING* Flashing the brushless driver on the Crazyflie with normal brushed motors connected
- *  will turn it on at full speed when it is powered on!
- *
- * Generates a PWM wave (50 - 200 Hz update rate with 1-2 ms high pulse) using the timer. That way we can use the same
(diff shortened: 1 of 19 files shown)


tobbeanton:
I've tested the pull request and the CareFree mode works great. There are however some commented code, unused code which would be bad to integrate. Can you clean up the pull request or divide it up in several pull request?

I have a hard time understanding what this code does in the stabalizer@259:
       complementaryFilter(acc, gyro, &eulerRollDesired, &eulerPitchDesired);

capriele:
Yes I can. In the pull there is lot of unuseful code so i clean it.

ibrahima:
Actually, the first commit 466911f is sufficient for the CareFree mode, the rest is not really related or necessary and would probably be better as a separate commit, right? I'm running that on my CF 2.0 and it works much better than the default because of the yaw drift from the gyro. But the other commits won't merge cleanly on the crazyflie2 branch so it's probably better to isolate the CareFree mode from the other improvements.

Thanks for the commit! It's really helpful for me.

tobbeanton:
I integrated the Carefree mode code in https://github.com/bitcraze/crazyflie-firmware/commit/7cc43495522b55036631c8bc66b0033375f83816 commit. Thanks for the code!

PR #41, opened 2015-01-30
This commit adds support for logging RSSI samples

nordmoen:
After updating the firmware on the radio this commit adds support for
those messages to the STM32. The only things this commit does is add a
new constant to match what is sent from the radio and logging of those
samples.

Note: This pull request is intended to work with https://github.com/bitcraze/crazyflie2-nrf-firmware/pull/1


ataffanel:
Thanks, merged.

This triggered a bug that was fixed with commit 275b2f19f

PR #43, opened 2015-03-10
Add option to set device address.

whoenig:
This is part 3 of the effort of improving the multi-flie support by allowing custom device addresses.
Please see https://github.com/bitcraze/crazyflie-clients-python/pull/161 for the matching UI changes.
Please see https://github.com/bitcraze/crazyflie2-nrf-firmware/pull/2 for the matching NRF changes.

The changes include:
* Update eeprom configurationblock to include 5 additional bytes for the device address
* communicate the device address to NRF51 via syslink interface (requires updated nrf firmware)

### DIFF
--- a/config/config.h
+++ b/config/config.h
@@ -109,6 +109,7 @@
 //The radio channel. From 0 to 125
 #define RADIO_CHANNEL 80
 #define RADIO_DATARATE RADIO_RATE_250K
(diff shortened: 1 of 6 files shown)


tobbeanton:
Really appreciate you pull request, this is something we should have done a long time ago!

I will add a version conversion function so that saved settings will be moved to the new version.

PR #45, opened 2015-03-25
Fix for spinning out of control when yawing hard

tomfelker:
Hi!  I just recently got some Crazyflie 2's - they're very cool.  I have some crazy ideas about getting a bunch of them flying in formation, and some in-progress work on getting them to use the magnetometer to be able to maintain absolute yaw, but for now, I have a simple fix for an issue I saw.

Basically if you bump into something and it upsets its yaw enough that it tries to yaw back to the right with full authority, it will suddenly spiral out of control.  Turned out to be an issue with how it truncates the yaw to 16 bits, and there was also a related problem where it would update some PID loops too much.  More details in the commit notes.

There's also some unrelated one-liners that seem like a good idea.  I'm a little new to git, so please let me know if I should be branching or cherry picking or something when making pull request.


tobbeanton:
Thanks for the pull request! You have just fixed what we have called the "spin of death" bug, many thanks!

dbrgn:
Oh, I think  I've experienced this many times! Thanks, I'll try this right away :)

kent-williams:
Great Fix! Thanks

PR #47, opened 2015-04-05
Debug newlines

dbrgn:
Some newlines were missing in debug output, making a mess out of it.

Btw, some of the files (like `hal/src/imu_cf*`) have DOS line endings. Is that on purpose? If not, a conversion of all files would make sense. Right now git diff looks a bit messy with all those `^M` in them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ bin/*
 config.mk
 cflie.*
 version.c
+*.swp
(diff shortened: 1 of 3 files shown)


tobbeanton:
The DOS line endings shouldn't be there as it makes a lot more sense to be consistent. Thanks for the pull request!

PR #51, opened 2015-05-08
Make the LED-ring deck white spin effect really white, not green

yedlosh:
In both crazyflie firmware and python client this effect is named "white", while it actually uses green color. So I changed it to refer the whiteRing intead of greenRing color array.


tobbeanton:
I think it was white at some point... Thanks for the fix!

yedlosh:
Hey, you're welcome. If the neopixel_dev branch is ever gonna be merged into this one, I see now that Arnaud took care of it as well in https://github.com/bitcraze/crazyflie-firmware/commit/7ed4f890714b65d2206de4cf758d4c8965e483b2 ;)

PR #54, opened 2015-07-06
Morse "CF" LED pattern in "calibrated" mode

ian-ross:
Make red LED flash "CF" in Morse code instead of regular on/off in "calibrated" mode.

Kind of dumb, but kind of fun too...

### DIFF
--- a/hal/src/ledseq.c
+++ b/hal/src/ledseq.c
@@ -1,6 +1,6 @@
 /**
- *    ||          ____  _ __                           
- * +------+      / __ )(_) /_______________ _____  ___ 
+ *    ||          ____  _ __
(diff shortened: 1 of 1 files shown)


tobbeanton:
I really like the morse code idea but I'm not sure it works that well in this case. Too me it looks like the Crazyflie want's to tell me that something is wrong (and yes the red led is not good using at all but was inherited from Crazyflie 1.0. Something we should change). What about we make it a compile option so users can choose to use it if they want?

ian-ross:
It was just a silly thing I did when I was first playing with flashing new firmware to the CF.  That irregularly flashing red LED definitely does look like "distress" though...  I'll add a compile option and redo the PR.

PR #57, opened 2015-07-21
Deck analog API

fredgrat:
New deck_analog.[hc] API as discussed in thread https://forum.bitcraze.io/viewtopic.php?p=8305#p8305

Most notably (not yet discussed), is that the API is now using ADC2 instead of ADC1. This since there are internal temp and vref connected to ADC1, and ADC1 should perhaps thus be used in a different way at a later stage.

Also, some changes/cleanups since last posts, please see if these are acceptable.


tobbeanton:
Thanks for the pull request! Now we are one step closer to an compatible Arduino API.

PR #58, opened 2015-07-24
Proximity hardware abstraction layer

fredgrat:
The proximity is a hardware abstraction layer for proximity sensors.

* Provides a new "proximity" task for sampling from proximity sensors. Task/sampling frequency is configurable in proximity.h
* Proximity layer provides unprocessed samples from drivers, in addition to average values and median values.
* A sliding window of samples (size configurable in proximity.h) is used to calculate the average and median values.
* Proximity layer also reports the accuracy of the samples as / if reported by the sensor driver.
* Log group "proximity" available (disabled by default in proximity.h). See attached image showing the log group (data from the LV-MaxSonar-EZ4) in the plotter view of the cfclient, with unprocessed values (green), average values (blue) and median values (pink).
* Also provided is a "maxsonar" driver for LV-MaxSensor-EZ4 (MB1040). Driver is prepared for adding more MaxSonar models.
* Both the maxsonar driver and the proximity abstraction layer are disabled by default in their respective header files. (updated: No longer in config.h)
* Initial testing with the LV-MaxSonar-EZ4 sensor seems to yield good results with 20Hz sampling, a sliding window size of 9, and reading median values. See attached image.

The first screenshot (with almost flat median values) shows the sensor with a constant distance from a wall. The second screen shot (showing step changes to the median values) shows the values when first introducing and subsequently removing an obstacle between the sensor and the wall.

screenshot - proximity and maxsonar ez4

screenshot - proximity and maxsonar ez4 - obstacle

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -124,7 +124,7 @@ PROJ_OBJ_CF2 = platform_cf2.o
 # Drivers
 PROJ_OBJ += exti.o nvic.o
 PROJ_OBJ_CF1 += led_f103.o i2cdev_f103.o i2croutines.o adc_f103.o mpu6050.o motors_f103.o hmc5883l.o ms5611.o nrf24l01.o eeprom.o
-PROJ_OBJ_CF2 += led_f405.o mpu6500.o motors_f405.o i2cdev_f405.o ws2812.o lps25h.o ak8963.o eeprom.o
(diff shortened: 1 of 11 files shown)


tobbeanton:
Thanks for this pull request. I found that it didn't build for Crazyflie 1.0 so I fixed that by having a ifdef around poximityInit.

fredgrat:
OK, thanks for the CF1 build fix!

PR #59, opened 2015-07-27
Situation Awareness framework

fredgrat:
This pull request includes the following features (all disabled by default):

* 'Free fall' detection, and enabling of AltHold when detected. This means you can throw the crazyflie into the air, and it starts hovering (as long as you don't throw it 'Tumbled', see below).

* 'Tumbled' detection, and disabling of thrust when detected. When the crazyflie is tumbling and subsequently crashing into walls or the ground, an active altHoldMode causes the propellers to run at maximum or high thrust as the crazyflie hits the floor upside down or sideways. This has broken several of my propellers during altHoldMode testing. Allowing the 'Tumbled' detection to kill the thrust to the motors have saved me many broken propellers already...and the crashes become less violent.

* 'At Rest' detection. Handy for Free Fall and altHold testing, just hold the crazyflie down on the table (to do altHold testing/measurements without flying), and it will disable altHoldMode after a couple of seconds.

These features can be enabled by uncommenting SITAW_ENABLED in sitaw.h

The implementation consists of:

* trigger.[hc] - a framework for watching variables and release triggers when said variables has stayed within a threshold for some time. Both the threshold and the time is configurable.

* sitaw.[hc] - a Situation Awareness framework, which uses the trigger framework to learn about its situation. Currently it understands 'Free Fall', 'Tumbled' and 'At Rest'. It should be fairly easy to add more.

* Integration in stabilizer.c, by creating appropriate CallOut functions. Please review and let me know if this is a possible way to tie into the stabilizer loop, without cluttering the code in the stabilizer loop. If you have another preferred approach, please let me know and I will update my patch accordingly.

* Some integration in commander.c, to be able to enable / disable the altHoldMode

The MPU contains some detection functionality in hardware, but there are a few reasons I preferred implementing this in software:

1) We have available CPU cycles in the crazyflie 2.0
2) Easier to integrate with existing calculated variables in stabilizer.c / commander.c
3) Since the implementation is now hardware agnostic, it should live longer across crazyflie hardware generations.
4) Much easier to extend, can use any variable in the firmware to detect situations.

This pull request could be the basis for further extensions of the framework in the future:

* Add trigger chains, making it possible to detect a series of events for more complex situations.
* Add an actions framework, and map detected situations (such as 'Free Fall' etc.) to actions.


tobbeanton:
This is a great start to some situation awareness. We think you are writing great code and well documented. Much better then we do so now we have to raise the level :-).

There are some functional aspects that needs to be thought through. Like how to get out off free fall recovery mode. Just by detecting manual throttle again or with the altHold configured button? This is also more difficult on mobile devices as they yet don't have the altHold button so in that sense manual throttle would be better. Also there might be other situations this functionality can be interfering?

Anyway these are minor things, and since the functionality have to be enabled it is good to have the code in the master repo for others to test and play with. Thanks again!

fredgrat:
Hi thanks for the positive feedback, and thanks for merging my pull request(s).

Wrt. to the functional aspects, I have used the altHold configured button to disable altHoldMode, but as I also use the android client some time, I know this will not work for that client.

But I am working towards "flight-assistant" features that will make the crazyflie easier to operate and fly, and that should also include mobile devices. Perhaps I should open a forum discussion on this, to discuss how this should be done...

Basically I believe getting the AltHold functionality to work in a robust manner is crucial to such "flight-assistant" features, I am currently trying to get a better understanding of how it works.

Issue #61, opened 2015-08-01
Sending bad log configuration can crash tasks, smash stack


cbiffle:
Found in firmware commit b7154ae924505f2d24cca38ee8f9cd231abb03bb.

# Steps to reproduce

1. Set up Crazyflie with released version of Python client, connected over wireless or USB (doesn't matter).
2. Create a log configuration containing a large-ish amount of data, e.g. `motor.m1` - `motor.m4` and `stabilizer.{pitch,roll,thrust,yaw}`.
3. Activate it in the Plotter tab.

# Expected results

The data I asked for shows up (best case).

An error is reported (worst case).

# Actual results

The Crazyflie's failure lights come on, signalling that something onboard has failed.  All air-to-ground telemetry halts, but ground-to-air control authority is retained.

# Diagnosis

If the client sends a sequence of log ops that will construct an overlong packet, the firmware happily follows the instructions.  This ripples out in two directions:

- It eventually kills off the crtpTxTask,which hits an `ASSERT` (`hal/src/radiolink.c` line 159).
- It overruns the buffer in the worker task, smashing its stack and leading to undefined further behavior.

tobbeanton:
Great find and fix cbiffle!

PR #62, opened 2015-08-01
Avoid smashing the stack when generating log packets

cbiffle:
This is one possible approach to fixing #61.  This branch includes two commits; the first is a diagnostic aid to make it easier to identify who's generating bogus CRTP packets.


cbiffle:
Ignore the second batch of commits, that was a rebase mistake. ;-)

tobbeanton:
Great findings cbiggle and thanks for the pull request!

Issue #63, opened 2015-08-02
Race condition prevents STM from coming back up when connection is up


cbiffle:
Found in b7154ae924505f2d24cca38ee8f9cd231abb03bb.

# Steps to reproduce

1. Power on CrazyFlie (tested with CF2 but applies to both).
2. Connect over Nordic.
3. Reset CrazyFlie STM32 CPU.  Easiest way: use button on debug breakout.

# Expected results

CrazyFlie STM32 comes back up.  Maybe connection drops, worst case.

# Actual results

STM32 hits assert, cannot reestablish connection, must be power-cycled.

# Diagnosis

There's a race condition in queue creation in `radiolinkInit`.  The `syslink` task is started before the queues it uses (particularly `crtpPacketDelivery`) are created.  Since the `syslink` task is higher-priority than the task that creates it (`system`), it immediately preempts.

Normally, no data is available at the nRF UART, and it blocks and returns to sleep before touching the CRTP queues.

However, if the connection is up, the nRF may have a message ready -- in which case `syslink` tries to process it before initialization has completed.

Because the task that does most of the initialization (`system`) is so low-priority, there are probably other bugs of this same shape lurking.

tobbeanton:
Thanks for the fix cbiffle!

PR #64, opened 2015-08-02
radiolink: create queue before starting syslink.

cbiffle:
This is a mitigation measure working around issue #63.

A more systematic fix might be to disable context switches during initialization, but I don't have the sort of global understanding of this codebase to do that.


tobbeanton:
I've seen this problem as well and as you say this is a first step of fixing it. The way the nRF51 and the STM32 are interacting during different resets have not really been thought trough.

PR #65, opened 2015-08-03
Assorted FPU fixes

cbiffle:
These commits...

- Make the `M2T`/`F2T` macros safer.
- Reduce image size by ~8%.
- Remove uses of `errno` and the associated `TODO`s.
- Should significantly increase math performance, though I haven't quantified that.

I've allowed a few lines over 80 characters in these commits, because I couldn't figure out what margin the existing code was using (some files have lines well over 100 characters).  I'm happy to wrap the lines to the margin of your choice.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -188,7 +188,8 @@ INCLUDES_CF2 += -I$(STLIB)/STM32_USB_OTG_Driver/inc
 INCLUDES_CF2 += -Ideck/interface
 
 ifeq ($(USE_FPU), 1)
-PROCESSOR = -mcpu=cortex-m4 -mthumb -mfloat-abi=hard -mfpu=fpv4-sp-d16
+	PROCESSOR = -mcpu=cortex-m4 -mthumb -mfloat-abi=hard -mfpu=fpv4-sp-d16
+	CFLAGS += -fno-math-errno
 else
 	ifeq ($(PLATFORM), CF1)
 		PROCESSOR = -mcpu=cortex-m3 -mthumb
--- a/config/FreeRTOSConfig.h
+++ b/config/FreeRTOSConfig.h
@@ -114,8 +114,8 @@ to exclude the API function. */
 #define vPortSVCHandler SVC_Handler
 
 //Milliseconds to OS Ticks
-#define M2T(X) ((unsigned int)(X*(configTICK_RATE_HZ/1000.0)))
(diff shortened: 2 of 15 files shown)


Makefile
```diff
@@ -188,7 +188,8 @@ INCLUDES_CF2 += -I$(STLIB)/STM32_USB_OTG_Driver/inc
 INCLUDES_CF2 += -Ideck/interface
 
 ifeq ($(USE_FPU), 1)
-PROCESSOR = -mcpu=cortex-m4 -mthumb -mfloat-abi=hard -mfpu=fpv4-sp-d16
+	PROCESSOR = -mcpu=cortex-m4 -mthumb -mfloat-abi=hard -mfpu=fpv4-sp-d16
+	CFLAGS += -fno-math-errno
```
cbiffle: FYI, this must go in `CFLAGS` and not `PROCESSOR` because you pass the contents of `PROCESSOR` to `as`, which interprets `-f` flags rather differently.

config/FreeRTOSConfig.h
```diff
@@ -114,8 +114,8 @@ to exclude the API function. */
 #define vPortSVCHandler SVC_Handler
 
 //Milliseconds to OS Ticks
-#define M2T(X) ((unsigned int)(X*(configTICK_RATE_HZ/1000.0)))
-#define F2T(X) ((unsigned int)((configTICK_RATE_HZ/X)))
+#define M2T(X) ((unsigned int)((X)*(configTICK_RATE_HZ/1000.0)))
```
cbiffle: Interestingly, the optimizer falls over if you replace `1000.0` here with `1000.0f`.  I haven't investigated further.

modules/src/sensfusion6.c
```diff
@@ -29,6 +29,8 @@
 #include "imu.h"
 #include "param.h"
 
+#define M_PI_F ((float) M_PI)
```
cbiffle: This might be a valuable macro to centralize, but I wasn't sure where you'd want it placed.  If you have suggestions....

tobbeanton:
Nice optimizations!

As for the line wrapping we don't have a hard margin. Making it easily readable is what we are aiming for. The lines shouldn't be much longer then 100 chars though as that tend to be less readable.

Thanks for the pull request!

PR #66, opened 2015-08-04
maxsonar: fix apparent unit errors.

cbiffle:
The MB1040 accuracy function was taking a measurement in millimeters,
multiplying it by 25.4, and comparing it with inches.  This appears to
have been a simple inversion of the desired math, which I've fixed.

I've also switched it to single-precision arithmetic; this escaped my
notice in my earlier batch of fixes because it's not compiled by
default.

Question: how do y'all test all these optionally-compiled-in features?  It seems like you've got a cartesian explosion of build types, but your Makefile only builds one at a time....

### DIFF
--- a/drivers/src/maxsonar.c
+++ b/drivers/src/maxsonar.c
@@ -31,6 +31,8 @@
 #include "maxsonar.h"
 #include "deck.h"
 
+#define IN2MM(x) ((x) * 25.4f)
+
 /* Internal tracking of last measured distance. */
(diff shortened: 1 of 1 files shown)


tobbeanton:
As for the optional compile-switched functionality we don't require them to be tested. However the basic functionality should build for both CF1 and CF2.

There is an idea to have most of the optional functionality being built and dynamically enabled by the deck subsystem during run time but that is far ahead...

PR #68, opened 2015-08-09
Adding automatic take-off function based on the sitaw framework

fredgrat:
This pull request adds an automatic take-off function based on the situation awareness framework. If pressing the altHold button while the crazyflie is at rest (and not tumbled), the automatic take-off function will activate.

The function deactivates automatically once the take-off has been conducted, and remains in altHold mode thereafter (as long as the user keeps the altHold button pressed).

Note: This function is disabled by default. Uncomment the #define SITAW_ENABLED in sitaw.h to enable this feature.


tobbeanton:
Works like a charm. To bad the altitude hold isn't that accurate though. The pressure accuracy seam to be a bottle neck but not much work has been done optimizing it so there might still be room for improvement.

Thanks for the pull request!

Issue #78, opened 2015-10-26
__packed undeclared with arm-none-eabi-gcc 4.7.4


erwincoumans:
Using arm-none-eabi-gcc compilation breaks at the use of __packed.
If the __packed is removed all compiles/links fine. Is there any replacement for __packed?

Version of arm-none-eabi-gcc is:
[~/develop/crazyflie-firmware]$ arm-none-eabi-gcc --version
arm-none-eabi-gcc (GNU Tools for ARM Embedded Processors) 4.7.4 20140401 (release) [ARM/embedded-4_7-branch revision 209195]

cstanke:
Try the build with 4.9-2014q4. I think that's the required version since the CF1 and CF2 firmware merge.

ataffanel:
__packed is defined in cdefs.h file by gcc. Maybe your GCC does not have it.

You can try to change the line https://github.com/bitcraze/crazyflie-firmware/blob/master/hal/interface/usb_conf.h#L258 into:
``` c
  #ifndef __packed
    #define __packed    __attribute__ ((__packed__))
  #endif
```
If it works I can push it.

We actually use either the latest Ubuntu package arm-none-eabi-gcc or gcc from the GCC ARM Embedded from launchpad.

erwincoumans:
@ataffanel Yes, that does the job, see pull request #79

ataffanel:
It is merged, thanks

PR #79, opened 2015-10-27
fix to compile using older compiler, fixes Issue 78

erwincoumans:
https://github.com/bitcraze/crazyflie-firmware/issues/78


ataffanel:
Thanks, merged.

PR #83, opened 2015-12-10
Log motor PWM ratios after voltage compensation

### LINKED ISSUE #77: Add log for real motors PWM
Currently the motor PWM ratio logging is done before battery voltage compensation. It would be useful to add log variables after the battery voltage compensation to get the PWM ratio actually applied to the motors.

ograff:
Update motors.c to log PWM ratios after voltage compensation. Fix #77

### DIFF
--- a/drivers/src/motors.c
+++ b/drivers/src/motors.c
@@ -37,10 +37,16 @@
 //FreeRTOS includes
 #include "task.h"
 
+//Logging includes
+#include "log.h"
+
 static uint16_t motorsBLConvBitsTo16(uint16_t bits);
 static uint16_t motorsBLConv16ToBits(uint16_t bits);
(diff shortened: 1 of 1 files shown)


ograff:
D'oh, looks like this overflows RAM. I'll see if I can come up with a way to grab this info from where it already exists, instead of storing it twice

ataffanel:
Yes the Crazyflie 1 is pretty tight in RAM, I am going to look if I can free some ram somewhere else as well.

ograff:
Hmm, I can't seem to figure out how I could coerce the data from the registers into the Log group in any useful way. Perhaps reducing other RAM usage is the best option.

ataffanel:
I freed some RAM by moving the led sequence in flash, so now it should fit in CF1.

Thanks!

Issue #80, opened 2015-10-27
The code does not build cleanly with some defines enabled


krichardsson:
I have added builds to travis where "all" defines are turned on to build code that is normally not built. Unfortunately the code base does not build cleanly with all of them, and we need to fix this.

To build with some extra defines, use 
make "EXTRA_CFLAGS=-DDEF_1 -DDEF_2"

The defines that do not work are.

CF1
~~-DUSE_ESKYLINK~~
~~-DCALIBRATED_LED_MORSE~~
~~-DUART_OUTPUT_TRACE_DATA~~

CF2
-DOW_WRITE_TEST
-DOW_READ_TEST
-DPM_IIR_FILTER

When fixed, add the (remaining) defines to the builds in travis.yml

PR #91, opened 2016-02-12
Fix Eskylink compilation option

ograff:
Added Eskylink.o to list of files to be compiled, included errno.h in
eskylink.c

Addresses #80


ataffanel:
Thanks! it is merged.

PR #93, opened 2016-02-18
set correct vid/pid for DWM1000 deck

whoenig:


ataffanel:
Thanks! I forgot to fix that.

PR #104, opened 2016-03-17
Added a snippet from Richard Wagner 'Ride of the Valkyries'

cstanke:
Had to be done. Wish I was a better musician and could program more of the song.


krichardsson:
Thanks!

PR #107, opened 2016-03-24
revive microsecond timer

jpreiss:
This change re-enables the microsecond timer function `uint64_t usecTimestamp()` for CF2. The the timer is changed from TIM1 to TIM7 which is the most minimal-featured timer on the STM32. The  prescaler is now determined from the `SystemCoreClock` variable instead of a hard-coded constant.


ataffanel:
Thanks, it is merged.

PR #108, opened 2016-03-30
Add gdb target for make

whoenig:
This is similar to what you have for the nrf-firmware and simplifies command-line debugging by just using "make gdb". Tested with a J-Link, but should work for STLink as well.


krichardsson:
Thanks!

PR #111, opened 2016-05-01
Add biq quad deck CFLAGS example

fredg02:


ataffanel:
Thanks Fred! It is merged.

PR #114, opened 2016-05-21
Allow custom arguments for the cload script.

whoenig:
This allows to reflash the Crazyflie without the need to put it manually in bootloader mode.


krichardsson:
Merged
Thanks!

PR #116, opened 2016-05-27
fix small spelling issues in commander

whoenig:
This also unifies the coding style (usage of brackets) in this file.


ataffanel:
Thanks, I merge it!

PR #117, opened 2016-05-30
Crazyflie2

Mitteau:
Hi
It would be excellent to install a parameter giving an initial value to thrust when putting on althold mode.
Thanks
Jean-Claude

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ bin/*
 config.mk
 cflie.*
 version.c
+*.swp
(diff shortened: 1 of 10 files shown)


ataffanel:
Hi,
This is already implemented. The param is called ```posCtlPid.thrustBase```. The default value is 36000 but you can change it from the ground depending of the weight of your copter.

About this pull request, it does not contain any code related to the altitude-hold so I assume it was a mistake and I close it.

Mitteau:
Merci
Bye
Jean-Claude

PR #118, opened 2016-06-01
Fixed I2C contention & timeout bugs

mikehamer:
This pull request fixes two related bugs, which resulted in high CPU load from a high-priority task causing I2C timeouts, which led to tasks spinning endlessly in the (now removed) read/write spin-locks, which finally resulted in an assertFail on reenabling a FreeRTOS timer. This issue was exacerbated by the recent size increase in gyro bias calibration (more calculations from a high priority task), particularly if a radio connection is also established during the calibration routine.

The bugs fixed by this pull request are:

1. Improper semaphore usage:
 * Previously the I2C semaphores were initialised as unlocked (the subsequent xSemaphoreTake in the initialization was forgotten), and as such that they could be taken, before first being given.
 * This caused the (intended-to-be-blocking) xSemaphoreTakes to always pass immediately, which is the reason for the spin-locks (since the code did not wait for the completion interrupts).
 * This commit changes from the deprecated vSemaphoreCreateBinary (which creates the semaphore in an unlocked state, and would require a subsequent xSemaphoreTake) to the more current xSemaphoreCreateBinary (which creates the semaphore in the locked state).

2. I2C Timeouts:
 * The calls to CPAL_I2C_Read and CPAL_I2C_Write contain a spin-lock to detect timeout.
 * If a context change occurs during this function (ie. if the task is interrupted by a higher priority task), then I2C will timeout.
 * This pull request disables context switching for the duration of these calls.

### DIFF
--- a/src/drivers/src/i2cdev_f405.c
+++ b/src/drivers/src/i2cdev_f405.c
@@ -71,7 +71,7 @@ static inline void i2cdevRuffLoopDelay(uint32_t us);
 
 #define SEMAPHORE_TIMEOUT M2T(30)
 static void semaphoreGiveFromISR(xSemaphoreHandle semaphore);
-static void i2cDevTakeSemaphore(CPAL_DevTypeDef CPAL_Dev);
+static BaseType_t i2cDevTakeSemaphore(CPAL_DevTypeDef CPAL_Dev);
 static void i2cDevGiveSemaphore(CPAL_DevTypeDef CPAL_Dev);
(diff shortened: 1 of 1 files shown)


mikehamer:
This should fix your problem, @ataffanel. Note that I only fixed this yesterday and thus it has only been running for a few hours, more testing is probably required.

ataffanel:
Thanks! I will merge it and so we have much more tester! :) This bug has be haunting us for a while so it is really exciting to see a potential fix.

PR #120, opened 2016-06-15
correct maximum payload for logging size

whoenig:
The logging packet first sends block id (1 byte) and timestamp (3 bytes) followed by the actual values. Hence, only 26 (out of the 30 CRTP payload) can be used.


ataffanel:
Thanks! I merged it.

PR #121, opened 2016-06-15
Fixed syslink unhandled overrun bug

mikehamer:
This commit fixes a bug, whereby an overrun of the syslink UART buffer was not handled in the interrupt routine, causing constant retriggering and reentry, stalling the system.

Overruns are caused when high CPU/interrupt load causes the syslink UART data buffer to not be processed fast enough.

This bug was fixed by:
 1. handling unhandled conditions in the ISR, and
 2. reordering/simplifying instructions in the ISR to reduce latency before data read.

As it is currently implemented, every received byte causes the CPU to be interrupted, and the byte to be sent to the syslink for processing (this is due to a variable packet size and hence unknown transaction length).

TODO: A nicer implementation would be to fix the packet length at its maximum (32 bytes), and use DMA for the entire RX transaction. This would entirely avert the problem of an overrun causing packet loss. I assert (without evidence) that the time required to interrupt the CPU every byte, is less than the overhead due to fixing the packet size at its maximum length.

### DIFF
--- a/src/drivers/src/uart_syslink.c
+++ b/src/drivers/src/uart_syslink.c
@@ -23,6 +23,7 @@
  *
  * uart_syslink.c - Uart syslink to nRF51 and raw access functions
  */
+#include <stdint.h>
 #include <string.h>
 
 /*ST includes */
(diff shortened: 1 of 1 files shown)


ataffanel:
Thanks a lot for the fix. I agree, sending fixed sized packet would make things easier and I actually though about doing it from the beginning but back then it would have been over-optimisation.

It is possible to switch to fixed sized packet right now without changing the protocol: we just have to pad the packets with zeros until we reach the wanted size. Then on the receiving side we can "synch" the DMAs (ie. at the beginning we might end up having one packet in the middle of a transaction, this can be fixed for the following transactions).

PR #122, opened 2016-06-15
Fixed crtpRxTask bug on nopLink

mikehamer:
This pull request fixes a bug, whereby the crtpRxTask would loop infinitely (and without delay) if a communication link had not been established (ie, link == nopLink).


ataffanel:
Thanks, I merged it.

PR #123, opened 2016-06-15
Fixed I2C delay bug

mikehamer:
This delay bug was two-fold:

1. Wrong loop delay
 - The loop delay was (I assume, based on a comment in `i2cdev.h`) set for the CF1, and not updated for the CF2. The CF2 runs these loops ~3.5x faster, so the delays were much shorter than expected. I'm not sure if this was a problem, but things seem to be more robust after this change.
 - The new loop speeds were timed using the DWT->CYCCNT to count clock cycles.

2. Overflow caused hangs
 - For some reason, it occasionally occurred that the delay variable would (due to it being a uint32_t and being decremented) underflow, wrapping to the maximum value.
 - The loop was changed to increment the value and terminate on all values greater than the target, rather than only on 0.

### DIFF
--- a/src/drivers/interface/i2cdev.h
+++ b/src/drivers/interface/i2cdev.h
@@ -33,9 +33,17 @@
 
 #include "cpal.h"
 
+#ifdef PLATFORM_CF1
 // Delay is approx 0.2us per loop @64Mhz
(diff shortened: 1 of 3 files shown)


ataffanel:
Looks good and tested. I merge it. Thanks.

PR #124, opened 2016-06-15
Enables all EXTI handlers to use callbacks

mikehamer:
This commit centrally enables all EXTI interrupts on the NVIC. Note that the individual EXTI line interrupts still need to be enabled for interrupts to be generated.

This fixes issues related to EXTI9_5 and EXTI15_10 handlers being initialized and defined multiple times.

This affects, primarily, the IMU interrupt (EXTI13) and interrupts on some of the deck pins (eg. dw1000 on EXTI11). The solution was therefore made global, and existing interrupts updated to use callbacks, rather than the handlers directly.

### DIFF
--- a/src/config/nvicconf.h
+++ b/src/config/nvicconf.h
@@ -84,5 +84,15 @@
 #define NVIC_RADIO_PRI        11
 #define NVIC_ADC_PRI          12
 #define NVIC_CPPM_PRI         14
+#define NVIC_SYSLINK_PRI      5
+
+// Priorities for external interrupts
+#define EXTI0_PRI NVIC_LOW_PRI
+#define EXTI1_PRI NVIC_LOW_PRI
+#define EXTI2_PRI NVIC_LOW_PRI
+#define EXTI3_PRI NVIC_LOW_PRI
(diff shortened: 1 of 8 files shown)


mikehamer:
The Travis CI build passed all CF2 tests, and I tested locally on my CF2 to ensure that performance was maintained.

It seems, however, that the Travis CI build failed on a CF1 check involving the line:

    RCC_AHB2PeriphClockCmd(RCC_APB2Periph_SYSCFG, ENABLE);

My guess is that the SYSCFG is on a different bus on the CF1 micro.

mikehamer:
The latest update compiles for the CF1, but has not been tested (I don't own a CF1).

ataffanel:
This looks good and is tested with my CF2. I will test on a CF1 tomorrow to make sure.

ataffanel:
It crashes the CF1. I am fixing it and then I will merge the PR (I wonder if I can add to the PR or if I need to first merge and then apply change .... Maybe I can PR to your branch :smile: ).

ataffanel:
Tested after the CF1 fix, merged, thanks!

PR #125, opened 2016-06-15
Adding interrupt-driven sensors task

mikehamer:
This pull request introduces an interrupt-driven task to handle IMU reads using DMA.

The task configures the IMU's master mode to handle barometer and magnetometer access.

The task waits on an IMU Data Ready interrupt on EXTI13 before initializing a DMA read (via CPAL) of the IMU's registers.

Since the IMU handles barometer and magnetometer access, all sensors can be read in a single DMA transaction. This eliminates I2C contention problems, since only one transaction occurs on the bus (rather than separately reading from 3 sensors on one bus).

This significantly improves performance of the sensors. For example, sensors_stock.c takes 700us to read the gyro and accelerometer data (most of the time spent waiting), and thus severely limits the rate of the stabilizer task. This was previously not an issue since the stock stabilizer is computationally light, but becomes important once more advanced onboard estimators or controllers are added to the mix.

**IMPORTANT NOTE:** As of this commit, the magnetometer only updates once, and its values stay constant in the IMU registers. I don't know why, but I also don't know anyone who uses a magnetometer for indoor flight... so this is a low priority for me to fix. I have disabled the magnetometer for the time being.


mikehamer:
Note that this pull request is a little dependent on the EXTI callbacks pull request.

If you decide to accept this pull request, without accepting the pull request for EXTI callbacks, you will need to initialize the EXTI15_11 IRQ on the NVIC  and replace the EXTI13_Callback in sensors_task.c with EXTI15_11_IRQHandler and do appropriate line checks.

mikehamer:
All CF2 tests passed, and I flew with my CF2 to ensure that performance was maintained.

The current implementation of `sensors_task.c` does not support the CF1 (IMU initialization is hard coded for the CF2 IMU and sensors). A compile error will be generated in this case, informing the user.

However this compile error (`sensors_task.c:51`) was generated below `#include <stm32f4xx.h>`, which obviously can't be found for the CF1. Hence the Travis CI build failed with this error, rather than something more informative.

I suggest moving the check to the top of the `sensors_task.c` file, such that the error is informative.

mikehamer:
I have a few updates and fixes to this pull request, which I will resubmit in a few days.

PR #126, opened 2016-06-20
Fixed semaphore logic in uart_syslink.c

mikehamer:
This should fix two issues:

1. Sometimes the queues and semaphores weren't initialized  before the first interrupt was triggered, and the interrupt tried to push data to a null queue. This has only occurred to me once, and on a warm restart, I'm not sure if it affects cold boots.

2. The semaphore logic seemed incorrect (eg. uartslkSendDataIsrBlocking wouldn't have blocked). I have now made the logic explicit using two semaphores. It might be overkill since I think there is only one task that calls these functions. But it solves the problem.

Note that this only affects the CF2 (since the file isn't included for the CF1), and I have been flying successfully with the change for the last few days.

### DIFF
--- a/src/drivers/src/uart_syslink.c
+++ b/src/drivers/src/uart_syslink.c
@@ -50,7 +50,8 @@
 
 static bool isInit = false;
 
-xSemaphoreHandle waitUntilSendDone = NULL;
+static xSemaphoreHandle waitUntilSendDone;
+static xSemaphoreHandle uartBusy;
 static xQueueHandle uartslkDataDelivery;
 
 static uint8_t dmaBuffer[64];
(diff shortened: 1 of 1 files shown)


mikehamer:
The tests failed due to:

    Pulling repository docker.io/bitcraze/builder
    Get https://registry-1.docker.io/v1/repositories/bitcraze/builder/tags/latest:
    dial tcp 54.175.37.127:443: i/o timeout

Which is not an issue with the PR. I have compiled this for CF1 and CF2, and flight tested on CF2.

ataffanel:
I am relaunching the build, this should only be transient.

ataffanel:
This looks great! It seems the solve the occasional crash upon connections. Thanks, it is merged.

PR #130, opened 2016-06-24
Added gtgps, compass, & supporting changes

jackemoore:
This update fails in the GitHub master  cf-fw with a watchdog timeout,
but works successfully in updates prior and up to 2016.05.19

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -100,7 +100,7 @@ ST_OBJ_CF2 += usb_core.o usb_dcd_int.o usb_dcd.o
 ST_OBJ_CF2 += usbd_ioreq.o usbd_req.o usbd_core.o
 
 # libdw dw1000 driver
-#VPATH_CF2 += vendor/libdw1000/src
+VPATH_CF2 += vendor/libdw1000/src
 
 # FreeRTOS
 VPATH += $(PORT)
@@ -149,7 +149,7 @@ PROJ_OBJ_CF1 += imu_cf1.o pm_f103.o nrf24link.o ow_none.o uart.o
 PROJ_OBJ_CF2 += imu_cf2.o pm_f405.o syslink.o radiolink.o ow_syslink.o proximity.o usec_time.o
 
 # libdw
-#PROJ_OBJ_CF2 += libdw1000.o libdw1000Spi.o
+PROJ_OBJ_CF2 += libdw1000.o libdw1000Spi.o
 
 # Modules
 PROJ_OBJ += system.o comm.o console.o pid.o crtpservice.o param.o mem.o
@@ -160,7 +160,7 @@ PROJ_OBJ_CF2 += platformservice.o sound_cf2.o extrx.o
 # Stabilizer modules
 PROJ_OBJ += commander.o attitude_pid_controller.o sensfusion6.o stabilizer.o
 PROJ_OBJ += position_estimator_altitude.o position_controller_pid.o
-PROJ_OBJ += estimator_$(ESTIMATOR).o controller_$(CONTROLLER).o compass.o
+PROJ_OBJ += estimator_$(ESTIMATOR).o controller_$(CONTROLLER).o
 PROJ_OBJ += sensors_$(SENSORS).o power_distribution_$(POWER_DISTRIBUTION).o
 
 
@@ -178,7 +178,7 @@ PROJ_OBJ_CF2 += bigquad.o
 PROJ_OBJ_CF2 += ledring12.o
 PROJ_OBJ_CF2 += buzzdeck.o
 PROJ_OBJ_CF2 += gtgps.o
-@PROJ_OBJ_CF2 += dwm1000.o
+PROJ_OBJ_CF2 += dwm1000.o
 PROJ_OBJ_CF2 += cppmdeck.o
 #Deck tests
 PROJ_OBJ_CF2 += exptest.o
--- a/Makefile
+++ b/Makefile
@@ -100,7 +100,7 @@ ST_OBJ_CF2 += usb_core.o usb_dcd_int.o usb_dcd.o
 ST_OBJ_CF2 += usbd_ioreq.o usbd_req.o usbd_core.o
 
 # libdw dw1000 driver
-VPATH_CF2 += vendor/libdw1000/src
+#VPATH_CF2 += vendor/libdw1000/src
 
 # FreeRTOS
 VPATH += $(PORT)
@@ -149,7 +149,7 @@ PROJ_OBJ_CF1 += imu_cf1.o pm_f103.o nrf24link.o ow_none.o uart.o
 PROJ_OBJ_CF2 += imu_cf2.o pm_f405.o syslink.o radiolink.o ow_syslink.o proximity.o usec_time.o
 
 # libdw
-PROJ_OBJ_CF2 += libdw1000.o libdw1000Spi.o
+#PROJ_OBJ_CF2 += libdw1000.o libdw1000Spi.o
 
 # Modules
 PROJ_OBJ += system.o comm.o console.o pid.o crtpservice.o param.o mem.o
@@ -160,7 +160,7 @@ PROJ_OBJ_CF2 += platformservice.o sound_cf2.o extrx.o
 # Stabilizer modules
 PROJ_OBJ += commander.o attitude_pid_controller.o sensfusion6.o stabilizer.o
 PROJ_OBJ += position_estimator_altitude.o position_controller_pid.o
-PROJ_OBJ += estimator_$(ESTIMATOR).o controller_$(CONTROLLER).o
+PROJ_OBJ += estimator_$(ESTIMATOR).o controller_$(CONTROLLER).o compass.o
 PROJ_OBJ += sensors_$(SENSORS).o power_distribution_$(POWER_DISTRIBUTION).o
 
 
@@ -178,7 +178,7 @@ PROJ_OBJ_CF2 += bigquad.o
 PROJ_OBJ_CF2 += ledring12.o
 PROJ_OBJ_CF2 += buzzdeck.o
 PROJ_OBJ_CF2 += gtgps.o
-PROJ_OBJ_CF2 += dwm1000.o
+@PROJ_OBJ_CF2 += dwm1000.o
 PROJ_OBJ_CF2 += cppmdeck.o
 #Deck tests
 PROJ_OBJ_CF2 += exptest.o
--- a/Makefile
+++ b/Makefile
@@ -159,8 +159,8 @@ PROJ_OBJ_CF2 += platformservice.o sound_cf2.o extrx.o
 
 # Stabilizer modules
 PROJ_OBJ += commander.o attitude_pid_controller.o sensfusion6.o stabilizer.o
-PROJ_OBJ += position_estimator_altitude.o position_controller_pid.o
-PROJ_OBJ += estimator_$(ESTIMATOR).o controller_$(CONTROLLER).o
+PROJ_OBJ += position_estimator_altitude.o position_controller_pid.o 
+PROJ_OBJ += estimator_$(ESTIMATOR).o controller_$(CONTROLLER).o compass.o
 PROJ_OBJ += sensors_$(SENSORS).o power_distribution_$(POWER_DISTRIBUTION).o
 
 
--- a/src/deck/drivers/interface/gtgps.h
+++ b/src/deck/drivers/interface/gtgps.h
@@ -1,6 +1,6 @@
 /**
- *    ||          ____  _ __                           
- * +------+      / __ )(_) /_______________ _____  ___ 
+ *    ||          ____  _ __
+ * +------+      / __ )(_) /_______________ _____  ___
  * | 0xBC |     / __  / / __/ ___/ ___/ __ `/_  / / _ \
  * +------+    / /_/ / / /_/ /__/ /  / /_/ / / /_/  __/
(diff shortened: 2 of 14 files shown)


ataffanel:
Hi,
It is apparently missing the file compass.c. Also could you precise the commit at which it stops working (ie. with a commit hash)?

jackemoore:
Hi,
I made a second commit to @130 that includes the file compass.c.  As noted in the Bitcraze Forum post Crazyflie 2 with GPS, I prepared a test build without gtgps and only included compass.h compass.c and the change to stabilizer.c to call the compassController.  The test build failed in the same manner as before.  I’ll look at how to do a commit hash, which at the moment is “greek” to me.

ataffanel:
Thanks for the update. The commit hash is simply the hash Git is showing for a commit, you can usually copy only the 6-10 first characters, for example 3c2080d11b8e3 is the last commit to date. If you just copy that instead of "2016.05.19" it makes it much easier and clearer to find the commit.

Apparently the branch is still failing to compile for CF1. 

Also a quick look shows that you disabled the dwm1000 files, they should be left untouched for me to be able to merge the PR.

jackemoore:
Hi,
Thanks for your continued help Arnaud.  Now that I know what’s expected when referencing a commit, in the future I’ll provide the commit hash.  My two commits started with the last commit 3c2080d11b that you mentioned below.  I’m using the default cload build and in the makefile disabled libdw and dwm1000 object references because that was the only way to make a build without unresolved references and fatal errors.  I’m not using the toolbelt in my windows 10 environment, VMPlayer, Xubuntu 14.10.
Best regards,
Jack

jackemoore:
Hi,
The 2016.05.19 crazyflie-firmware baseline that appears to work fine with my gtgps and compass changes had a commit hash of 6a6b3681f1.
Best regards,
Jack

ataffanel:
I do not think I understand. What does break after the commit 6a6b368?

Also you added a new control mode called "modeAbsVel" what is the difference with "modeVelocity"?

jackemoore:
Hi,

I’ll try to answer your questions.

Topic: New Control Mode modeAbsVel
In my implementation changes for altHoldMode and posHoldMode, a new control mode was added that permits both position x,y,z (modeAbs) and velocity x,y,z (modeVelocity) inputs to be used at the same time to set the effective position setpoints.

When altHoldMode is disabled, setpoint->position.z is continuously updated to the latest state->position.z (derived from baro.asl). When posHoldMode is disabled, setpoint->position.x and setpoint->position.y are continuously updated to the latest gps position.x (delta latitude ) and gps position.y (delta longitude). The deltas are set in the gps driver by subtracting a gps home position from the current gps fix positions (which are scaled by 1x10^7) . This preserves 1cm precision in the gps positions for x and y. 

When altHoldMode and/or posHoldMode are/is enabled, the position setpoints are held fixed at the values last set when entering these modes. Velocity inputs from the controller device (in my case Xbox360) for thrust stick changes (velocity.z * dt) and pitch & roll stick changes (velocity x & y * dt) added together to set the effective position setpoints (effectively modeAbs plus modeVelocity).

The new control mode modeAbsVel is defined in stabilizer_types.h and is applied in commander.c, function commanderGetSetpoint(), for altHoldMode and for posHoldMode.

In position_controller_pid.c, function runPid(), when modeAbsVel is enabled, the setpoint->positions are combined with the setpoint->velocities, and using a lastsetpoint variable to make the effective position setpoints work.

Topic: Custom Firmware and PC Client Builds from Source Setup

Two desktop PCs and three laptops are used for my various crazyflie projects. Setup was originally in Windows 7 platforms, which all but one have recently been upgraded to Windows 10 platforms. 

Python 2.7 and 3,4 installs are used for cfclient builds from source. VMplayer and Oracle VM Virtual Box are used with either VM 0.6 or VM 0.8 to build cf2 firmware within eclipse. Flashing of firmware is primarily performed using the bootloader in cfclient. Sometimes from VM using the crazyradio.

Toolbelt or ARM Toolchain are not used at all. Builds in VM eclipse are made using Make Targets clean & Make CLOAD. Changes are made to Makefiles removing the libdw dw1000 driver VPATH, and PROJ_OBJ_CF2 for libdw1000.o, libdw1000Spi.o, and dwm1000.o. 

Topic: GPS & Compass Additions

As previously stated I have successful firmware using Bitcraze baseline crazyflie-firmware prior to and leading to commit 6a6b368 (2016.05.19) for both the gtgps (GlobalTop Titan-2) driver and for the dkgps (Ublox Max M8c Pico Breakout) driver.

Yesterday, I did uncovered a latent bug in these builds, corrected it with a fix, and made a commit (branch gtgps-6) to my github jackemoore/New-GPS-Additions. Information on this was posted to the Bitcraze forum: https://forum.bitcraze.io/viewtopic.php?f=6&t=1987&start=30#p10534 The bug fix was appled in sensors_stock.c, function sensors Acquire(), in the baseline code itself. The cause of this failure is unknown, whether solely in the original baseline code or something created by what I've done or added.

The changes in gtgps.c, compass.h, and compass.c were primarily to disconnect anything bad happening when the compass was not calibrated. The compass yaw needs to be calibrated in order to correctly apply gps coordinates to the crazyflie orientation, and to correctly apply drift compensation to the gyro.yaw. The latter feature is not required for posHoldMode nor gps position fixes, but was just an interesting exercise to control the gyro.yaw from drifting the crazyflie yaw while flying in hover mode. As a part of these changes, I did a little bit of cleaning up which might have somewhat lessened the processor workload before a 3D gps fix or the compass is calibrated.

I should mention that the watchdog timer has been disabled so that when a compass calibration is performed, the calibration values are saved in eeprom, via configblockeeprom.c. 

I tried this new code in commit 6a6b368 (2016.05.19) and everything appears to work as advertised. Then tried it in the commit 3c2080d (2016.06.24). In the limited testing completed so far, everything appears to work correctly until performing a compass calibration. The crazyflie red led goes off, the cfclient thinks it is still connected to the crazyflie, but packet data is not being updated in cfclient. The failure appears to be at the instant the crazyflie would be asking the configblockeeprom to write to eeprom. So this looks like a timing problem, interrupt problem, stack problem, memory problem, or whatever doesn't like writing to eeprom in the middle of on going tasks. Yet, this problem is not showing up in the earlier commit 6a6b368.

When first experiencing a crazyflie failure in commit 6051b3a (2016.06.14), this was before implementing code to save the compass calibration values in eeprom and the watchdog timer was active. When encountering a wdg timeout, either immediately after a bootloader flash download or when first attempting to connect to cfclient, I clearly didn't have a good picture of the symptoms to describe what was happening. A few symptoms looked like the gps driver wasn't receiving interrupts. That's when I thought there was a uart1 problem.

Since then, symptoms pointed to other things and not the uart1 communication. It wasn't until disabling the wdg timer that better insight and more meaningful symptoms emerged. If the problem seen is related to timing, overrun or overflow, then it's natural to assume that the failure will behave differently depending upon start-up conditions, how quickly the gps obtains a 3D fix, whether the compass was previously calibrated and eeprom already has calibration values, etc.. Perhaps the latent bug uncovered in function sensorsAcquire() has been a part of the symptoms all along.

I don't wish to waste your valuable time. Unfortunately, I don't have enough experience with all of the source code changes implemented between commits 6a6b368 and 6051b3a. I see that Mike Hamer has made a lot of cpu and interrupt changes. It appears that some of what he fixed where hidden problems to begin with.

So, I'm open to suggestions on what to do next. Is there a way to save calibration data in eeprom without causing the crazyflie headaches? In order for you to see the same failure mechanisms I'm seeing, you would need to have a gtgps running and the ability to perform calibration using a controller device similar to my Xbox360. The alternative is for me to bifurcate in some way my firmware where the simplest code shows up an easily observable failure condition.

Best regards,

Jack

tobbeanton:
Hi Jack!

Thanks for the detailed answer! You have come a long way and we are certainly impressed with your quick progress!

I'm curious how well you have got it to work as I was not able to get it to work my self due to not calibrating the compass. E.g. can you get to CF2 to hold its position? Would be nice to see a video :-)

You definitely not wasting our time, don't think that, and we will do our best to help you out!

I suggest we try to find the bugs first as it is a general problem (we must be able to write to the eeprom). Then it might be time to implement the none-volatile parameters to be saved in eeprom for things such as calibration parameters. It will be needed for the Loco positioning system as well. Would it help if we sent you a debug adapter board so you can step code? Do you have a jtag/swd debugger?

Could you please write down in a step-by-step manner on how to replicate the bug?

jackemoore:
Hi,

Thanks for your comments and suggestions.

I made two changes to the firmware code and writing to eeprom appears to be working fine. (1) Increased interrupt buffer size from 64 bytes to 248 bytes in uart1. I know this might be overkill, but I don't have the current means or where-with-all to optimize it. (2) The function call to compassInit() in stabilizerInit() was missing in all commits after commit 6a6b368 (2016.05.19). It's unclear how this mistake would have contributed to the crazyflie crashes when writing calibration values to eeprom, but haven't conclusively proved it didn't.

In trying to keep up with the latest commits being applied to the github crazyflie-firmware-master, each time, I copied my changes into the new commits. The call to compassInit() fell through the cracks.

The binary message size coming out of the GT Titan-2 is 45 bytes @ 5Hz updates & 115200 bps. In the case of the ublox gps there are two binary messages totaling 96 bytes @ 5Hz updates & 115200 bps. Before increasing the buffer size in uart1, I was seeing more failures with the ublox than with the globaltop.

With the above firmware code changes, I'm not seeing any problems with commit 3a2080d (2016.06.24). Yeh!!!

I see today that there is a new commit on github crazyflie-firmware-master. I propose to withdraw my rfp commit #130, fork the newer master commit and then submit a new rfp commit containing all of my latest code changes. Is withdrawing a rfp commit something Bitcraze does?

I apologize, but I seem to spend more time on code that appears to be working rather than spending a lot of time investigating why something in the code is not working.

In summary, the failures I previously saw appear to have gone away with buffer size changes in uart1, reinstating the call to compassInit() in stabilizer.c , replacing the sensorData->position = position; statement in sensors_stock.c, the little bit of cleanup in compass.c and gtgps.c, and disabling the (big umbrella) watchdog timer.

I have a crazyflie 2.0 debug adapter kit, but have never used it. I should be able to easily get a jtag/swd debugger with usb interface.

I haven't yet taken off the training wheels to try out the hover mode (altHoldMode & posHoldMode) in free flight. I'm waiting for seeedstudio to deliver some fresh batteries. The ones I have are more than a year old and these may not be able to hold as full a charge. My cf2 with gtgps weighs about 36 g. If each motor has a thrust potential of 10 g, then the total lift might be 40 g, and this yields a net lift of only 4 g. Currently, in altHold Mode, it uses up too much of the battery charge for me to do any serious testing. Currently, my cf2 with ublox gps weighs 31.5 g, and altHoldMode works fine. The above code problems have limited my testing of posHoldMode, and I will make up for lost time in the coming days. I can see the latitude/longitude corrections are being applied correctly, but haven't yet taken off the belaying lines and tested how good the default pidx and pidy constants perform. Btw, I mount the battery on the underside of the control board to lower the cog. Also, the adaptive altHoldMode using the filtered baro.asl instead of a fixed thrust approach appears to work much better with the heavier cf2's. 

Also, I'm not totally satisfied with the rube-goldberg-machine approach I've taken to implementing drift correction to gyro.yaw. I intend to explore using a formal complementary filter (hpf on gyro & lpf on mag).

Best regards,
Jack

tobbeanton:
Great that you found a solution! I also found that problem which I wrote about in the GPS thread in the forum. It is a uart owerrun that causes the hangup.

You should be able to withdraw the pull request. You can then do changes and make a new one. That is better then starting all over again. Also the commit record is saved so it is possible to track when things where done rather then having it all in one commit.

As for the weight I don't think it is a problem. It will change the flight dynamics a bit and lower the flight time with ~30sec but not much more. Having a lighter GPS would be nice though :-)

jackemoore:
Hi,
Good detective work!  It’s great that you isolated and diagnosed this problem.  It helps confirm that changing the ISR buffer size is one effective way to reduce the overrun potential.

This morning, I updated my github branch gtgps-6 firmware with changes to the uart1 buffer size and with a fix to compass.c to keep the gyro.yaw , after drift compensation is applied, to stay within +/- 180 degrees.

Unfortunately,  the chip antenna on the ublox is such a poor performer and is very unreliable receiving satellite signals when mounted anywhere near the crazyflie.  I’m hoping new batteries will help me with the gygps.

Best regards,
Jack

jackemoore:
Hi,
I forgot to mention that in my latest config.mk  it contains CFLAGS += -DDECK_FORCE=bcGTGPS and CFLAGS += -DEXPERIMENTAL_ATTITUDE_PID
Best regards,
Jack

jackemoore:
Hi,
Well, the attempted fix to sensors_stock.c to fix the problem caused by the statement “sensors->position = position;”, that was causing a problem with baro.asl, moved the problem from sensorsData_t to another structure.  In so doing it broke the posHoldMode operation and clobbered “this,pidx.setpoint” & “this.pidy.setpoiont” in module position_controller_pid.c.  Or clobbered some other structure variables (rates ?) that make up these target x & y setpoints.  I’m going to need help from Bitcraze who are more knowledgeable about these structures.  Until resolved, I suspect LPS integration will also encounter this same problem.
Best regards,
Jack

jackemoore:
Hi,
The sensors_stock.c problem mention below has been resolved and corrected.  Please see my new post under Crazyflie 2 with GPS in the forum for more details.
Best regards,
Jack

PR #131, opened 2016-06-30
Convert deck_spi to use DMA

mikehamer:
This pull request converts the deck_spi to use DMA. This has been developed and tested with the DW1000 UWB module, and hence slow mode is hard-coded to ~2.7MHz, and fast mode to ~21MHz.

This pull request significantly reduces the load on the CPU (since it is not interrupted for every SPI byte), and is particularly useful for applications requiring a high number of frequently occurring SPI transactions (eg. UWB localization). 

In addition, this pull request enables SPI speed switching for the dw1000 deck driver.

### DIFF
--- a/src/deck/api/deck_spi.c
+++ b/src/deck/api/deck_spi.c
@@ -26,10 +26,6 @@
 
 #include "deck.h"
 
-#include "stm32fxxx.h"
-
-#include <string.h>
-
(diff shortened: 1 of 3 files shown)


ataffanel:
Thanks it is merged. This should help a lot the UWB software :-)

PR #132, opened 2016-07-01
Update motors_def_cf2.c

TU-KL-Crazyflie:
First of all the comment in line 402 claims to use Timer 3 Channel 1 for MOSI on PA7. However PA7 only has acces to Timer 3 Channel 2! 
Either the comment needs to be changed, or an other timer should be used. Since CPPM is unlikely wanted to be used when someone changes the motormapping to PA7, I propose to use timer 14 channel 1 in the motordefinition.

Furthermore the default motor mapping of the Bigquad Deck as described in lines 437 to 440 is wrong and should be replaced with my changes.

### DIFF
--- a/src/drivers/src/motors_def_cf2.c
+++ b/src/drivers/src/motors_def_cf2.c
@@ -399,7 +399,7 @@ static const MotorPerifDef DECK_MISO =
     .preloadConfig = TIM_OC1PreloadConfig,
 };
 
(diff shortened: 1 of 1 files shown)


tobbeanton:
As only the motor mapping has been changed you can ignore my suggestions in the [forum thread](https://forum.bitcraze.io/viewtopic.php?t=2077&p=10580#p10580).

PR #134, opened 2016-07-04
Added Replacement for Deck Driver gtgps.c

jackemoore:
> > Driver handles custom firmware in GT Titan-2, and only processes
> > one-sentence binary messages. NMEA message parsing has been removed to
> > lessen microprocessor workload and possible timing conflicts.  Driver
> > also corrects coding errors in the baseline gtgps.c.
> > Log_Group(GPS) supports baseline cfclient GpsTab.py
> > (lat,lon,hMSL,nsat) as well as the proposed GpsTab.py update (adds hAcc,
> > fixType) that has previously been submitted in a cfclient pull request.
> > config.mk needs to have CFLAGS += -DDECK_FORCE=bcGTGPS
> > COMPASS_ENABLE is not defined and disables link to compass.c.
> > Driver code includes support for posHoldMode using gps.
> > Proposed modules supporting the compass and compass calibration are
> > not included in this commit.
> > Proposed modules supporting posHoldMode using gps are not included in
> > this commit.

### DIFF
--- a/src/deck/drivers/interface/gtgps.h
+++ b/src/deck/drivers/interface/gtgps.h
@@ -1,0 +1,34 @@
+/**
+ *    ||          ____  _ __
+ * +------+      / __ )(_) /_______________ _____  ___
+ * | 0xBC |     / __  / / __/ ___/ ___/ __ `/_  / / _ \
+ * +------+    / /_/ / / /_/ /__/ /  / /_/ / / /_/  __/
+ *  ||  ||    /_____/_/\__/\___/_/   \__,_/ /___/\___/
+ *
+ * Crazyflie control firmware
+ *
+ * Copyright (C) 2011-2012 Bitcraze AB
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


tobbeanton:
This is a great step forward for the GPS support, thanks Jackemoore!

Some thoughts:
* Variable naming is not really consistent with the rest of the code. We normally don't use underscores e.g. gps_x but instead camel case with small letter in the beginning e.g. gpsX. However the [coding standard](https://wiki.bitcraze.io/codingstandard:index#coding_standard_for_python) is not really up to date an easy to find, so shame on us. Anything you might be willing to fix, else I think we are ok with integrating it anyway.
* Any possibility to separate the parsing from the decoding so the parsing could be reused for other GPS vendors?

jackemoore:
Sorry, I've been gone on holiday since August 13th where there is no internet or cellular coverage.  I'll reply as soon as I get caught up on a few things.

Jack

krichardsson:
We close this pull request for now as it can not be merged in its current state and there has not been any progress for a long time. If you want to reopen it and continue the work we will be more than happy to take another look at it.

Thanks for you effort!

PR #139, opened 2016-08-06
Creating two variables for VBat an ExtVBat in mV

theseankelly:
I'm working on plumbing telemetry support for vBat for both the crazyflie and the bigquad expansion deck into the deviationTx Cflie protocol. The CPU on most deviation controllers doesn't have an FPU, and so we need vbat variables with mV resolution


krichardsson:
Merged.
Thanks!

PR #140, opened 2016-08-23
Adding tmp36 (temperature) sensor

josephinemonica:
This is to add tmp36 sensor to crazyflie.

- This code assumes a default value of Vref to tmp36: 3000 mV. 
To change : hal/src/tmp36.c edit data2C function, change 3000 to the new Vref
- This code assumes to connect the measuring pin of tmp36 to TX2
To change : driver/interface/tmp36_driver.h , change #define tmp36_GPIO DECK_GPIO_TX2 to the desired pin.
- If you don't want to use tmp36:
open driver/interface/tmp36_driver.h  and hal/interface/tmp36.h 
Comment all the #define tmp36_driver_ENABLED and #define tmp36_ENABLED

You can see a new logging group "tmp36" in the Logtoc tab in the crazyflie client.
You can plot the temperature too : setting-> logging configuration ->select tmp36 log group, choose logging period & name, and proceed. Then plot!


josephinemonica:
Hi, can help to figure out what causes travis CI build to fail? The code works fine on my laptop.

krichardsson:
In the travis build we run a number of builds with various combinations of build flags to verify that the code compiles. It looks like your code does not compile for Crazyflie 1 (PLATFORM=CF1).

/module/src/modules/src/system.c:128: undefined reference to `tmp36Init'

josephinemonica:
@krichardsson I'm quite new to it. Can you help to suggest how to fix it? Thanks

josephinemonica:
@krichardsson Thank you for commenting. In gave me enlightment

ataffanel:
Since this is added to Crazyflie 2.0 I think this code would be better as a deck driver (https://wiki.bitcraze.io/doc:crazyflie:api:firmware:deck). This will allow to add the code with minimal modification of the rest of the firmware and only people with this sensor attached will then run this driver.

Otherwise the code in itself looks good to me.

krichardsson:
Please let us know if you need any help in refactoring the code as a deck driver. We will be more than happy to help out!

krichardsson:
Closing due to lack of activity, please feel free to reopen.

PR #144, opened 2016-09-09
Master.mag driver fixes

stephanbro:
I've fixed the configuration of the mag so the values are updated properly now when using the sensors task and kalman estimator.

The barometer isn't working right now and I was planning on tackling that next.

### DIFF
--- a/src/drivers/interface/ak8963.h
+++ b/src/drivers/interface/ak8963.h
@@ -54,7 +54,7 @@ THE SOFTWARE.
 #define AK8963_RA_HZH             0x08
 #define AK8963_RA_ST2             0x09
 #define AK8963_RA_CNTL            0x0A
-#define AK8963_RA_RSV             0x0B // RESERVED, DO NOT USE
+#define AK8963_RA_CNTL2           0x0B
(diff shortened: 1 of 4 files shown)


tobbeanton:
Many thanks for this fix! I'm curious what moving  mpu6500SetClockSource(MPU6500_CLOCK_PLL_XGYRO); does?

tobbeanton:
I will check with a i2c analyser to see how the reading behaves. Then we know if it is better to revert it.

tobbeanton:
After checking the i2c log I can say it behaves pretty well so no need to revert it. One thought i got after looking att the sensors_task way of reading the magnetometer is to do the same in imu_cf2. That is reading x,y,z and ST2 (overflow) in one read.

By the way, in the long run I think we should move the sensors_task way of reading using the MPU9250 master into imu_cf2 and make them use the same API and not have duplicated code.

stephanbro:
That's great to hear! I was thinking the same thing about moving a lot of the sensors task code over to the imu_cf2 module. There is a fair bit of cleanup that can be done on the slave reading and writing also.

PR #145, opened 2016-09-13
[WIP] Baro driver slave configuration

stephanbro:
I was able to configure the barometer to get data coming out, but if I have both the magnetometer and barometer enabled, I only get data from the magnetometer. I've verified the registers on the devices stay the same and the only line I'm changing is `mpu6500SetSlaveEnabled(0, false);` to `mpu6500SetSlaveEnabled(0, true);`

@tobbeanton, could you use your I2C magic skills to help me out?

### DIFF
--- a/src/modules/src/sensors_task.c
+++ b/src/modules/src/sensors_task.c
@@ -86,7 +86,7 @@ static bool isMagnetometerPresent = false;
 
 #define MPU6500_BUFF_LEN 14
 #define MAG_BUFF_LEN 8
-#define BARO_BUFF_LEN 5
+#define BARO_BUFF_LEN 6
 
 // This buffer needs to hold data from all sensors
 static uint8_t buffer[MPU6500_BUFF_LEN + MAG_BUFF_LEN + BARO_BUFF_LEN] = {0};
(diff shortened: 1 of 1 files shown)


tobbeanton:
Checked the AUX i2c bus with an analyzer with both mag and baro activated. Looks like the mag and baro does the setup sequence but the baro never gets called again after that, only the mag.
[AUX_I2C_startup baro and mag.csv.txt](https://github.com/bitcraze/crazyflie-firmware/files/471815/AUX_I2C_startup.baro.and.mag.csv.txt)

tobbeanton:
I was just thinking. Wouldn't it be easier to use AUX i2c passthrough when setting up the slaves and configuring them and just setup the data transfer through the MPU9250 master?

tobbeanton:
Something does not seem right with the sample rate when checking the i2c timing log. Setting mpu6500SetSlave4MasterDelay seems to follow the formula 8000Hz / (1+I2C_MST_DLY). Investigating...

tobbeanton:
I think I found why. A small note in the register datasheet: 
> Data should be sampled at or above sample rate; SMPLRT_DIV is only used for 1kHz internal sampling.

As we use DLPF of 250Hz which uses 8kHz it doesn't apply and the formula 8000Hz / (1+I2C_MST_DLY) makes sense.

tobbeanton:
Can't get why it only starts on slave and now I have to go home for the day. I attached the changed sensors.c file if you would like to use that.
[sensors_task.c.txt](https://github.com/bitcraze/crazyflie-firmware/files/472216/sensors_task.c.txt)

stephanbro:
Nice clean up! I had originally done the configuration through the master because I was having issues if I had the normal configuration and self tests running before setting up the master items, but those issues were probably related to something else. In the MPU's datasheet it even says that the configuration should be done with the I2C bus set to pass through so your changes are very welcome.

Something strange is definitely going on with the DLPF and how the clock relates to that. When I commented out that macro, both sensors started working at the same time. I'm going to do some more digging today.

stephanbro:
Something else I saw in the datasheet for the same register:
_4.4 Register 25 - Sample Rate Divider_
>NOTE: This register is only effective when Fchoice = 2’b11 (fchoice_b
>register bits are 2’b00), and (0 < dlpf_cfg < 7), such that the average filter’s
>output is selected (see chart below).

Right now fchoice_b isn't explicitly set so it's hopefully 2'b00, but the more interesting part is the 0 < dlpf_cfg < 7 part. It looks at though when the internal sampling rate is > 1kHz then the SMPLRT_DIV register "isn't effective". I assume that means it doesn't have an effect on the output?

tobbeanton:
I will merge this pull request to bring it in so I can continue to work on it even though it is not fully working yet.

tobbeanton:
After several hours of trail and error I found out a fix. The DLPF needs to have a value using 1kHz internal sampling rate (2 to 7). Then it reads more then one slave. Can't find any information about this in the register datasheet...

stephanbro:
Good find. I figured it was something nefarious with the DLPF and the internal sampling rate. Hopefully it makes sense from the gyro and accel data to be at a higher filtering rate or something else is figured out for reading from the two slaves.

PR #147, opened 2016-09-16
Master.crtp position update

stephanbro:
This pull request adds a new port to CRTP for sending an external position. The commander registers the callback and buffers the data, while the stabilizer calls the commander to queue it's current data.


carlin-psvl:
To clarify the purpose of these commits, it allows the CF2 to be controlled by a motion capture or similar system when the EKF estimate fails or the CF2 starts to move out of LPS range. This is useful in testing or demonstration, as you can provide a virtual safety net of sorts.

ataffanel:
Thanks, this looks good. I only have one comment: I would like this functionality to be put in another file than the commander. Commander.c is already too big and full of random stuff. What about moving the code to extPosition.c or any other name that makes sense?

I like the renaming of the commander port in setpoint port. It makes more sense to call it setpoint.

tobbeanton:
Shouldn't we use a new "channel" in the commander/setpoint port for the position setpoint rather then occupying a new port?

ataffanel:
No I do not think it should be on the same port: the commander is about setpoint, this is more like an external sensor. To me it makes sense to have them separated.

tobbeanton:
I see, the commender is about commanding things and the position is just information. Agreed we should keep them separated.

stephanbro:
Thanks for the feedback! I've moved the external position items to their own file. Let me know if you have any other changes you'd like me to make.

I also noticed that libdw1000's submodule wasn't pointing at master as of d38113b8ac58f9191b0fc956036c5fcad4c15045 and doesn't actually build with `-DDWM1000_ENABLE_TDMA`! I added a commit to point it back to master. It might be good to add that build config to TravisCI.

ataffanel:
I merged it. I am not so sure about where to call getExtPosition but this is now together with the kalman code that we intend to clean-up anyway (it would be nice to keep the stabilizer loop generic).

Thanks! it is nice to have this in the master banch.

PR #150, opened 2016-09-23
Broadcast support

whoenig:
This change is related to https://github.com/bitcraze/crazyflie2-nrf-firmware/pull/15. It adds the STM32 side of the broadcasting support.


tobbeanton:
Thanks! I've merged this as it doesn't break anything even though the nRF part still has some work.

PR #151, opened 2016-09-26
Accelerometer Filtering

stephanbro:
In an attempt to improve the Z estimate, I've added some filtering to the accelerometer's output values that is based on the PX4 MPU9250 driver (BSD-licensed).

The changes are as follows:
* Increase MPU6500 range to 16g's
* Set on chip DLPF to 42Hz
* Add a 2-pole low pass filter w/ 30Hz cutoff

The results look positive and while flying using the LPS, I did not have a large deviation in Z estimate vs actual as I had prior to filtering. Both of the plots show below consist of a takeoff, hovering in place, and a landing.

Before:
stock_acc
After:
filtered_acc

### DIFF
--- a/src/hal/src/sensors_cf2.c
+++ b/src/hal/src/sensors_cf2.c
@@ -51,6 +51,7 @@
 #include "nvicconf.h"
 #include "ledseq.h"
 #include "sound.h"
+#include "filter.h"
 
 /**
  * Enable 250Hz digital LPF mode. However does not work with
@@ -119,6 +120,7 @@ static Axis3f  gyroBiasStdDev;
 static bool    gyroBiasFound = false;
 static float accScaleSum = 0;
 static float accScale = 1;
+static lpf2pData accLpf[3];
 
 static bool isBarometerPresent = false;
 static bool isMagnetometerPresent = false;
@@ -153,6 +155,7 @@ static void sensorsCalculateBiasMean(BiasObj* bias, Axis3i32* meanOut);
 static void sensorsAddBiasValue(BiasObj* bias, int16_t x, int16_t y, int16_t z);
 static bool sensorsFindBiasValue(BiasObj* bias);
 static void sensorsAccAlignToGravity(Axis3f* in, Axis3f* out);
+static void sensorsAccApplyLpf(Axis3f* in);
 
 bool sensorsReadGyro(Axis3f *gyro)
 {
@@ -293,6 +296,7 @@ void processAccGyroMeasurements(const uint8_t *buffer)
   accScaled.z =  (az) * SENSORS_G_PER_LSB_CFG / accScale;
 
   sensorsAccAlignToGravity(&accScaled, &sensors.acc);
+  sensorsAccApplyLpf(&sensors.acc);
 }
 
 static void sensorsDeviceInit(void)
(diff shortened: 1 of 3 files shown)


src/hal/src/sensors_cf2.c
```diff
@@ -348,9 +351,13 @@ static void sensorsDeviceInit(void)
   // To low DLPF bandwidth might cause instability and decrease agility
   // but it works well for handling vibrations and unbalanced propellers
   // Set output rate (1): 1000 / (1 + 1) = 500Hz
-  mpu6500SetRate(1);
+  mpu6500SetRate(0);
   // Set digital low-pass bandwidth
-  mpu6500SetDLPFMode(MPU6500_DLPF_BW_98);
+  mpu6500SetDLPFMode(MPU6500_DLPF_BW_42);
```
tobbeanton: This will set the gyro DLPF. Is that what you intended to do or you wanted to change the accel DLPF? I think this setting might slow down the system.

src/hal/src/sensors_cf2.c
```diff
@@ -335,10 +339,9 @@ static void sensorsDeviceInit(void)
   // Set accelerometer full scale range
   mpu6500SetFullScaleAccelRange(SENSORS_ACCEL_FS_CFG);
 #ifdef ESTIMATOR_TYPE_complementary
+  mpu6500SetRate(0);
   mpu6500SetAccelDLPF(MPU6500_ACCEL_DLPF_BW_20);
```
tobbeanton: This is the accel DLPF setting. Probably what you want to change?

src/hal/src/sensors_cf2.c
```diff
@@ -348,9 +351,13 @@ static void sensorsDeviceInit(void)
   // To low DLPF bandwidth might cause instability and decrease agility
   // but it works well for handling vibrations and unbalanced propellers
   // Set output rate (1): 1000 / (1 + 1) = 500Hz
-  mpu6500SetRate(1);
+  mpu6500SetRate(0);
```
tobbeanton: This will change setting to 1kHz sampling output right? Please update comment above if that is the case. Also does 1kHz sample output improve anything, latency maybe?

src/hal/src/sensors_cf2.c
```diff
@@ -398,7 +405,7 @@ static void sensorsSetupSlaveRead(void)
   // SMPLRT_DIV is only used for 1kHz internal sampling." Slowest update rate is then 500Hz.
   mpu6500SetSlave4MasterDelay(15); // read slaves at 500Hz = (8000Hz / (1 + 15))
 #else
-  mpu6500SetSlave4MasterDelay(4); // read slaves at 100Hz = (500Hz / (1 + 4))
+  mpu6500SetSlave4MasterDelay(9); // read slaves at 100Hz = (500Hz / (1 + 4))
```
tobbeanton: Please update comment as well. Faster sampling rate might improve latency? Reason why you set it to 100Hz?

tobbeanton:
I was eager and merged the changes plus the suggested review points and some other position control improvements. Can you please try the changes @stephanbro? Now the more aggressive PID is active for kalman filter as well. It makes it nervous but more precise.

src/hal/src/sensors_cf2.c
stephanbro: Good catch. I think that's likely the case.

src/hal/src/sensors_cf2.c
stephanbro: You are right, I don't think we want to change the DLPF on the gyro.

stephanbro:
Sorry I didn't get back sooner about your comments; I've been out sick unfortunately. I know you guys have the maker faire coming up this weekend, so I will see what I can do to get your changes tested today.

stephanbro:
@carlin-psvl were you able to get any test flight in today?

carlin-psvl:
I wasn't able to gather more accelerometer data but I did test some quick test flights. Qualitatively speaking, height control is improved over my previous flights. I needed to add an integral limit to to the position PIDs, without them the CF had a tendency to build up a lot of integral momentum and orbit its goal point.

PR #153, opened 2016-09-29
Added full kinematics for the not flying case

mikehamer:
This improves the state estimate for cases where the quad is being carried, or in free fall. The quad's position is no longer constrained to being on the ground when the props aren't spinning, and body velocities are no longer constrained to zero.


ataffanel:
Thanks! it is merged.

We have been experimenting with starting the autonomous flight when throwing the Crazyflie, this will help make it more stable :).

PR #156, opened 2016-10-10
VL53L0X Time of Flight Sensor Driver

stephanbro:
Based on the arduino driver from Pololu found here: https://github.com/pololu/vl53l0x-arduino.

Items remaining to do are setting a proper VID, PID, and used pins.


tobbeanton:
Many thanks for the pull request. As soon as we have some HW to test it with we will!

PR #157, opened 2016-10-15
README: Add build instructions for Windows

aswild:
The ARM Embedded toolchain is available for Windows, so I wrote up instructions for using it within Cygwin (which supports git and make, which is all you really need to build the CF firmware).

Disclaimer: While I tested the steps provided in a fresh install on a Windows 7 VM and verified that the bin/elf/hex files get created without errors (for CF1/CF2), I don't have any crazyflie hardware to test the images on.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,19 @@ sudo pacman -S community/arm-none-eabi-gcc community/arm-none-eabi-gdb community
 
 #### Windows
 
-> `TODO: Please share!`
+The GCC ARM Embedded toolchain for Windows is available at [launchpad.net](https://launchpad.net/gcc-arm-embedded/+download). Download the zip archive rather than the executable installer. There are a few different systems for running UNIX-style shells and build systems on Windows; the instructions below are for [Cygwin](https://www.cygwin.com/).
+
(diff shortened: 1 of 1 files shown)


ataffanel:
Thanks! that can be very useful. Have you tried running openocd and the bootloader client inside cygwin?

aswild:
Hey there!
I hadn't checked openocd, but I just tried it out (on my main environment with lots of extra cygwin dev packages installed) and it seems to work.

I didn't find binary packages but you can get the source tarball (zip/gz/bz2) and do the standard `./configure; make; make install`. There's even a man page!

The Python Client was easy on Windows back when I worked on Crazyflie 2 in early 2015, but now it's being a pain in the butt. I had to install `pip` dependencies and then the new python lib, and now I'm manually building libzmq for some C++ dependency of the python client...

Installing the libtool Cygwin package, I built libzmq (as `cygzmq.dll`) with `./autogen.sh; ./configure --without-docs; make; make install`

With that, I can finally get `python3 setup.py install` to work in `crazyflie-clients-python`, and the `crazyloader` command works (but `make cload` in `crazyflie-firmware` segfaults, presumably because I have no hardware)

Issue #136, opened 2016-07-25
Revamp position controller


ataffanel:
Currently the position controller is a single position PID, it should be revamped to include a velocity loop. This would allow to use the new experimental PID for the attitude controller (see a0628ec2c).

mikehamer:
Another very important addition would be the inclusion of feedforward terms. For example, if a trajectory is planned ahead of time, the desired position, velocity and acceleration (and therefore thrust and body angle) are known for each time instant. Giving these derivative terms to the controller as "feed forward" terms allows significant improvement in trajectory tracking.

ataffanel:
Shall we add that as a separate ticket maybe and concentrate this one on fixing the controller for the curent setpoint structure. Or is it required to define the "feed forward" architecture first?

stephanbro:
I've implemented a velocity controller that takes in the position PID controller output and generates the attitude setpoint information. You can see the commit here: c3c72d7414aca1c907f16acb9c7a1fc627c35500

This gives us much better control over things like the maximum velocity and we also have the potential to run the velocity control at a higher frequency.

ataffanel:
We just tested it and saw great improvement in stability for stationary setpoint and it basically remove position overshoot. We have seen some strange trajectory when going from one setpoint to another but we have to test more and it is anyway much better than before! I have tried to set the max x/y velocity to 5m/s (being a bit greedy :) but it then starts to overshoot again.

For the height control it still oscillates a bit but previously I got it more stable by setting PID_YAW_RATE_KP at 150.0. We will test that tomorrow as well.

stephanbro:
Glad to hear of the improvement! I haven't had a ton of opportunity to do tuning, but right now there is no D term, which could likely help out with the overshoot you were seeing.

Feel free to send me any rosbags or logs if you notice anything funny.

mikehamer:
Reviving this discussion...

I have a student working on swarm control and he would benefit greatly from being able to send feed-forward terms (eg. a trajectory in terms of position, velocity and acceleration). I've therefore thrown together a framework for a controller and communication packets that should enable this, and am well on the way towards implementing the controller.

The packets are defined with minimizing communication in mind:

1. You shouldn't need to send a separate packet to indicate an emergency, this is therefore included in the header. It should interact with potential emergency shut-off functionality (see #168).
2. You should be able to define the control mode in the same packet that has the control input (no more setting parameters)
3. You shouldn't need to send a separate packet for an external position

[...]

PR #158, opened 2016-10-18
Velocity Controller and Additional Filtering

stephanbro:
This PR adds a velocity controller as an intermediary between the position and attitude controllers which gives a lot better performance on holding position as well as gives the ability to limit the velocity on board. I've done some basic tuning on the controller, but the tunings could still use work!

I've also added an LPF filter to the gyro at 80 Hz to help filter out some noise for doing attitude rate controls.

A few other changes are listed below:
* Increase yaw rate KP to 120
* Add error max and integral disable to PID library
* Add quaternion to kalman log
* Change position/velocity PID parameters to modify stored pid values instead of init values


tobbeanton:
This will be interesting to test. I've seen this used in many flight controllers so it is most likely the right way to go. I've recently discovered [this](https://github.com/cleanflight/cleanflight/pull/2087) pull request for cleanflight which might be of interest.

tobbeanton:
I did a quick test and I'm not sure it made any improvement. It is now wobbling much more, maybe because of delay in D-Term caused by LPF? Are you seeing the same? Maybe the stiffer motor mounts you are using could make a difference?

stephanbro:
Interesting... Thanks for linking that issue, it was a very interesting read. Definitely something to think about going forward.

I wasn't able to reproduce the wobbling that you speak of. Are you using the stock PID tunings in the build? In what way is there wobbling? They delay in the D-Term due to the filter can definitely create some issues. Could you try increasing the cutoff frequency and see if that helps?

I did some data collection on the stiffer motor mounts I was experimenting with and I wasn't able to find anything conclusive to say that they are better, so I've been using the stock arms for this work. I've also begun work on separating the position and velocity controllers. Have you started work on that yet?

ataffanel:
Hi Stephan, we have not started on separating the controllers yet. It feels that getting a good setup of the controller with position and velocity separated would be a good first step to improve them. Maybe we can sync on #136?

stephanbro:
@tobbeanton I increased some of the cutoff frequencies which will hopefully alleviate your issues.

I've also implemented a velocity controller as an intermediary between the position and attitude controllers here: c3c72d7414aca1c907f16acb9c7a1fc627c35500. My qualitative impression is it a major improvement on the overall ability to hold position, so please give it a try!

tobbeanton:
@stephanbro We tested the velocity controller and it works great! It is very stable holding the position now in x,y. Now the height isn't that precise but @ataffanel might have some tuning settings for that. Interesting times indeed!

As for the filtering I don't see the wobble any more when flying manually but still some small oscillations. It is very hard to tell if the performance is better, maybe you can test that easier? Without a performance increase there is just added computations for the filtering so it would be nice to have some kind of measurement. If I set the cutoff frequencies high will it be the same as disabling the filtering (getting rid of the delay)?

stephanbro:
@tobbeanton Thanks for the feedback! The Z position and velocity generally looked okay with the tunings I had, but I was having my usual issues of the Z position estimate being a bit jumpy so it was hard to get it super good.

I decided to remove the LPF on the D term after doing some test flights and not seeing much improvement in the flight (besides the setpoints looking a lot smoother). In the future we may want to revisit what is the best solution is to the setpoint spikes I had posted before on this PR.

stephanbro:
Is there anything else needed to get this merged? Maybe more tuning?

ataffanel:
This is good to merge as such, I just wanted to test it but we did not get the time today, will be merged tomorrow. We just talked about re-adding the filter with a compile flag so that it will be easy to enable if someone wants to play with it. The plan is to test and merge tomorrow morning (in europe :-). Sorry for the delay.

More tuning can come later, this is already much better than master!

ataffanel:
I reverted your last commit and added a configuration flag to enable the filter per-pid. When doing some tests I found that the filter actually helps for the position controller so I re-enabled it for position controller and merged the PR.

Thanks a lot, this is starting to fly really well now!

PR #159, opened 2016-11-03
Log additional variables and fix a bug

lkumar93:
Hi,

I have logged variables in sensfusion6.c, which would be useful for off-board processing. I have also fixed a param group naming bug in sensors_stock.c


krichardsson:
Merging
Thanks!

Issue #113, opened 2016-05-19
New commander packet


ataffanel:
Currently the commander packet format assumes a "normal" angle-based control mode and every other modes like altitude hold, position hold, rate mode, are handled using parameter and by fitting the relevant values in the current commander fields.

The plan is to make a new commander packet that encodes both the flight modes and the relevant setpoints values. This would make handling thing like altitude-hold and rate mode easier and it is required to start implementing position control in the Crazyflie.

We need to handle at least:
- Teleop in angle and rate mode
- Altitude-hold in angle and rate mode
- Position hold (keep current position and accepts x/y/z speeds to change position, kind of X,Y,Z teleop)
- Position Set (accepts absolute X,Y,Z)
- Others? ...

whoenig:
- I am not sure what the difference between "Position hold" and "Position Set" is. For X/Y/Z teleop, position set seems to be sufficient (i.e. using the joystick updates the goal position). So it seems to me that "Position Set" is sufficient.
- Position Set/Hold should also get a target yaw angle (i.e. X/Y/Z/Yaw)
- There should be a more advanced version for the future which takes a trajectory point. The minimal version would be X/Y/Z/X dot/Y dot/Z dot/Yaw. More aggressive controllers might also need the accelerations (i.e. X dot dot, Y dot dot, Z dot dot) and yaw rate. This could be even combined with the simple version (for hover the target velocities and accelerations are simply 0).

ataffanel:
- With position hold the setpoint is X,Y,Z velocity. With position set the setpoint is an absolute X,Y,Z. I see two different use-case that needs these two modes:
   - For position set the use-case is the copter receiving commands from a computer, for example for waypoint navigation
   - Position hold if more for 'advanced teleop'. For example it has been useful when debuging the internal controller: I flew the copter in the middle of the room and activate position hold, then the copter just stays there and I can move its setpoint. This could also be useful for GPS flight later. Basically in case where a computer is not necessary.
 - Agreed
 - We should find a way to keep simple control simple and allow for more complex one. If we combine the two mode these extra data could be optional.

whoenig:
[...]

PR #161, opened 2016-11-11
Pid update

stephanbro:
* Changes Z command to a float to allow negative values
* Change errorMax to outputLimit in the PID module and does as the name implies
* Smarter cap on the integral
* Add roll, pitch, and yaw rate limits
* Tweak the ordering of assignment in the commander module to make things more explicit
* Properly set PID outputLimit so a parameter can change the value

Also see PSVL/lps-ros#1 and whoenig/crazyflie_ros#49

### DIFF
--- a/src/modules/src/commander.c
+++ b/src/modules/src/commander.c
@@ -37,7 +37,8 @@
 #include "debug.h"
 #include "ext_position.h"
 
-#define MIN_THRUST  1000
+#define MIN_THRUST  0
+#define IDLE_THRUST 1000
 #define MAX_THRUST  60000
 
 /**
@@ -77,7 +78,6 @@ static CommanderCache* activeCache;
 
 static uint32_t lastUpdate;
 static bool isInactive;
-static bool thrustLocked;
 static bool altHoldMode = false;
 static bool posHoldMode = false;
 static bool posSetMode = false;
@@ -93,7 +93,7 @@ static void commanderCrtpCB(CRTPPacket* pk);
 static void commanderCacheSelectorUpdate(void);
 
 /* Private functions */
-static void commanderSetActiveThrust(uint16_t thrust)
+static void commanderSetActiveThrust(float thrust)
 {
   activeCache->targetVal[activeCache->activeSide].thrust = thrust;
 }
@@ -113,7 +113,7 @@ static void commanderSetActiveYaw(float yaw)
   activeCache->targetVal[activeCache->activeSide].yaw = yaw;
 }
 
-static uint16_t commanderGetActiveThrust(void)
+static float commanderGetActiveThrust(void)
 {
   commanderCacheSelectorUpdate();
   return activeCache->targetVal[activeCache->activeSide].thrust;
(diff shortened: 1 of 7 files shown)


whoenig:
Hi,
I was wondering what the benefit of the float-thrust value is. I see that this allows negative values now, but this doesn't have any physical meaning as the rotors can not spin in opposite directions?
Thanks,
Wolfgang

stephanbro:
It's true that the thrust cannot go negative, but within the crazyflie-firmware the name thrust is somewhat of a misnomer, as it is used for direct/scaled thrust, z velocity, or z position. The value should be properly limited based on what the flight mode is within the crazyflie-firmware.

ataffanel:
Hi,
Thanks for the PR, the PID related things looks good!

For the commander packer: we will not change the current commander packet since too many client are currently dependent on it. CRTP has port and channels to route packets, the current plan is to create a new commander packet format on a new channel (see #113). The current usage of thrust as height setpoint is a hack made to work until we have the new commander packet, going forward we should go with well formed packets using SI units.

If you remove the commander part I can merge the PR and lets talk on #113 on how to go forward with the commander packet. I had grand plan to revamp the code but we can just add a new packet quickly to get the functionality in and clean the implementation later. Basically anything as long as we do not break the outside interface :).

theseankelly:
Huge +1 to not changing the current commander packet!!

src/modules/src/commander.c
```diff
@@ -37,7 +37,7 @@
 #include "debug.h"
 #include "ext_position.h"
 
-#define MIN_THRUST  1000
+#define MIN_THRUST  0
```
theseankelly: I'm curious why this was originally 1000

src/modules/src/commander.c
```diff
@@ -174,10 +175,6 @@ static void commanderCrtpCB(CRTPPacket* pk)
   crtpCache.targetVal[!crtpCache.activeSide] = *((struct CommanderCrtpValues*)pk->data);
   crtpCache.activeSide = !crtpCache.activeSide;
   crtpCache.timestamp = xTaskGetTickCount();
-
-  if (crtpCache.targetVal[crtpCache.activeSide].thrust == 0) {
```
theseankelly: I'd prefer to keep the thrust lock -- it's a good safety feature, especially when using transmitters with a non-spring loaded throttle that might not be at zero on power-on, and with BigQuad builds.

src/modules/src/commander.c
tobbeanton: Yep good question. The brushed motors don't start spinning until ~3000 so I think it was initially intended so that power isn't consumed without any movement. When it spinns it can go lower so that is were 1000 comes from.

stephanbro:
That makes sense. I can remove the non-PID related changes and close the other pull requests.

src/modules/src/commander.c
stephanbro: Interesting. Is this a change you'd like me to keep @tobbeanton @ataffanel ?

src/modules/src/commander.c
stephanbro: Understandable. I'll remove this along with the other commander changes I've made.

ataffanel:
Hi, Thanks! Tested and merged.

PR #162, opened 2016-11-18
Fix the state update in the prediction step of the EKF

estromb:
Currently, the update overwrites the state and then updates the other states with the overwritten ones. I added a few temp variables to save the old state through the update.

I also changed some declarations pertaining to this state update to static, so the wouldn't be re-declared at 100 Hz.

The code builds in the Bitcraze VM, but I don't have access to a Crazyflie to see if it runs.

### DIFF
--- a/src/modules/src/estimator_kalman.c
+++ b/src/modules/src/estimator_kalman.c
@@ -619,18 +619,22 @@ static void stateEstimatorPredict(float cmdThrust, Axis3f *acc, Axis3f *gyro, fl
   }
   quadIsFlying = (xTaskGetTickCount()-lastFlightCmd) < IN_FLIGHT_TIME_THRESHOLD;
 
+  static float dx, dy, dz;
+  static float tmpSPX, tmpSPY, tmpSPZ;
+  static float zacc;
+
   if (quadIsFlying) // only acceleration in z direction
   {
     // TODO: In the next lines, can either use cmdThrust/mass, or acc->z. Need to test which is more reliable.
     // cmdThrust's error comes from poorly calibrated mass, and inexact cmdThrust -> thrust map
     // acc->z's error comes from measurement noise and accelerometer scaling
     // float zacc = cmdThrust;
-    float zacc = acc->z;
+    zacc = acc->z;
 
     // position updates in the body frame (will be rotated to inertial frame)
-    float dx = S[STATE_PX] * dt;
-    float dy = S[STATE_PY] * dt;
-    float dz = S[STATE_PZ] * dt + zacc * dt2 / 2.0f; // thrust can only be produced in the body's Z direction
+    dx = S[STATE_PX] * dt;
+    dy = S[STATE_PY] * dt;
+    dz = S[STATE_PZ] * dt + zacc * dt2 / 2.0f; // thrust can only be produced in the body's Z direction
 
     // position update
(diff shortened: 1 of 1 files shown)


src/modules/src/estimator_kalman.c
```diff
@@ -619,35 +619,44 @@ static void stateEstimatorPredict(float cmdThrust, Axis3f *acc, Axis3f *gyro, fl
...
     // acc->z's error comes from measurement noise and accelerometer scaling
     // float zacc = cmdThrust;
-    float zacc = acc->z;
+    zacc = acc->z;
 
     // position updates in the body frame (will be rotated to inertial frame)
-    float dx = S[STATE_PX] * dt;
-    float dy = S[STATE_PY] * dt;
-    float dz = S[STATE_PZ] * dt + zacc * dt2 / 2.0f; // thrust can only be produced in the body's Z direction
+    dx = S[STATE_PX] * dt;
```
stephanbro: Why change these variables to a global scope? They are really only used for readability, as the compiler will likely optimize them out.

stephanbro: http://wiki.c2.com/?GlobalVariablesAreBad

estromb:
I may have gone out on a limb with the static keyword, I admit. The idea behind it was to define it on the heap instead of the stack, since it was redeclared at every loop.

src/modules/src/estimator_kalman.c
stephanbro: Ah I just realized with your comment that these variables aren't global, but just declared static within their scope. Regardless, this is only typically done if the variable's value needs to persist between function calls (the allocation will be .data portion of memory instead of the stack with this change) .

src/modules/src/estimator_kalman.c
```diff
@@ -619,35 +619,44 @@ static void stateEstimatorPredict(float cmdThrust, Axis3f *acc, Axis3f *gyro, fl
...
     // position updates in the body frame (will be rotated to inertial frame)
-    float dx = S[STATE_PX] * dt + acc->x * dt2 / 2.0f;
-    float dy = S[STATE_PY] * dt + acc->y * dt2 / 2.0f;
-    float dz = S[STATE_PZ] * dt + acc->z * dt2 / 2.0f; // thrust can only be produced in the body's Z direction
+    dx = S[STATE_PX] * dt + acc->x * dt2 / 2.0f;
+    dy = S[STATE_PY] * dt + acc->y * dt2 / 2.0f;
+    dz = S[STATE_PZ] * dt + acc->z * dt2 / 2.0f; // thrust can only be produced in the body's Z direction
 
     // position update
     S[STATE_X] += R[0][0] * dx + R[0][1] * dy + R[0][2] * dz;
```
mikehamer: Note that the same bug is present in the else clause further below, and should also be updated. Good catch.

estromb: Ah, thanks. I haven't implemented that case in my own work. I will add a commit for it.

mikehamer:
Just following up on this, these changes seem good.

I suggest you accept the pull request @ataffanel but remove the static keywords from the variables (on :622-:624) since there is no need for them to persist between calls and, I assume, the compiler will therefore more easily be able to optimize the variables away (?)

ataffanel:
Thanks, it is merged!

@mikehamer thanks for reviewing. I agree for static, these variables will likely be optimized by the compiler and never leave the registers.

Issue #170, opened 2016-11-28
Altitude hold not working


ataffanel:
In the current master branch, altitude hold seems to be broken. This seems to come from the work on the position controller (see #136) that broke the control modes contained in the setpoint structure.

stephanbro:
This has definitely been broken as part of the position controller changes, as you suspect. I have a fix I will try to push today.

ataffanel:
I was about to look at it but if you already have a fix that is great! thanks.

ataffanel:
With PR #172 the altitude hold now holds a constant vertical velocity. The current altitude estimator is not outputting vertical velocity which would explain this phenomena.

ataffanel:
39831c3b fixes altitude-hold. I also tried to tune a bit the Z estimator (there was an oscillation between the thrust and the height measurement, fixed by filtering the pressure sensor a bit harder).

PR #172, opened 2016-11-28
Allow velocity only modes and remove const modifier to setpoint in th…

stephanbro:
…e pid controller.

Addresses #170

### DIFF
--- a/src/modules/src/controller_pid.c
+++ b/src/modules/src/controller_pid.c
@@ -32,9 +32,9 @@ bool stateControllerTest(void)
   return pass;
 }
 
-void stateController(control_t *control, const sensorData_t *sensors,
+void stateController(control_t *control, setpoint_t *setpoint,
+                                         const sensorData_t *sensors,
                                          const state_t *state,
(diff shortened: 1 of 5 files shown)


src/modules/src/controller_pid.c
```diff
@@ -51,7 +51,11 @@ void stateController(control_t *control, const sensorData_t *sensors,
   }
 
   if (RATE_DO_EXECUTE(POSITION_RATE, tick)) {
-    positionController(&actuatorThrust, &attitudeDesired, state, setpoint);
+    if (setpoint->mode.z == modeAbs) {
+      positionController(&actuatorThrust, &attitudeDesired, setpoint, state);
+    } else if (setpoint->mode.z == modeVelocity) {
+      velocityController(&actuatorThrust, &attitudeDesired, setpoint, state);
+    }
```
stephanbro: This strategy could be improved by moving this check to the PID controller.

ataffanel:
Thanks! This seems to fix the controller part of things. Though there is still a bug with the height estimation (without the EKF).

PR #174, opened 2016-11-28
Add measurement update with elevation laser ranging in the EKF

mgreiff:
Added support for laser ranging in the EKF such that measurements are taken every 10 [ms] and ququed in the EKF, where they enter the filter by means of a scalar update. The standard deviation of the measurements are set low, and should be made a function of the distance - but this version of the code is stable and has been used to fly with the LPS system.

Edited with a second commit so as not to make pushing into the Kalman filter the default setting.

### DIFF
--- a/src/deck/drivers/src/vl53l0x.c
+++ b/src/deck/drivers/src/vl53l0x.c
@@ -44,6 +44,7 @@
 #endif
 
 //#define UPDATE_KALMAN_WITH_RANGING
+#define RANGE_OUTLIER_LIMIT 3000 // the measured range is in [mm]
 
 static uint8_t devAddr;
 static I2C_Dev *I2Cx;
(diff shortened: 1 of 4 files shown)


src/deck/drivers/src/vl53l0x.c
```diff
@@ -1116,7 +1134,3 @@ static const DeckDriver vl53l0x_deck = {
 
 DECK_DRIVER(vl53l0x_deck);
 
-LOG_GROUP_START(range)
-LOG_ADD(LOG_UINT16, range, &range_last)
-LOG_GROUP_STOP(range)
```
ataffanel: We should not remove this kind of log variables that could be used by other tools.

ataffanel:
Hi Marcus, thanks for the PR!
I tried to activate it by adding ```CFLAGS+=-DUPDATE_KALMAN_WITH_RANGING``` in my config.mk but I cannot really see that it is working: I tried to fly with TDoA and then put my hand close under the CF to make it rise (to verify that the range sensor was used) and nothing happened. How can I verify that the ranging sensor is used?

mgreiff:
Hello Arnaud! I didn't quite know if this was the way to push into the master branch, so do let me know if there is something that I should do differently. I have now revised the code, which now works well and have included more rigorous outlier detection and the laser angle of aperture.

>I tried to activate it by adding ``CFLAGS+=-DUPDATE_KALMAN_WITH_RANGING`` in my config.mk

This doesn't work for some reason, you have to manually uncomment the ``#define UPPDATE_KALMAN_WITH_RANGING`` on line 49 in vl53l0x.c, at which point measurements enter the filter correctly. My hope was to do something like

``KALMAN_RANGING_ENABLE ?= 1
   ifeq ($(KALMAN_RANGING_ENABLE), 1)
   CFLAGS += -UPDATE_KALMAN_WITH_RANGING
   endif``

ataffanel:
Merged, thanks!

PR #175, opened 2016-11-29
Prevent double promotion

stephanbro:
Prevents double promotion to ensure the floating-point unit is used instead of software libraries. Also adds a warning if defines are missing for TDMA.

Depends on bitcraze/libdw1000#3

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -294,6 +294,8 @@ CFLAGS += -Wall -Wmissing-braces -fno-strict-aliasing $(C_PROFILE) -std=gnu11
 CFLAGS += -MD -MP -MF $(BIN)/dep/$(@).d -MQ $(@)
 #Permits to remove un-used functions and global variables from output file
 CFLAGS += -ffunction-sections -fdata-sections
+# Prevent promoting floats to doubles
+CFLAGS += -Wdouble-promotion
 
 # Fail on warnings
 CFLAGS += -Werror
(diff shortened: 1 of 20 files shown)


krichardsson:
Nice improvements!

The build is failing on travis though. I looks as the problem is that travis can not check out the vendor/libdw1000 submodule. All might work out when we merge bitcraze/libdw1000#3, I'm not sure but lets fix libdw1000 first and then address this build problem.

krichardsson:
The git checkout is working OK now that bitcraze/libdw1000#3 has been merged. 
There seems to be a problem with mpu6050.c though. mpu6050 is only used on Crazyflie 1 so to compile the file you must add PLATFORM=CF1 when making.

stephanbro:
@krichardsson, thanks for the info. I changed the build for CF1 and addressed those issues. Now the remaining build issue is due to a warning for TDMA I added here: 533e1d05ab878576ff716447141cefda47d21674.

One way to address this is for me to prevent errors on these warnings, or instead you could possibly update the config.mk used for the build target. Which option would you prefer?

krichardsson:
Great!

I added TDMA_NSLOTS_BITS and TDMA_SLOT as parameters to the TDMA build in the .travis file when I merged. This way we still get the warnings when a user builds the code without the configuration.

Thanks!

PR #178, opened 2016-12-05
Library that implements the MSP protocol (used for OSDs and data loggers)

theseankelly:
Initial implementation of the MSP protocol, targeting a few specific functionalities for BigQuad builds with OSDs (on-screen-displays) running MWOSD firmware.

This implements just enough MSP functionality to tell the OSD that the flight controller is armed, and report attitude readings so that the level bars get updated on the OSD.

Also included some code to initialize and use this library from the big quad's UART connection (the GPS pins), protected behind a compile time define.

### DIFF
--- a/src/deck/drivers/src/bigquad.c
+++ b/src/deck/drivers/src/bigquad.c
@@ -35,6 +35,11 @@
 #include "deck.h"
 #include "extrx.h"
 #include "pm.h"
+#include "uart1.h"
+#include "msp.h"
+
+#include "FreeRTOS.h"
+#include "task.h"
 
 #define BIGQUAD_BAT_VOLT_PIN       DECK_GPIO_MISO
 #define BIGQUAD_BAT_VOLT_MULT      7.8f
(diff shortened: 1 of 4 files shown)


tobbeanton:
Great and many thanks! Now I have to get a OSD to try it out, any recommendations :-)?

theseankelly:
Sure! I'm using this because it's also a PDB and LC power filter built in:
https://www.amazon.com/dp/B01H6MG4OK/ref=sr_ph_1?ie=UTF8&qid=1481037065&sr=sr-1&keywords=osd+pdb

It's been working well but I haven't actually put this quad build through its paces in the air to know how it holds up just yet.

Really, you need anything that can run the [MWOSD firmware](http://www.mwosd.com/), which is just about anything. The following two are pretty tried-and-true in the racing drone community, depending on what you need:

Micro Minimosd - standalone simple, solid OSD. You'll probably want an LC filter if you're doing this with a 3s or higher battery.
https://www.amazon.com/LHI-Micro-Minimosd-Minim-Naze32/dp/B0135SMR72/ref=sr_1_1?s=toys-and-games&ie=UTF8&qid=1481037146&sr=1-1&keywords=micro+minimosd 

OSDoge -- Pricey. I don't quite understand why. I think the main sell is the integrated LC filter and pin compatibility with the Naze32 flight controller family (which we don't care about, do we? :D)
http://www.getfpv.com/osdoge-osd-for-the-naze32.html?utm_source=bing&utm_medium=cpc&utm_campaign=**LP%20-%20NonTM%20-%20FPV%20Equipment&utm_term=OSDoge&utm_content=OSDoge (I don't understand the

tobbeanton:
Thanks for the links Sean! I'll order a miniOSD so I can try it out :-).

PR #189, opened 2017-01-05
Fixing the build when the OSD library is enabled

theseankelly:
in the bigquad deck code. msp.o needs to be
added to the Makefile.


ataffanel:
Thanks! merged.

PR #194, opened 2017-01-30
Disable the PID output constraint for the attitude PIDs (rate/angle) …

### LINKED ISSUE #185: Commander inputs feel delayed/sluggish/strange on latest built firmware (complementary filter)
My CF2 isn't flying well on firmware built with the latest commits. Some symptoms include:
- Sluggishness
- Commander packets randomly feel delayed (anywhere from 0 to 3 second delays before CF reacts to some stick movements)
- Yaw is not right: Hammering the yaw stick full left or right proudces very slow rotation and occasional glitches where the CF rotates backwards (almost like a yaw wrap-around)

I'm flying with the complementary filter and a Devo7e and Devo10 radio (deviationTx).

theseankelly:
…and correct how the integral limit is computed/enforced
Fixes #185

### DIFF
--- a/src/modules/interface/pid.h
+++ b/src/modules/interface/pid.h
@@ -63,7 +63,8 @@
 #define PID_YAW_INTEGRATION_LIMIT     360.0
 
 
-#define DEFAULT_PID_INTEGRATION_LIMIT  5000.0
+#define DEFAULT_PID_INTEGRATION_LIMIT 5000.0
+#define DEFAULT_PID_OUTPUT_LIMIT      0.0
 
 #else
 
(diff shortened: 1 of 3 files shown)


src/modules/interface/pid.h
```diff
@@ -115,8 +117,8 @@ typedef struct
   float outP;         //< proportional output (debugging)
   float outI;         //< integral output (debugging)
   float outD;         //< derivative output (debugging)
-  float iLimit;       //< integral limit
-  float outputLimit;  //< output limit
+  float iLimit;       //< integral limit, absolute value. '0' means no limit.
+  float outputLimit;  //< total PID output limit, absolute value. '0' means no limit.
```
theseankelly: I still don't think it's a great idea to have the outputLimit functionality at all -- if the output of the PID needs to be limited, that suggests the coefficients need to be tuned instead. However, the current position controller makes use of this functionality, and I don't have a way to test out changes to that layer, so I've left it in.

See my comments on issue #185 for some more details.

ataffanel:
Thanks for the PR @theseankelly. I cannot see any difference when flying autonomously in the UWB system so I merged it.

PR #195, opened 2017-01-31
Adjusted spi deck to achieve full speed. Before the spi interface was…

jassi0709:
… turned-off after each byte/communication, while now it is always on.

Tested on my Arducam and this helped to achieve higher frame-rate.

### DIFF
--- a/src/deck/api/deck_spi.c
+++ b/src/deck/api/deck_spi.c
@@ -136,6 +136,8 @@ void spiBegin(void)
   /*!< SPI configuration */
   spiConfigureSlow();
 
+  SPI_Cmd(SPI, ENABLE);
+
   isInit = true;
 }
 
@@ -252,14 +254,14 @@ bool spiExchange(size_t length, const uint8_t * data_tx, uint8_t * data_rx)
   SPI_I2S_DMACmd(SPI, SPI_I2S_DMAReq_Rx, ENABLE);
 
   // Enable peripheral
-  SPI_Cmd(SPI, ENABLE);
+  // SPI_Cmd(SPI, ENABLE);				/*Note(Jaskirat): looks like if you enale/disable each time you have slower performance*/
 
   // Wait for completion
   bool result = (xSemaphoreTake(txComplete, portMAX_DELAY) == pdTRUE)
              && (xSemaphoreTake(rxComplete, portMAX_DELAY) == pdTRUE);
 
   // Disable peripheral
-  SPI_Cmd(SPI, DISABLE);
+  // SPI_Cmd(SPI, DISABLE);				/*Note(Jaskirat): looks like if you enale/disable each time you have slower performance*/
   return result;
 }
 


ataffanel:
Thanks for the PR!

On the shape of it: please remove code for real instead of commenting lines out, since we are using git we can always come back and see what has been changed.

Unfortunately this change freezes on startup when used with the LPS deck.

@mikehamer you wrote the SPI DMA driver, do you remember the reason for disabling and re-enabling the SPI peripheral for every transfers?

mikehamer:
SPI peripheral was enabled and disabled around the DMA transfer in accordance with Section 4.3 of the ST Application Note: [Using the STM32F2, STM32F4 and STM32F7 Series DMA controller](http://www.st.com/content/ccc/resource/technical/document/application_note/27/46/7c/ea/2d/91/40/a9/DM00046011.pdf/files/DM00046011.pdf/jcr:content/translations/en.DM00046011.pdf). In particular:

> If the user enables the used peripheral before the corresponding DMA stream, a “FEIF” (FIFO Error Interrupt Flag) may be set due to the fact the DMA is not ready to provide the first required data to the peripheral (in case of memory-to-peripheral transfer).

I guess this might be the reason for it freezing when using LPS, since there are a lot of memory-to-peripheral transfers (affected by the above). It sounds as if @jassi0709 might only be using peripheral-to-memory transfers (apparently unaffected by the above).

jassi0709:
@ataffanel @mikehamer 
I will debug a bit more, but in make case with the ArduCAM, that edit fix a lot of issues, but I never tried to use the LPS deck. I will do my checks and in few days (~4/5) I will let you know.

jassi0709:
@mikehamer 
You are actually right. Indeed I've found that it's the ArduCAM that create all the issues. The original deck spi seems to work fine but I have do perform a bit rotation on the byte that I read but honestly I have no idea why I need that (and I've found it almost randomly).

I will ask to ArduCAM suppor for some explanation and in the meantime I will perform additional test and, if needed, I will close this merge request.

mikehamer:
Your problem sounds like a clocking or polarity issue. Check that the SPI mode of your device is the same as the Crazyflie's SPI mode (CPOL Low, CPHA Rising, see `deck_spi.c:195`), you will often see this effect if the device is sending/receiving on the wrong clock edge.

ataffanel:
I agree with @mikehamer, looks like a clock polarity issue. Have you tried to change the polarity there: https://github.com/bitcraze/crazyflie-firmware/blob/master/src/deck/api/deck_spi.c#L214-L215?

jassi0709:
Hey guys, sorry for the late reply.
Well it is not a polarity issue since I've already checked that and it looks fine and it does fulfil the datasheet requirements.
I've asked to the ArduCAM support (https://github.com/ArduCAM/Arduino/issues/127) and apparently they were aware of this shift issue with ARM architecture.

To conclude, it looks like the deck spi was working correctly and I'm sorry that I had to waste you time on this...

Best regards,
J

PR #200, opened 2017-02-21
Implementing a rate-based commander packet type

theseankelly:
nearly the same as the standard CRTP commander packet except values for
roll and pitch will always be interpreted as rates instead of angles
(this is 'rate mode' or 'acro mode' in other flight controllers)

Also added code to reset the pitch/roll attitude PIDs if the pitch/roll
setpoints are set to type velocity. This prevents error from
accumulating, which can cause unstable behavior if level mode is
re-engaged during flight

### DIFF
--- a/src/modules/interface/attitude_controller.h
+++ b/src/modules/interface/attitude_controller.h
@@ -54,6 +54,16 @@ void attitudeControllerCorrectRatePID(
        float rollRateDesired, float pitchRateDesired, float yawRateDesired);
 
 /**
+ * Reset controller roll attitude PID
+ */
+void attitudeControllerResetRollAttitudePID(void);
+
+/**
+ * Reset controller pitch attitude PID
+ */
+void attitudeControllerResetPitchAttitudePID(void);
+
+/**
  * Reset controller roll, pitch and yaw PID's.
  */
 void attitudeControllerResetAllPID(void);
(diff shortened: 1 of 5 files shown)


ataffanel:
Hi, Thanks it works great!

I have just one comment: what about setting rates in rad/s instead of deg/s? We have a plan to convert everything in the Crazyflie in S.I. units (including the setpoint structure) and I would like to keep the interface SI as much as possible (while writing this I realized that there is a bug in the velocity control packet, I forgot to convert rad/s to deg/s, I will fix that).

theseankelly:
Sure, I can do that -- will just have to update the deviationTx side to also send radians so I can re-test

theseankelly:
On second thought -- the rate controller expects a degrees input -- how do you want to handle that for this commit? I can convert from rad/s back to deg/s in my commander handler, and then when the PIDs are converted to SI someday that conversion can just be removed?

ataffanel:
Hi, Yes currently it means doing the conversion when receiving the packet, soon we will integrate the non-linear controllers in the master branch and the plan is to convert the setpoint structure to radian by then, then the conversion is going to be removed in the commander and the kalman filter and we will have to either add a conversion to the complementary filter and PID controller of convert them to radian (the PID should be fairly easy to convert to radian).

theseankelly:
I've decided to implement an RC passthrough packet type for sending raw channels from transmitters. That would make this change irrelevant since one of the channels would be a mode switch, and it doesn't really make sense to fly in rate mode unless using a real rc transmitter.

PR #213, opened 2017-03-31
Implementation of a CPPM-like passthrough CRTP packet type

theseankelly:
Values are represented as a set of channels which range from 1000 to
2000 similar to how classic CPPM functions. Defaults to RPYT and has
room for 10 additional optional auxiliary channels. Aux channel 0 is
reserved for selecting between rate and level flight modes.

Also added code to reset the pitch/roll attitude PIDs if the pitch/roll
setpoints are set to type velocity. This prevents error from
accumulating, which can cause unstable behavior if level mode is
re-engaged during flight

### DIFF
--- a/src/modules/interface/attitude_controller.h
+++ b/src/modules/interface/attitude_controller.h
@@ -54,6 +54,16 @@ void attitudeControllerCorrectRatePID(
        float rollRateDesired, float pitchRateDesired, float yawRateDesired);
 
 /**
+ * Reset controller roll attitude PID
+ */
+void attitudeControllerResetRollAttitudePID(void);
+
+/**
+ * Reset controller pitch attitude PID
+ */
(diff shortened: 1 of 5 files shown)


theseankelly:
Will be submitting a PR to deviation that can make use of this feature as well

theseankelly:
The corresponding change went in to DeviationTx

ataffanel:
Sorry for the delay, there has been internal miscommunication: I was waiting for Tobias comment and he though I merged it already.

Anyway, thanks a lot, it looks good and I merge it in. We havent had time to test yet but we will soon.

theseankelly:
No problem -- just wanted to make sure it got in before someone else added a different commander type with value 0x03, since the Deviation change already went in :)

Thanks for looking.

PR #215, opened 2017-04-11
converted uart1.h to unix style line endings

JGSuw:
uart1.h for some reason had CRLF line endings. I have fixed this so that it is has UNIX line endings, same as the rest of the sources in crazyflie-firmware.

I stumbled upon this while adding some basic functionality to uart1, which I would like to create a future pull request for.


ataffanel:
Thanks, I merge the PR.

PR #217, opened 2017-04-12
Removed memset setpoint overwrite

arunlakshmanan:
We were running some tests with the generic message type and realized that the generic decoder was memset-ing the setpoint variable to zeros, which erased any previous changes made to setpoint from [crtp_commander_rpy.c](https://github.com/bitcraze/crazyflie-firmware/blob/master/src/modules/src/crtp_commander_rpyt.c).

### DIFF
--- a/src/modules/src/crtp_commander_generic.c
+++ b/src/modules/src/crtp_commander_generic.c
@@ -247,8 +247,6 @@ void crtpCommanderGenericDecodeSetpoint(setpoint_t *setpoint, CRTPPacket *pk)
 
   uint8_t type = pk->data[0];
 
-  memset(setpoint, 0, sizeof(setpoint_t));
-
   if (type<nTypes && (packetDecoders[type] != NULL)) {
     packetDecoders[type](setpoint, type, ((char*)pk->data)+1, pk->size-1);
   }


ataffanel:
Thanks for the report and PR.

The reset is on purpose and is required for the generic commander to work properly: since the generic commander packets can be send in any order, the full setpoint structure has to be set by each packet decoder and this is made easier if the setpoint structure is reset to a known state each time (at 0).

The problem here is that this setpoint is actually shared between the generic and RPY commander. It was not causing any problems when we tested the code since the RPY decoder always set all required values but if you want to keep the RPY values from one packet to the next this is the root cause of it.

One solution is to always consider the value of the structure as garbage when it enters the decoder and to fill it up completely. This is what the generic and rpy commander decoder are doing today. What is your use-case for keeping the value from one packet to the next?

If there is a compelling use-case for keeping the set-point value between packets, the fix should be in [crtp_commander.c](https://github.com/bitcraze/crazyflie-firmware/blob/master/src/modules/src/crtp_commander.c) and look something like that (I duplicated the static setpoint so that each master decoder has its own):
``` c
static void commanderCrtpCB(CRTPPacket* pk)
{
  if(pk->port == CRTP_PORT_SETPOINT && pk->channel == 0) {
    static setpoint_t setpoint;
    crtpCommanderRpytDecodeSetpoint(&setpoint, pk);
    commanderSetSetpoint(&setpoint, COMMANDER_PRIORITY_CRTP);
  } else if (pk->port == CRTP_PORT_SETPOINT_GENERIC && pk->channel == 0) {
    static setpoint_t setpoint;
    crtpCommanderGenericDecodeSetpoint(&setpoint, pk);
    commanderSetSetpoint(&setpoint, COMMANDER_PRIORITY_CRTP);
  }
}
```

arunlakshmanan:
Hi :) We have manipulators attached to the crazyflie (see this [paper](https://arc.aiaa.org/doi/10.2514/6.2017-1374) and photos [here](http://naira.mechse.illinois.edu/wp-content/uploads/2017/04/Aerial_Manipulation_-_Google_Docs-768x510.jpg) and [here](http://naira.mechse.illinois.edu/wp-content/uploads/2017/04/Aerial_Manipulation_-_Google_Docs-2-768x512.jpg)) which we are attempting to control using the generic message type while sending high-level commands (setpoints, etc.) through rpy. Eventually we may just transfer everything to generic, but we saw this issue when using both at the same time. We can make the suggested changes to our local repo but sharing the same setpoint for both decoders works just fine at the moment. Thanks!

PR #218, opened 2017-04-18
LOG bugfix

jsschell:
logGetVarId("range", "range") returns the id of group start. Better call the log entry something else :)


ataffanel:
The new name makes more sense, I can merge. Thanks for the PR!

PR #221, opened 2017-04-21
Update sensors_bosch.c

jsschell:
critical bugfix: uninitialized variable could cause unexpected behaviour (SensorTask could run with higher frequency)


tobbeanton:
Thanks for the find!

Issue #216, opened 2017-04-11
Mixed line endings appearing in new commits.


JGSuw:
I'm seeing carriage returns showing up in places where they don't belong in recent commits, which will certainly lead to spurious diffs in future pull requests. Perhaps a repository [.gitattributes](https://help.github.com/articles/dealing-with-line-endings/#per-repository-settings) file can be used to correct this issue in the future. Also, it may be worth adding a small section to the readme informing Windows contributors of the CRLF vs LF issue, as they may want to config their git to checkout as-is and commit unix-style in the future. At least they should just be aware of the issue.

ataffanel:
Hi, this sounds like a good idea. So do I understand well that if we add a .gitattribute file git will transparently handle the conversion during commit?

For windows people, I think the line ending setting is a question during installation, so it might not be necessary to repeat it in our readme.

PR #222, opened 2017-04-26
bosch contribution

jsschell:
- updates on bosch sensor drivers
- µSD-Deck Sensor Logging

btw: I don't know what's wrong with the diff of config.h ...


ataffanel:
Thanks a lot for the PR, this looks great!

Maybe there is a problem with line ending between unix and Windows. There has been ticket about this problem, see #216.

On a side note, it would be easier if the two functionality where pushed in two different PR.

PR #225, opened 2017-05-09
µSD-Card logging of sensor data & BOSCH driver update

jsschell:
## µSD-Card logging
Data can be logged with up to 1kHz. in tests this worked good with a
buffer of 100 elements. The buffer is allocated from heap, thus the task
doesn't require a big stack.

HOWTO:
1. create config file with python module in tools/usdlog
2. place config on µSD-Card 
3. make firmware with 'CFLAGS += -DDECK_FORCE=bcUSD'
4. fly around with your crazyflie
5. decode binary log with python module

##  BOSCH driver update & other improvements

* updated drivers
* bias buffer allocated from heap; freed after calibration
  -> increased heap to 40k
  - calibration reset and reallocation is possible
* reorganized code (mostly DRY)


jsschell:
wip: (re-)enable parameters to allow in-fly sensor changes

tobbeanton:
Thanks for the pull request! The logging framework is a great addon! Some issues I bumped in to while testing.
1. The watchdog kicks in if logging is enabled and build with DEBUG=0
2. Would be nice to split the python script in two. One for configuration and one for decoding the logged file.
3. If you run out of heap when logging is setup it would be nice to get a descriptive message in the cfclient console.

jsschell:
1.	I’ve never had any issues with the watchdog. How was your log config? 
2.	Should be possible, also thought about adding an example to plot data
3.	I’ll see what I can do =)

tobbeanton:
1. Ah, I might have gone all in there. I enabled 5 things I think. I have attached the config file.
2. This is just a nice thing to have but it took a while for me to get to a point with usable data. I altered the script to do that, maybe that is not the intended way?
3. =)
[config.zip](https://github.com/bitcraze/crazyflie-firmware/files/995410/config.zip)

tobbeanton:
Nice and great with the debug messages. One thing left. It seams you use tabs in your editor so it doesn't come out aligned. Can you reformat it being tab = 2 spaces?

tobbeanton:
Thanks a lot and nicely done! Next up is some form of documentation. Preferably on the wiki. Up for that ;-)

PR #245, opened 2017-08-07
driver for PCA9685 16-channel I2C PWM chip

jpreiss:
this driver only implements the minimal features of the hardware needed to drive ESCs:

- initialize with a fixed PWM frequency
- set duty or pulse duration
- async task for writes, to avoid blocking stabilizer task

notable missing features include:

- change PWM frequency after init
- sleep/wake
- software reset
- out-of-phase PWM cycles (for reducing peak current when PWM wave directly drives load)
- programmable behavior of OE pin
- "all call" and "sub call" multi-device addressing
- external clock

if other users need these features, let me know...


tobbeanton:
Thanks a lot! We don't have anything to test it with but the code looks really good so we give it an OK :-)

PR #247, opened 2017-08-11
LOCO deck alternative pins for IRQ and RESET

WoNd3RBoY:
Loco deck has an option for alternative pins for IRQ and RESET. This is the implementation that changes to the alternative pins set in the config.mk.


ataffanel:
Thanks! I tested and it still work fine without the flag. I am merging.

Issue #127, opened 2016-06-20
CF2 Syslink UART suffers overrun during high CPU load


mikehamer:
During periods of high CPU load (eg. running a kalman filter or hefty control algorithm), the syslink uart is not always handled fast enough, causing an overrun of the UART data buffer and therefore packet loss.

Bug fix #121 reduces this issue noticeably, however the larger problem is that the CPU is interrupted for every RX byte of the 32-byte packet. This also has an effect on the CPU-time available for tasks requiring significant computation.

This problem (and the overrun problem in general) would be resolved by using DMA for RX transactions, as discussed in #121. The current protocol can be retained by padding the packet to its maximum length (32 bytes).

RX DMA transactions would need to be synchronised. One possibility might be to wait for a correct UART start byte, and then initialise the DMA for numBytes-1. This may only need to be done once, on task startup, to handle the case where the task is started in the middle of a transaction.

Something to look into, perhaps first by profiling the performance hit of constantly interrupting the CPU.

knmcguire:
It seems like this hasn't been an issue for a really long time... so probably the fixes that have been implemented here and there has automatically already fixed this issue. So I will close it, but if you deemed it not fixed please reopen!

mikehamer:
Good call, fixes long since implemented! :-) Good luck with the summer cleanup & greetings to all at Bitcraze!

PR #250, opened 2017-08-25
Move syslink packet parsing into ISR

theseankelly:
Handle the parsing of syslink packets in ISR and post complete packets
to a queue. Client syslink blocks on this queue and is only woken when a
full packet is available.

This eliminates a context switch on every character received and
possibly allows for lowering of the syslink task priority without
risking dropped/corrupt packets.

### DIFF
--- a/src/drivers/interface/uart_syslink.h
+++ b/src/drivers/interface/uart_syslink.h
@@ -30,6 +30,7 @@
 
 #include "crtp.h"
 #include "eprintf.h"
+#include "syslink.h"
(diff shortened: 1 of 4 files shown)


theseankelly:
This is phase #1 of addressing issue #127 - next steps are to look in to lowering the syslink task priority and then to enable partial DMA or full DMA with IDLE detection for packet Rx.

tobbeanton:
Thanks for this Sean!

Issue #237, opened 2017-07-01
Add parameters to check that deck driver have been enabled


ataffanel:
It would be useful for diverse client to be able to check that a deck driver has been enabled or not. 

Today this is done in the PC client checking for the presence of the one-wire memory contained in the decks but being able to check for driver initialization would also allow to check a manually enabled driver.

Doing so with parameters is also easier to use than parsing the one wire memories.

The proposal is to have one read only parameter per driver with the name "deck.<driver name>". The content is a boolean, 1 for initialized, 0 for not initialized.

fredg02:
I noticed that driver initialization seems to be independent from the actual existence of a deck. For example, the deck driver for the zRanger is initialized even when the deck is not actually connected. This is in line with the description in this bug, but not expected by the user, I think

Is this working as intended or is this a bug?.

krichardsson:
@fredg02 , that is not the intention. Only drivers for decks that are detected or drivers that are forced using the compile flag (like ```CFLAGS += -DDECK_FORCE=bcZRanger``` in config.mk) should be initialized.

Can you describe how to reproduce it?

fredg02:
I used the Python client to connect to a CF2 without a zRanger or Flow deck. In the parameters tab the ```deck.bcZRanger``` parameter showed "1", when it should have showed "0".

krichardsson:
What firmware are you running?

krichardsson:
Yes, that sounds like the z-ranger driver has been initialized, the question is why.
I can not see the same behaviour on my CF with (more or less) the same FW. Is the firmware a clean build without any flags set in config.mk?

fredg02:
AFAIK it was a clean build, but I will do some more testing to isolate/reproduce the problem.

fredg02:
OK. So apparently I don't know how visibility of global variables in C works. :(
Deck initialization works as expected, but I was testing a new parameter for the buzzer deck and used the ```isInit``` variable, which is also used by the ```zranger.c``` class. That caused the issue I was seeing. Sorry for the noise.

krichardsson:
@fredg02 The isInit should have been static to avoid interference with other modules. Fixed in #262

PR #260, opened 2017-11-07
Add buzzer deck init parameter (#237)

fredg02:


krichardsson:
Thanks @fredg02 ! 
Can I be a pain in the ** and ask you to add curly braces to the if()? I know we don't have any coding guidelines but I think it is a good practice to always use {}

fredg02:
Sure. I agree that curly braces should be used. I just copied over Arnaud's code from ```zranger.c``` and followed his lead. ;)

krichardsson:
Merged
Thanks!

PR #263, opened 2017-11-10
Add deck parameter for BigQuad, Flow, OA and LED ring deck (#237)

fredg02:
Tested successfully with all mentioned decks except the OA deck (which is not available yet).


krichardsson:
Thanks Fred!

PR #274, opened 2017-12-20
add support for attitute quaternion state estimation in complementary…

whoenig:
… filter

This will avoid potential issues with new controllers while using the complementary filter instead of the EKF as state estimator.


tobbeanton:
A step in a good direction whoenig!

PR #276, opened 2017-12-22
add comments including the units used in datatypes

whoenig:
I found the units used scattered around in the code and hope I didn't make any mistake (please double check). Also, from what I understand pitch is only inverted in state_t.attitude, but not setpoint_t.attitude?

(@jpreiss feel free to take a look if that matches your investigation)


krichardsson:
Thanks!

PR #278, opened 2018-01-05
Port and channels dedicated to GPS receiver exchange of informations

Mitteau:
Two channels to be dedicated to communication with GPS receiver. Useful to avoid mixing of standard console communication and differentiation between NMEA and chip proprietary messages. If not, everything would be mixed. 
Branch nexon_gps
 Validated modifications:
	modified:         src/modules/interface/crtp_localization_service.h

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,6 @@ OPENOCD_TARGET    ?= target/stm32f4x_stlink.cfg
 USE_FPU           ?= 1
 endif
 
-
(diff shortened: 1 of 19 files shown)


tobbeanton:
I think this functionality is highly wanted but the pull request needs to be broken down. It is changing way to many things as it is now. To start with I suggest only doing a pull request related to the update of crtp_localization_service.h and crtp_localization_service.c.

PR #281, opened 2018-01-12
history buffer to control errors on the integral side of the PID

sdalu:
adding history buffer for pid and apply it to the Z axis, making altitude more stable

### DIFF
--- a/src/modules/interface/pid.h
+++ b/src/modules/interface/pid.h
@@ -63,6 +63,16 @@
 
 #define DEFAULT_PID_INTEGRATION_LIMIT 5000.0
 #define DEFAULT_PID_OUTPUT_LIMIT      0.0
+#define DEFAULT_PID_HISTORY_SIZE      16
+ 
+typedef struct
+{
+  uint8_t capacity;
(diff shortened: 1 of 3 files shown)


krichardsson:
Hi!

Thanks for the pull request!

I have tested this and tried to figure out if it improves the situation or not, but it is a bit hard to judge. Do you have a specific scenario that you have tried to improve?

I'm also a bit curious of what the code is actually doing and what the rationale is behind it :-) It looks like you are adding a FIR filter to the I-part in the PID, that is high pass filtering the I-part?
Could you elaborate on this a bit further?

Thanks!
- Kristoffer

sdalu:
I'm using loco and z deck (with only 3 anchors). Trying to make some autonomous flight at constant altitude or keeping position, I found the drone was oscillating a lot on the z-axis. A co-worker suggested to use a buffer to constrain the integral on the few (10-20) last values for the anti-windup instead of having just a limit as in the current code. That seemed to (visually, as I don't have equipment to do proper measurement) improved a lot the stability.

PR #284, opened 2018-01-19
Add a nonlinear trajectory tracking controller.

whoenig:
To test, follow these steps:
1. compile the firmware with ESTIMATOR set to "kalman" and CONTROLLER set to "mellinger".
2. roslaunch crazyflie_demo external_position_vicon.launch uri:=radio://0/100/2M/E7E7E7E728
3. In "crazyflie_ros/crazyflie_demo/scripts": python execute_trajectory.py figure8withTakeoffAndLanding.csv

The controller can also be used with manual flight

### DIFF
--- a/src/modules/interface/math3d.h
+++ b/src/modules/interface/math3d.h
@@ -1,0 +1,624 @@
+#pragma once
+
+/*
+The MIT License (MIT)
+
+cmath3d
+Copyright (c) 2016 James Alan Preiss
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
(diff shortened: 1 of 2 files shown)


ataffanel:
Thanks a lot for the PR! We plan to look at it on Friday (the flight lab is a bit busy right now ... :).

ataffanel:
Thanks, it works fine with a Qualisys system too.
One question: at the end of the trajectory, the CF is held running at (0,0,0) and if I stop the execute_trajectory script with ctrl-C the CF will jump away. Is that something you have observed?

whoenig:
Thanks! I've seen the issue and I am working on a fix for that. I already found the root cause: The controller doesn't handle the "timeout" setpoint sent [here](https://github.com/bitcraze/crazyflie-firmware/blob/master/src/modules/src/commander.c#L82-L94) well. I hope that I can address it in the controller itself without changing the setpoint logic, but I am not sure yet.

whoenig:
Update: I was able to fix the CTRL+C problem in controller_mellinger alone. I put a proposed fix for the issue in another PR: https://github.com/bitcraze/crazyflie-firmware/pull/288. I'll fix the other issue (CF not landing without CTRL+C) in crazyflie_ros.
What changes did you have to do to support Qualisys?

PR #287, opened 2018-01-26
Fix error checking for log/param var lengths

whoenig:
This change fixes undefined runtime behavior in case a developer used variable names for logging that are too long:
1.) The existing checks for both params and logging at startup were incorrect, because they did not look at both group and variable name.
2.) There was no check in logging CMD_GET_ITEM, potentially causing undefined runtime behavior. This change implements the same check as in the parameter framework.

The issue will be detected at startup in all cases. If the debug mode for logging or parameters is enabled, the first variable that violates the condition will be printed.

### DIFF
--- a/src/modules/src/log.c
+++ b/src/modules/src/log.c
@@ -49,6 +49,7 @@
 #include "cfassert.h"
 
 #if 0
+#include "debug.h"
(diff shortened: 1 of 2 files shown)


krichardsson:
Thanks!

PR #288, opened 2018-01-26
ctrlMellinger: Handling of commander timeout

whoenig:
In case the commander didn't receive an update for COMMANDER_WDT_TIMEOUT_STABILIZE, it tries the level the CF.
This change adds the required combination of setpoint modes to handle this case gracefully.

Tested with: manual joystick flight and trajectory following using crazyflie_ros


krichardsson:
Thanks!

PR #289, opened 2018-01-30
Use static functions

sdalu:
Define function `initiateRanging` and `sendLppShort` as `static` (`sendLppShort` was already defined `static` in `lpsTdoaTag.c`)


krichardsson:
Merged. Thanks!

PR #294, opened 2018-02-09
Improved debugger support

whoenig:
This change fixes a number of issues:
 * when using "make flash" it now flashes at the correct address, depending on CLOAD
 * A new command "make flash_verify" can be used to just verify the current firmware
 * "make flash" and "make flash_verify" now use the same (bin)-file as the cloader script
 * Included a template for the use of J-Link debuggers

Tested with ST-Link/V2 and J-Link EDU Mini using openocd 0.10.0


ataffanel:
Thanks, merged!

PR #295, opened 2018-02-09
Fix HardFault in Logging

whoenig:
The logging subsystem could trigger a hard fault caused by unaligned assignment "variable = *(float *)ops->variable;".
This issue can be reproduced with the crazyswarm_master firmware when connecting to the cfclient.

The fix uses lowlevel memcpy to a local (aligned) variable before assignment.

### DIFF
--- a/src/modules/src/log.c
+++ b/src/modules/src/log.c
@@ -541,44 +541,74 @@ void logRunBlock(void * arg)
 
   while (ops)
   {
-    float variable;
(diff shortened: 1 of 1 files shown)


ataffanel:
I though CM-4 allowed unaligned access so this is surprising, but after checking it seems that un-aligned access only work for some instructions.
Thanks for the PR! I merge it.

Issue #293, opened 2018-02-06
High-Level Commander Design


whoenig:
This issue is to discuss the design of a high-level commander (@jpreiss , @mgreiff , @mikehamer).

# Goal
The current [commander](https://github.com/bitcraze/crazyflie-firmware/blob/master/src/modules/src/commander.c) allows for different high-frequency setpoints, such as attitude control, velocity control, or trajectory control. A high-level mode would allow to set high-level goals, while the low-level setpoints are generated on-board. Such a mode can, for example, include:

- Takeoff
- Landing
- GoTo waypoint (where a smooth trajectory is planned on-board)
- Execute Trajectory (that might be built into the firmware, uploaded on demand, or stored on an SD-card)

# Challenges
- The current [generic commander](https://github.com/bitcraze/crazyflie-firmware/blob/master/src/modules/src/crtp_commander_generic.c) can only switch between low-level/high-frequency setpoints. A different layer is needed for high-level commanders.
- A backup pilot should be able to take over control at any point

# Proposed Design
1. A new task for the high-level commander will be created. The task handles new CRTP commands using a new CRTP port. Commands can include takeoff, landing, etc. New commands can be added, similar to the current crtp_commander_generic. The task also provides a function to fill-in the current setpoint.
2. The logic of [commanderGetSetpoint](https://github.com/bitcraze/crazyflie-firmware/blob/master/src/modules/src/commander.c#L74-L95) will be changed such that the high-level setpoint is returned if a) no low-level command was received for a specific time, and b) the high-level commander is enabled (a parameter and perhaps a compile-time flag). Check a) will allow backup pilots to take control. Also, once a low-level commander packet was received we should disable the high-level mode, until it gets explicitly enabled again (by setting a parameter).

We can discuss the actual CRTP packets later; the first step is to agree on how to integrate such a on-board setpoint generator properly. We already successfully use such a high-level commander for the Crazyswarm.

jpreiss:
The interface of the Crazyswarm high-level commander on the firmware side is summarized in [planner.h](https://github.com/USC-ACTLab/crazyflie-firmware/blob/crazyswarm_master/src/modules/interface/planner.h). The PC side is close to a 1:1 wrapper.

The manual backup is the main component missing from our implementation, I think.

ataffanel:
Hi, thanks for starting a ticket about that!

[...]

PR #299, opened 2018-03-07
Add High-Level Commander, see #293

whoenig:
The initial high-level commander supports takeoff, landing, stop, and goTo. Each CF can be part of one of 8 groups, and commands can be broadcasted for specific groups, only.
Tested in a mocap system with kalman estimator and mellinger controller using crazyflie_ros (high-level-setpoints branch; test_high_level.py).

I am planning to add more commands in the future (e.g., trajectory upload and execution), but this PR adds all the necessary components first.


ataffanel:
Thanks a lot! I merge it.

PR #303, opened 2018-03-25
High-level commander: trajectory upload and exec.

whoenig:
This adds the possibility to upload polynomial trajectories using the existing memory interface.
Trajectories can be executed with the high-level commander.
Tested with crazyflie_ros (test_high_level.py).
See issue #293.


ataffanel:
Merged, thanks a lot! Sorry for the late review.

PR #308, opened 2018-04-07
HighLevelCmd: Add missing shift for traj. eval

whoenig:
The evaluation function in the high-level commander ignored the trajectory shift
after a trajectory was completed, resulting in a jump in the setpoint.


ataffanel:
Thanks, merged

PR #319, opened 2018-05-10
Remove clean from compile

acecilia:
In this PR:
* Make the build target run in serie: when using `-j2`, there were cases were the version file was being cleaned after the build, and not before. NOTE: compilation is still done in paralell, which is what gives the speed improvements.
* Avoid cleaning the version when compiling. From the readme:
```
compile    : Compile cflie.hex. WARNING: Do NOT update version.c
build      : Update version.c and compile cflie.elf/hex
```
* Remove unnecesary dependencies between the `compile` target and the `print_version` and `size` targets.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -323,6 +323,7 @@ endif
 
 all: check_submodules build
 build: 
+	# Each target is in a different line, so they are executed one after the other even when the processor has multiple cores (when the -j option for the make command is > 1). See: https://www.gnu.org/software/make/manual/html_node/Parallel.html
 	@$(MAKE) clean_version
 	@$(MAKE) compile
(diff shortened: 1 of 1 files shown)


acecilia:
Also, at them moment, when the `version.c` file is cleared:
* Expectations: if the newly created `version.c` is the same as the old `version.c`, the compiler should finish without regenerating `version.o`
* Reality: even when the newly created `version.c` is the same as the old `version.c`, the compiler regenerates `version.o`. This is not much time (1 second approx), but still is a waste if you consider all the number of builds that can be done while working on the code.

Are the expectations correct, or am I missing something?
In case they are correct, I could not find the way to fix it :(

PR #324, opened 2018-05-25
 Add option to dynamically change estimator using "stabilizer.estimator" parameter

whoenig:
This adds a new parameter ("stabilizer.estimator") such that a user can switch the estimator at runtime, independent of the decks available.


krichardsson:
Thanks!

PR #325, opened 2018-05-29
Add option to switch controllers dynamically

whoenig:
* New parameter: "stabilizer.controller"
* Default controller as before (PID)
* It is now possible to switch at runtime to another
  controller (e.g., mellinger, for the Crazyswarm)


krichardsson:
Great!

Issue #270, opened 2017-12-01
µSD-logging fails in combination with other SPI decks like loco-deck and flow-deck


Woody7777:
Using **both** the µSD-card deck and the loco deck, logging will work only a short time (a few 100 log entries) before it stops. Interestingly, this behaviour seems not to depend on variables logged, sampling rate or buffer used for µSD-config file but solely on time after startup, e.g. with 1000 Hz you'll get ~150 log entries while with 10 Hz you'll get only 20 or so (not exactly reproducible to my experience).

I used the latest firmware and VM2017.03 to reproduce this behaviour. Unfortunately, I don't have a debugger at hand currently. Maybe someone can have a look at the debug prints?

tobbeanton:
I've started to investigate this and my first impression is that the usdWriteTask doesn't get time to run. Will investigate further why and if so really is the case.

ataffanel:
The problem has also been observed with the flow deck: https://forum.bitcraze.io/viewtopic.php?f=2&t=2821

simplexsigil:
I played around with changing the task priorities of the USDLOG_TASK_PRI and USDWRITE_TASK_PRI as well as the priority of the "pamotion" task of the flow deck (it was hard coded to 3).

However, apart from not getting enough runtime, there also appears to be a mutex hold problem.

For the values
USDWRITE_TASK_PRI = 2
"pamotion"-Task-Priority = 1
USDLOG_TASK_PRI = 0

The following error occurs: 
SYS: The system resumed after watchdog timeout [WARNING]
SYS: Assert failed at src/lib/FreeRTOS/tasks.c:3495

With these settings this always happens, with the standard settings it only happens sometimes.

Woody7777:
I made an interesting discovery when playing around with the baudrate for SPI communication: 
As long as it is set to the default (2MBit), the error described in the post before occurs.
However, changing it to something higher, the error disappears but logging still seizes after a couple of log entries - as well when using higher priorities. Still, I was asking myself why we're using the lowest baudrate available?

Unfortunately, a watchdog timeout sometimes occurs when starting up using these settings (with no logged assert). I only have one crazyflie for testing, so I can't verify the behaviour.

Well, **priorities** and **baudrate** are just the most obvious settings that I think may lead to this issue.
I'm sure there are people out there knowing more details regarding this module.
I know it's not easy to instruct someone on issues like that, but I'd be really happy to help... on my own, I'm missing clues where to have a look next :-(

[...]

PR #326, opened 2018-05-30
Generalize logging to uSD card

whoenig:
* Support for any logging variables
* Config file can now be written in a standard text editor
  and contains the list of variables to log (see
  tools/usdlog/config.txt for an example)
* Support for other native datatypes (e.g., uint8_t)
  to avoid memory overhead
* Example scripts are updated to reflect the
  (slightly) modified binary data format

If accepted, I can also update the wiki (https://wiki.bitcraze.io/projects:crazyflie2:expansionboards:microsd#data_logging) accordingly. 
This was also requested by another user in the forum (see https://forum.bitcraze.io/viewtopic.php?f=11&t=2916).


tobbeanton:
This is a very appreciated functionality. Nicely done Whoenig! If you could alter the wiki as well that would be even more appreciated ;-).

Now it reminds me to try and optimize the SPI driver #270.

Issue #313, opened 2018-04-20
Number of Logging variables limited to 256


whoenig:
The number of logging variables is currently limited to 256, because a uint8 is used as internal id in CMD_GET_INFO and CMD_GET_ITEM (see https://github.com/bitcraze/crazyflie-firmware/blob/master/src/modules/src/log.c#L238-L289). This is an issue, because the current firmware already uses ~240 logging variables.

An obvious fix is to switch to uint16 instead, but the question is on how to do that in a version-safe way for the various clients. Perhaps we can:
 * Add CMD_GET_INFO_V2 and CMD_GET_ITEM_V2 that use uint16
 * Let clients implement the new version
 * Remove the old CMD_GET_INFO and CMD_GET_ITEM functions in a later firmware version

PR #337, opened 2018-06-23
Allow more than 256 logging vars/parameters

whoenig:
This addresses the firmware side of issue #313 by adding additional
CRTP commands (called V2) that uses 16 instead of 8bits for indexing.

This change keeps the old API the same, but caps the number of vars
to 255 to allow older clients to keep working. Eventually, the old
API should be removed.

This change reduced the possible length of variables by 1 byte,
which affected two variables whose name had to be shortened.

Tested with crazyflie_cpp commit 3eb3203. Other clients will need
similar changes to use more than the first 255 variables.

### DIFF
--- a/src/config/config.h
+++ b/src/config/config.h
@@ -46,7 +46,7 @@
 #include "trace.h"
 #include "usec_time.h"
 
-#define PROTOCOL_VERSION 3
+#define PROTOCOL_VERSION 4
 
 #ifdef STM32F4XX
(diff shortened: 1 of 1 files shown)


whoenig:
Thanks @ataffanel! I added a new issue for crazyflie_cpp to use the protocol version to distinguish old/new firmwares, see https://github.com/whoenig/crazyflie_cpp/issues/6.

Issue #335, opened 2018-06-19
Controller mellinger yaw direction reversed


kyuhyong:
I tried to compare using PID and Mellinger for controller and found their yaw direction is opposite. 
In cfclient, right yaw commands gives positive deg/s while left yaw gives negative deg/s
with PID controller, cf2 turns right when I commanded right yaw however
with Mellinger controller, cf2 turns left for the same command.
Is this intentional? 
I tried to modify some code in the mellinger.c but changing the input does not help solving this issue. 
Thanks,

Kyu

krichardsson:
I think I'll pass this one on to @whoenig
Thanks!

whoenig:
This is not intentional and sounds like a bug to me. The mellinger controller was more heavily tested for operation in a mocap, so this might be related to the slightly separate manual flight mode. I'll take a look.

kyuhyong:
I applied the fix and tested with my drone today. 
It seems the issue is gone with both CONTROLLER: Using controller 2 or 1
Thanks,
Kyu

PR #338, opened 2018-06-23
Manual flight: same yaw rate for PID and Mellinger

whoenig:
The PID and mellinger controllers had different behaviors (inverse
yaw direction) in manual flight mode, see issue #335.  The root cause
is that the legacy yaw input is actually inverse (with respect to
the CF coordinate system). This change changes the sign accordingly
before processing and fixes some other math issues related
to yaw control.

Tested with manual flight using PID and Mellinger controllers.

### DIFF
--- a/src/modules/src/controller_mellinger.c
+++ b/src/modules/src/controller_mellinger.c
@@ -179,9 +179,9 @@ void controllerMellinger(control_t *control, setpoint_t *setpoint,
     }
   }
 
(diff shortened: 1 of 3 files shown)


ataffanel:
Thanks for the PR. I added a commit to fix the yawrate sign in the commander_generic packets as well since it caused problem for assisted modes in the client (like for flying with the flow deck)

PR #341, opened 2018-06-28
Fixed default address of VL53L1X driver from 0x51 to 0x52 to reflect …

JGSuw:
Fixed default address of VL53L1X driver from 0x51 to 0x52 to reflect the actual default address as reported in the datasheet.

### DIFF
--- a/src/drivers/interface/vl53l1x.h
+++ b/src/drivers/interface/vl53l1x.h
@@ -39,7 +39,7 @@ extern "C"
 {
 #endif
 
-#define VL53L1X_DEFAULT_ADDRESS 0b0101001
+#define VL53L1X_DEFAULT_ADDRESS 0b01010010
 
 #define USE_I2C_2V8
 


ataffanel:
I assume this was not working?
I2C addresses are always confusing: when sent to the bus there is a 7 bit address followed by 1 bit direction. Some system use 8 bits addressed with the direction set to 0 (like the ST doc), some use the 7bit address (like the stm32 peripheral lib). This should explains the inconsistency.

JGSuw:
Thank you Arnaud,

Yes this was caused some head-scratching for a colleague and I. We didn't
get around to testing it, but rather when I learned that the VL53L1X and
VL53L0X have the same address, I checked the headers used in the z_ranger
deck, lo and behold the address was the same. A silly mistake.

By the way, would you happen to know when the multi-ranger deck will be
available? I ask because I'm working on a project right now and I would
like to use one of the VL53L1X sensors as a lateral range finder. For the
time being, I will probably just throw together an i2c mux interface board
that does the job, but I am really excited to get my hands on multi-ranger!

Regards,
Joseph Sullivan

PR #347, opened 2018-07-04
Fixed wrong assertions

bbingju:
I modified wrong assertions which is possible to cause a hard fault.


ataffanel:
Thanks for catching it! Merging.

PR #351, opened 2018-07-26
HLCmd: Avoid setpoint discontinuities

whoenig:
The previous HLCmd frequently planned from the current state
rather than the current setpoint. This can cause jumps in the
setpoint, and, in case of a swarm, inconsistent behavior such
as different flight heights across the CFs.


ataffanel:
Thanks! Merged.

Issue #352, opened 2018-08-02
Compiler error with newer compiler


whoenig:
Compiling the firmware with the following compiler version:

```
$arm-none-eabi-gcc --version
arm-none-eabi-gcc (GNU Tools for Arm Embedded Processors 7-2018-q3-update) 7.3.1 20180622 (release) [ARM/embedded-7-branch revision 261907]
Copyright (C) 2017 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

Leads to a compiler error:
```
In file included from src/modules/src/crtp_commander_generic.c:35:0:
/usr/arm-none-eabi/include/math.h:133:12: error: expected identifier or '(' before 'double'
 extern int isnan (double);
            ^
src/utils/interface/num.h:41:28: error: expected ')' before '?' token
 #define isnan(n) ((n != n) ? 1 : 0)
                            ^
tools/make/targets.mk:26: recipe for target 'crtp_commander_generic.o' failed
make[1]: *** [crtp_commander_generic.o] Error 1
Makefile:330: recipe for target 'build' failed
make: *** [build] Error 2
```

The issue is that isnan is part of math.h (see https://en.cppreference.com/w/c/numeric/math/isnan), but it is redefined in num.h.

PR #353, opened 2018-08-02
Remove max,min,isnan macros in favor of built-ins

whoenig:
This change removes the max,min and isnan macros
in num.h. The min, max macros are replaced by
fminf and fmaxf, avoiding side-effects that can
be caused by macros.

isnan is replaced by the compiler-provided
macro (part of math.h).

This fixes issue #352.


krichardsson:
Thanks!

PR #371, opened 2018-09-23
Non static estimator kalman

acecilia:
This PR provides with:
* A non static implementation of the kalman filter based on the existing implementation:
    - Now it is possible to instantiate multiple filters. 
    - It uses a more Object Oriented API.
    - It uses a simpler implementation of FIFO queues (used to pass data to the filter), removing the dependency on FreeRTOS. Also, this queues are statically allocated (they can also be dynamically allocated, if that is something interesting).
    - This implementation does not process sensor information (accelerometer/gyroscope), but works well when providing distance or position data. Providing sensor information should be easy to implement by using the new FIFO queues (famous last words).
    - It simplifies the implementation of the prediction step (`static void predict(estimatorKalmanStorage_t* storage, Axis3f* acc, Axis3f* gyro, float dt)` function), removing [the hacky way of checking if the quadcopter is flat](https://github.com/bitcraze/crazyflie-firmware/blob/master/src/modules/src/estimator_kalman.c#L700).
    - It further simplifies the prediction step, extracting the covariance update step to its own function: `static void introduceQuadcopterDynamicsIntoCovariance(estimatorKalmanStorage_t* storage, Axis3f* gyro, float dt)`.
* Tests for the kalman filter using ceedling/cmock (the test suit in place).

This is work in progress. In my opinion, this would be ready to go when the changes done to https://github.com/geekfactory/FIFO are merged to master (at the moment the FIFO repository being used is a fork with some additions, ex: PR in https://github.com/geekfactory/FIFO/pull/2).

Is this something interesting for you, or I should better keep it in a different repository?


krichardsson:
Thanks! Will take a look after IROS

krichardsson:
I like this change but I have not had the time to look at it. We will most likely refactor the kalman estimator at some point and do more or less what you have done. Unfortunately the code base has changed and I don't feel I have the time to handle this, so I'll close the PR without merge. Thanks for your contribution!

PR #408, opened 2019-03-18
Simplify logic for external position information.

whoenig:
Tested with the Crazyswarm using a VICON Mocap.


ataffanel:
Looks good to me, merging. Thanks!

PR #411, opened 2019-03-22
Add getDeviceTypeName to platform service.

whoenig:
This will allow us to change the client behavior based on the device type.
Tested with CF 2.0, CF 2.1, RR 1.0 using crazyflie_tools/version.


ataffanel:
This is a good idea, will be useful. Thanks!

PR #412, opened 2019-03-24
uSD: Allow comments in config.txt

whoenig:
This change allows comments (#) and whitespace in config.txt
used for logging data on a uSD card. This will be in particular
useful as we add more features to config.txt in the future.


ataffanel:
Looks great, thanks! Merging.

PR #413, opened 2019-03-25
Add status messages in stabilizer for sensor calibration

whoenig:
This helps debug cases where the CF is not taking off, because
the sensors were not (yet) successfully calibrated without the need
of analyzing the blinking pattern of the LEDs.


tobbeanton:
Simple and informative! :-)

PR #414, opened 2019-03-25
uSD card: Add synchronous logging mode.

whoenig:
The synchronous logging mode can log variables synchronously in the
stabilizer loop. This helps to analyze flight crashes etc. where
the order and consistency between variables is crucial. The older
(asynchronous mode) is retained, although the config file format
changed slightly.

Tested on a CF 2.0 (both synchronous and asynchronous modes using
the config.txt that is part of this repository.)

### DIFF
--- a/src/deck/drivers/interface/usddeck.h
+++ b/src/deck/drivers/interface/usddeck.h
@@ -4,23 +4,23 @@
 #include <stdint.h>
 #include <stdbool.h>
 
-typedef struct usdLogDataPtr_s {
-  uint32_t* tick;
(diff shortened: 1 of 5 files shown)


tobbeanton:
Great functionality. Our dream would be to be able to configure and trigger it from the cfclient, maybe one day :-)

tobbeanton:
Out of curiosity I checked the task-load and the USDTASK is taking around 40% when testing the config.txt file in synchronous mode. Since the IDLE task is also at 40% there shouldn't be a problem but might be worth to have this in mind.

PR #419, opened 2019-04-01
Add support for DEBUG_PRINT using SEGGER RTT

whoenig:
Segger Real Time Transfer (RTT) is a high-speed protocol that can
be used with JLink debug probes. It allows to exchange data much
faster than radio or uart by using shared memory and SWO. It can
be used for debugging (printing inside interrupt routines works fine)
or sensor data collection (as long as the CF is connected via debug
cable).

Tested with JLink EDU Mini and CF2.0.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -227,6 +227,12 @@ PROJ_OBJ += configblockeeprom.o crc_bosch.o
 PROJ_OBJ += sleepus.o
 PROJ_OBJ += pulse_processor.o lighthouse_geometry.o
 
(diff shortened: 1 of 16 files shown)


whoenig:
Note that the license specifically allows redistribution, as long as the license file is kept (which I did). I find RTT very useful for development/debugging and even the EDU Mini supports RTT (which is cheap enough to be used by educational institutions.)

ataffanel:
Thanks for the PR. License-wise this looks like a classical 3-Clause BSD license which is alright.

PR #420, opened 2019-04-03
Add logging variables for compressed state and setpoint.

whoenig:
This will allow to create a log block that can capture position,
orientation (and if desired vel/acc) in a single, time-syncronized
block. That is important for accurate pose reporting (e.g., for
roadrunner, or for CFs that are not controlled with external
mocap).

The compressed setpoint allows to capture both estimated and
desired pose, which is important to compute metrics such as
tracking errors.

### DIFF
--- a/src/modules/src/stabilizer.c
+++ b/src/modules/src/stabilizer.c
@@ -48,6 +48,7 @@
 
 #include "estimator.h"
 #include "usddeck.h"
+#include "quatcompress.h"
 
 static bool isInit;
 static bool emergencyStop = false;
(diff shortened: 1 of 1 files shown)


ataffanel:
Thanks, merged.

I guess this would allow the Crazyflie ROS driver to create a TF frame for the Crazyflie? I want to completly retire the lps-ros repos and as far as I understand this is the only missing functionality to Crazyflie-ros to use it with LPS/Lightouse/....

whoenig:
Yes, the goal here is to have the missing features from lps-ros directly in crazyflie_ros, which hopefully reduces maintenance for all of us in the future:-) I already implemented the crazyflie/pose topic, but I can also add the population to tf for convenience. Then, the only reason for people to use lps-ros would be to use an external state estimator (such as the particle filter).

PR #421, opened 2019-04-04
Update math3d library to the latest version.

whoenig:
94f08dd from https://github.com/jpreiss/cmath3d

This includes some more consistent naming as well as avoiding
some unnecessary double promotion.


krichardsson:
Thanks!

PR #438, opened 2019-06-25
Support to read latest logging file from uSD card over radio

whoenig:
This change adds a new memory type (USD), which maps to the last
logging file on the microSD card. The file can be read over the
radio if logging is disabled (i.e., usd.logging is set to 0).

Tested with downloadUSDLogfile from crazyflie_tools with and
without the uSD-card deck using a CF2.0.


krichardsson:
Thanks!

Issue #246, opened 2017-08-09
Implement full pose measurement in EKF


jpreiss:
The current EKF implementation supports a position measurement à la GPS, but no full pose measurement. In a motion capture system the orientation measurement is also available. Including this information can only improve the state estimation.

I am happy to do the needed work on the radio protocol and the ROS driver, but it would be nice to get some support from an `estimator_kalman.c` author for the EKF changes.

ataffanel:
As far as I understand the full pose input implementation in the current EKF is not obvious. Though, I think this kind of interface should be (as far as possible) independent of any particular algorithm anyway so feel free to propose an interface.

Issue #203, opened 2017-03-06
Packet for externally measured position and orientation


pittersdb:
Hi,

Considering the current implementation that allows to send measured position from client to a crazyflie using cf.extpos.send_extpos via a packet to the EKF, it is needed a way to send orientation as well, in order to consider the measured yaw and start the CF front facing on any direction more than only the X when it take-off.

pittersdb:
Hi @ataffanel , 
I have been checking the firmware code to have an idea about how the data (x,y,z) flows from client to the EKF for performing estimation, I think that I can understand how an adicional parameter (like yaw)can be received; however, I can realize that (x,y,z) are processed to obtain an updating of the covariance matrix. How the external measured yaw can be included in that process ?, Do you think you can help us with this implementation?. Our goal is that several CF can be positioned in space (using a mocap ) and produce swarm algorithms to lead those drones to make a coordinated task, our focus would be in the algorithms assuming that the CF can be positioned in a point in space, however we have this current problem in positioning (because orientation is not considered)

pittersdb:
Hi,
@ataffanel suggested me to expose this issue to @mgreiff.
Before to start this issue, I moved the coordinate system to get always the front of the CF facing to the X axis from mocap, however, I thought that there would be no problem if always (even while flying) this solution were applied, but it doesn't work at all. What I can realize is: if the CF change the front direction while is flying, the CF fails and turns over. I supposed that, considering that EKF works if always the front is facing X, when the yaw change because of a movement, it cannot restore the init yaw(which faces X) in which it starts, it is true ?, If that is the case then I think the controller should keep the init yaw (the setpoint yaw would be zero), and when the setpoint yaw is different from zero, the CF should rotate in yaw and update what is needed to be adjusted to the new orientation, this suggestion makes sense ?. Something that I would like to do is that a CF follows a trajectory and change the yaw while is flying, is that possible ?

PR #445, opened 2019-07-16
Implement full pose measurement in EKF.

whoenig:
This adds two new messages to the crtp_localization_service to
receive external pose information (compressed and not compressed)
such as information obtained by a motion capture system. (Fixes
issue #203).

It also adds a full pose update in the EKF using a scalar update
for both position and orientation errors. (Fixes issue #246).

Tested with the Crazyswarm using a single Crazyflie and a VICON
motion capture system.


krichardsson:
Thanks!

PR #451, opened 2019-08-02
Add support to control the yaw angle (and not only rate).

whoenig:
Note that stabilizationModeYaw is in the firmware already, but
was not used until now.


krichardsson:
Thanks!

PR #457, opened 2019-08-22
fix compilation of unit tests on macOS

ntamas:
This PR fixes the compilation of unit tests on macOS where the compiler does not seem to be happy with the original `__attribute__((section(...)))` tag and insists on formatting it the way I modified it in the PR.

Note that the PR does not change anything when the firmware is compiled with `arm-none-eabi-gcc`; the `__APPLE__` macro is defined only when compiling _for_ macOS and not cross-compiling for another platform.

### DIFF
--- a/src/modules/interface/log.h
+++ b/src/modules/interface/log.h
@@ -74,9 +74,15 @@ struct log_s {
    { \
   .type = TYPE, .name = #NAME, .address = (void*)(ADDRESS), },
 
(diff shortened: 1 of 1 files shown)


krichardsson:
I was actually annoyed by this problem earlier today and thought that I maybe should take a look at it, but I didn't, and now you have fixed it. Awesome! 

Thanks a lot!

PR #460, opened 2019-08-22
LED Ring - add lighthouseRateEffect

NicksonYap:
With full reception / max quality, it is expected to receive 120 position samples per second (posRt)

0 is the worst quality possible

LED Ring will be Red if 0, Green if 120

Tested to work, though the reaction time is a bit slow since the rate counter can only work at 1 second intervals (tried to lower, it seems the rate is not uniform so it'll produce unstable rate counts)

### DIFF
--- a/src/deck/drivers/src/ledring12.c
+++ b/src/deck/drivers/src/ledring12.c
@@ -683,16 +683,30 @@ static float badRssi = 85, goodRssi = 35;
 static void rssiEffect(uint8_t buffer[][3], bool reset)
 {
   int i;
-  static int rssiid;
+  static int isConnectedId, rssiId;
   float rssi;
+  bool isConnected;
 
-  rssiid = logGetVarId("radio", "rssi");
-  rssi = logGetFloat(rssiid);
+  isConnectedId = logGetVarId("radio", "isConnected");
+  isConnected = logGetUint(isConnectedId);
+
+  rssiId = logGetVarId("radio", "rssi");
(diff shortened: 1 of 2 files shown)


NicksonYap:
@krichardsson it seems I need to create a temporary branch for each PR?

My PR gets updated immediately after I push to my own `master`

krichardsson:
It seems as github recommends using branches for PRs https://help.github.com/en/articles/about-pull-requests
Not sure what is happening

krichardsson:
Thanks for the pull request!
I'll wait to merge this since it is based on the current lightouse implementation that most likely it will change.

NicksonYap:
Yes, I'm currently working out the maths and algorithm for calculating position from any combination of sensor detections and base stations receptions.

The concept will be similar as we'd still like to know the rate at which position data is obtained.
Just that not sure if the max rate is still 120Hz

ataffanel:
Lighthouse V1 runs at 120Hz with TDMA: so each sensors will receive an angle measurement at 120Hz but we have all 4 angles from both lighthouse at 30Hz. This is a constant in a lighthouse V1 system.

Lighthouse V2 runs at ~50Hz and provides 2 angles per turns (so it is 2 angles at 50Hz). There is no TDMA so each lighthouse in sight will add 50Hz of data.

NicksonYap:
Single basestation is working, the rates will be halved so I will close this until single BS is merged

PR #466, opened 2019-09-08
LED RSSI - turns white if disconnected (1 sec timeout)

NicksonYap:
Previously LED RSSI effect will freeze color if disconnected (1 sec timeout)

If connection lost, This will make it turn white instead

If regained, it will be back between green and red


krichardsson:
Thanks!

Issue #444, opened 2019-07-14
get_bs_position.py errors out in Python 3.7


mkohler99:
Trying to calculate position of lighthouse base stations on 2 different machines, one with Python 3.6.4 and one with 3.7.4. On the machine with 3.7.4 the script errors out and claims it cannot decode the string in line 52. Works fine on 3.6.4. 

Seems like its maybe related to this? 

https://vstinner.github.io/python37-new-utf8-mode.html

Removing the .decode() call seems to make it work fine.

PR #467, opened 2019-09-08
get_bs_position - fix decode("utf-8")

NicksonYap:
Solves: https://github.com/bitcraze/crazyflie-firmware/issues/444


krichardsson:
Thanks!

Issue #258, opened 2017-11-03
Crash/tumble detection should not re-enable motors when flipped back


fredg02:
Crash/tumble detection (#248 ) is really helpful and prevents damage, but I think it can be improved a little bit. It's still potentially dangerous to re-enable the motors when the copter is flipped back up. If setpoints are still sent (e.g. a flow sequence is still running), the motors can start spinning again violently. This can cause (minor) injuries or lead to another crash, when the copter is dropped/flies away.

It would be best to cancel the running sequence or not accept setpoints until some kind of "reset"-packet is sent. Since this would be rather hard to implement (a flow sequence is most likely sent by a client and can't be canceled on the copter side), I'd recommend to kill the connection. There might be use-cases where this is inconvenient, but it's the safest option IMHO.

krichardsson:
Agree, some form of freeze mode that requires a reset to fly again.

ntamas:
A less intrusive option to the current codebase would be to disable the high-level commander when a tumble is detected. This would not solve the problem when a client is continuously setting new setpoints, but it would solve the cases when the Crazyflie is flying a pre-programmed sequence autonomously with the high-level commander.

I think that a patch like this would essentially need to add a function to `commander.c` to disable the high-level commander, and the situation awareness module would need to call this function when a tumble is detected. Would you accept such a PR?

krichardsson:
Pull requests are always welcome :-)
Another option could be to simply turn off the motors, regardless of using the HL commander or not.

ntamas:
This seems to do the trick for me; if I understand correctly the thrust is already set to zero, which should turn off the motors, all that we need is to ensure that the HL commander does not turn them back on when the drone is picked up from the floor:

```patch
diff --git a/src/modules/src/sitaw.c b/src/modules/src/sitaw.c
index 87a9707..a3699a2 100644
--- a/src/modules/src/sitaw.c
+++ b/src/modules/src/sitaw.c
@@ -32,7 +32,7 @@
 #include "param.h"
 #include "trigger.h"
 #include "sitaw.h"
-#include "commander.h"
+#include "crtp_commander_high_level.h"

 /* Trigger object used to detect Free Fall situation. */
 static trigger_t sitAwFFAccWZ;
@@ -138,6 +138,9 @@ static void sitAwPreThrustUpdateCallOut(setpoint_t *setpoint)
         setpoint->mode.y = modeDisable;
         setpoint->mode.z = modeDisable;
[...]

PR #470, opened 2019-09-13
Cut motor thrust permanently after a tumble

ntamas:
This PR is related to #258 and the discussion therein. It extends the current tumble detector such that:

* it stops the high-level commander from sending setpoints
* it preserves the "tumbled" situation until the operator explicitly marks it as resolved by resetting the state of the tumble detector with a designated parameter

I'm not fully satisfied with the PR yet; the current implementation would also detect a tumble if you hold the Crazyflie upside down for a while before takeoff, and would block the takeoff itself. I'm not sure what the best way would be to address it - we could either extend the tumble detector in `sitAwTuTest()` to also require significant motor thrust (not only a significant roll / pitch angle), or check the motor thrust in `sitAwPreThrustUpdateCallOut()` and don't set the `sitAwMotorThrustCut` flag if the motors are not running. Suggestions are appreciated.

### DIFF
--- a/src/modules/src/sitaw.c
+++ b/src/modules/src/sitaw.c
@@ -32,7 +32,7 @@
 #include "param.h"
 #include "trigger.h"
 #include "sitaw.h"
-#include "commander.h"
+#include "crtp_commander_high_level.h"
 
 /* Trigger object used to detect Free Fall situation. */
 static trigger_t sitAwFFAccWZ;
@@ -42,6 +42,12 @@ static trigger_t sitAwARAccZ;
 
 /* Trigger object used to detect Tumbled situation. */
 static trigger_t sitAwTuAngle;
+
+/* Flag to denote whether we have cut the motor thrust because of a tumbled situation earlier. */
+static bool sitAwMotorThrustCut;
+
+/* Flag that the user can set to 1 to mark a tumbled situation as resolved. */
+static bool sitAwResetRequested;
 
 #if defined(SITAW_ENABLED)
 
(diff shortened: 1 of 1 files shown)


krichardsson:
What about keeping this functionality passive until taking off and activating it when leaving the ground?
Not sure how to detect that we are flying though. One suggestions could be to look at the thrust to the motors as is done in the kalman estimator https://github.com/bitcraze/crazyflie-firmware/blob/master/src/modules/src/estimator_kalman.c#L300.

whoenig:
1. I think it might be better to actually use the existing flag for an emergency here, rather than adding a new one: https://github.com/bitcraze/crazyflie-firmware/blob/master/src/modules/src/stabilizer.c#L303. This way we can deal consistently with resetting from various emergencies. There is currently no CRTP packet to reset the emergency, but that could be added in the future, or we can simply expose it as parameter.

2. As for the flight detection, I think in this case you can check if the current controller output is non-zero, as suggested. One could either pass in the current control-struct pointer to the sitAw framework (similar to the kalman filter), or use pointers to existing logging variables (similar to the led-ring deck).

3. (slight sidetrack from this issue). I found that the current tumble detector does not work reliably when using the EKF, because it relies on the current state estimate, which in case of a crash might or might not reflect the "upside-down" state. I suggest that we change the current check, or add another condition, that checks for sensorData->acc.z < -0.5f instead. We use this check in the crazyswarm firmware (in a hacky way, see https://github.com/USC-ACTLab/crazyflie-firmware/blob/crazyswarm-next/src/modules/src/stabilizer.c#L308-L317), and it has prevented many permanent motor/prop damages after crashes.

krichardsson:
Closing since this was implemented in PR #482 
Thanks @ntamas for your effort!

ntamas:
Thanks and sorry for the lack of response; I would have needed some more time to finish this off, but I'm glad that there's a proper solution now!

(For the record, @krichardsson meant PR #481, not PR #482).

PR #478, opened 2019-09-27
Refactor planner to avoid global static members

whoenig:
This simplifies using the same code in other C-projects as well
as creating Python bindings using SWIG. The latter is used in the
Crazyswarm simulator for more accurate (and efficient) simulation.


krichardsson:
Thanks!

PR #480, opened 2019-10-03
PULP-Shield/PULP-DroNet integration

DP8:
Hi,
I introduced in the firmware all the modifications/extensions to make it work with the PULP-Shield (pluggable PCB) and the PULP-DroNet CNN for autonomous navigation.

All the best,
Daniele

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,6 @@ bin/*
 .settings/*
 .vscode/*
 *~
-tools/make/config.mk
 cflie.*
 version.c
 tags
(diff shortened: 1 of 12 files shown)


knmcguire:
The checks have failed for this pull request. Moreover, we will have the integration of the AI-deck soon in the firmware which is based on the pulp-shield, but uses a different port for deck communication. The autonomous navigation can then be achieved by using the new app-layer functionality to keep the firmware more generic. But let's wait until the AIdeck has been fully integrated.

PR #482, opened 2019-10-04
Avoid compilation error on newer toolchains (2019-q3).

whoenig:
This change temporarily disables a warning (that leads to a
compiler error in release builds) in USB_OTG_WritePacket and
USB_OTG_ReadPacket.


krichardsson:
Thanks!

PR #491, opened 2019-10-29
Update types for parameters in i2cdrvCreateMessage and i2cdrvCreateMe…

markhermeling:
…ssageIntAddr

The type for direction in both these differ in the .c from the .h. Sensitive parsers don't like it.


krichardsson:
Thanks!

PR #492, opened 2019-10-30
Uninitialized variable in skip_to_next

markhermeling:
steps needs to be initialized to 0 or undefined behavior may happen. Found during static analysis by GrammaTech CodeSonar


krichardsson:
Thanks!

PR #498, opened 2019-11-01
Uninitialized variable if queue is empty

markhermelinggt:
xQueueReceive can return xQueueReceive, in which case work should not be used.

This is a pattern that is repeated in other places as well. If people consider this valuable enough to fix, then I will continue to find the other locations and issue a pull request combining them.

### DIFF
--- a/src/modules/src/worker.c
+++ b/src/modules/src/worker.c
@@ -65,10 +65,8 @@ void workerLoop()
     return;
 
   while (1)
-  {
-    xQueueReceive(workerQueue, &work, portMAX_DELAY);
-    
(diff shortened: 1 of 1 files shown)


ataffanel:
Thanks for your PR!

Since the timeout for this call is portMAX_DELAY, it is guaranteed by the OS that the call will not return until there is a data available (see https://www.freertos.org/a00118.html). So as long as this call is made with portMAX_DELAY as timeout it is not a bug.

So I am closing this without merging. If you find any case where the return value is not checked while the timeout is not portMAX_DELAY, feel free to post a PR for those.

PR #501, opened 2019-11-05
xQueuePeek with third parameter '0' does not initialize the 2nd param…

markhermelinggt:
…eter

If the queue is empty, then priority in commanderSetSetpoint is not initialized. The same holds for commanderGetSetpoint (the XQueuePeek would not initialize set point on line 81 if the queue is empty, but it is not clear to me how to fix that code.

@ataffanel : This is a similar problem as I reported earlier. Unclear to me if this queue can ever be empty.

### DIFF
--- a/src/modules/src/commander.c
+++ b/src/modules/src/commander.c
@@ -66,9 +66,9 @@ void commanderInit(void)
 void commanderSetSetpoint(setpoint_t *setpoint, int priority)
 {
   int currentPriority;
-  xQueuePeek(priorityQueue, &currentPriority, 0);
 
-  if (priority >= currentPriority) {
+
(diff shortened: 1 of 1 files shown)


krichardsson:
Hi!

This queue should never be empty since an element is added in the commanderInit() and elements never are removed. However the queue is currently not defined as static, which it should be to make sure no other code outside the file is tampering with it, so this is not obvious.
It is probably a good idea to check the return value of the peek call, but since the queue should never be empty (by design), it might be better to check the return value using an assert to indicate that this is an unexpected event and fail fast, as opposed to handle the error.

I will merge the pull request, but make some modifications in a second commit. 

Thanks for finding dodgy code and your effort!

PR #504, opened 2019-11-12
Added compressed trajectory format

ntamas:
This PR adds support for an alternative, more space-efficient trajectory representation in the trajectory memory. (This is the one that Gábor was telling you about at IROS 2019).

The implementation boils down to two key ideas:

* Instead of storing the raw polynomial coefficients, we store the coordinates of the control points of an equivalent Bézier curve. This is because the control points correspond to spatial coordinates so they are in a predictable range that can safely be quantized to, say, millimeters, without losing accuracy, allowing us to use only two bytes per coordinate.

* Each segment in the trajectory is preceded by a header that describes how the coordinates will be encoded in the segment. For each coordinate, one may choose from constant, linear, cubic Bézier and 7th degree Bézier representations. This helps a lot for trajectories where the yaw does not change, or where most of the movement happens in one of the XY, YZ or ZX planes.

More details are to be found in `docs/trajectory_formats.md`.


ntamas:
Note: CI failure is probably unrelated.

src/modules/interface/pptraj_compressed.h
```diff
@@ -0,0 +1,71 @@
+/*
+
+Compressed piecewise polynomial trajectories for Crazyflie.
+
+Copyright (c) 2019 CollMot Robotics. All rights reserved.
```
ntamas: This line can be removed if the PR is accepted.

src/modules/src/pptraj_compressed.c
```diff
@@ -0,0 +1,284 @@
+/*
+
+Compressed piecewise polynomial trajectories for Crazyflie.
+
+Copyright (c) 2019 CollMot Robotics. All rights reserved.
```
ntamas: This line can be removed if the PR is accepted.

krichardsson:
Great functionality and very good pull request! Thanks!

Would it be possible for you to add a license section at the top of the new files? Something similar to what we have in other files. You can choose what ever license you want, as long as it is compatible with GPL V3. There are examples in the code base for various licenses that contributors have used if you want some inspiration.

br
- Kristoffer

ntamas:
Yes, absolutely. GPLv3 is okay for me -- I just wanted to keep our copyright notice in the headers for the duration of the review in case it gets rejected. Once it passes review, I'll replace the copyright notice with a GPLv3 header.

krichardsson:
I'm happy with the code and I'm ready to merge when you have added the license text. Please leve the copyright notice though, it should be there as well.

ntamas:
Done - I'm not that good at ASCII arts though so I just generated one with `figlet` :)

krichardsson:
Thanks!

PR #510, opened 2019-11-26
Feature/freertos thirdparty

pablogs9:
This PR sets the FreeRTOS in src/lib/FreeRTOS as git module.

It uses the branch V10.2.1-convergence-FreeRTOS-Source from  [FreeRTOS-Kernel](https://github.com/FreeRTOS/FreeRTOS-Kernel/tree/V10.2.1-convergence-FreeRTOS-Source) repo. This approach is used, for example, by Amazon AWS FreeRTOS ([check this repo](https://github.com/aws/amazon-freertos)).

Regarding the upgrade from V8.2.3 to V10.2.1, documentation says that is a drop-in compatible replacement. Check [upgrading from V8 to V9](https://www.freertos.org/FreeRTOS-V9.html) and [upgrading from V9 to V10](https://www.freertos.org/FreeRTOS-V10.html).

First build and flash works.


krichardsson:
Thanks!

ataffanel:
Shouldn't we move the FreeRTOS subrepos to the ```vendor/``` subdirectory? So far we have put copy-pasted libs in ```lib/``` and definitly unmodified subrepos in ```vendor/```.

krichardsson:
Agree, vendor is a better location.

PR #513, opened 2019-11-29
Heap management options

pablogs9:
This PR enables OOT apps to overwrite MEMMANG_OBJ with a custom implementantion of the heap manager.

It also enables the build system to set FREERTOS_HEAP_SIZE from CFLAGS.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -96,7 +96,7 @@ VPATH += $(LIB)/vl53l1/core/src
 VPATH += $(PORT)
 PORT_OBJ = port.o
 VPATH +=  $(FREERTOS)/portable/MemMang
-MEMMANG_OBJ = heap_4.o
+MEMMANG_OBJ ?= heap_4.o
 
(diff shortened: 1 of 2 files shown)


ataffanel:
This looks good, I am merging it, thanks!

One side note about that: we are planning on removing all dynamic memory allocation from the Crazyflie firmware itself since it does not fit very well with hard real-time. In a sense it makes this PR more relevant though since only the app layer might then be interested in dynamic memory.

pablogs9:
I've read something about this migration to static memory, the v10 version of the kernel (merged in #510) allows this static allocation of tasks, queues, etc. Thaks for the merge :).

pablogs9:
@ataffanel is this PR going to be in next-release milestone? Thanks

ataffanel:
Yes sure, we release the tip of the master branch, so everything that is in master will be released.

PR #521, opened 2019-12-05
stabilizer.thrust is a float so it should be logged as a float

ntamas:
In `stabilizer.c`, the `thrust` log variable in the `stabilizer` group (which actually comes from `control.thrust` where `control` is of type `control_t`) is logged as an `UINT16_T`, even though it is declared as a float in `control_t`. This PR fixes the type of the log variable.


krichardsson:
Thanks!

PR #545, opened 2020-01-30
Set correct RAM sizes in linker scripts

### LINKED ISSUE #544: Wrong RAM size in linker scripts
Looks like the RAM size is incorrect in the linker scripts for the STM32F405. The one on the CF2.0 has 192kB of RAM, but is declared as 128kB. Obviously the limit is not being hit in the main firmware, but in a project I'm doing right now I've run into it.

I'm also submitting a PR to fix it

https://github.com/bitcraze/crazyflie-firmware/blob/master/tools/make/F405/linker/DEF_CLOAD.ld#L13
https://github.com/bitcraze/crazyflie-firmware/blob/master/tools/make/F405/linker/DEF.ld#L13

abrandemuehl:
Not sure how valid the copyright headers are, so I've left them in.

This is to fix #544


abrandemuehl:
As mentioned in #544, this is due some extra RAM not being enabled in the code, so this wouldn't work anyways.

PR #548, opened 2020-02-04
Improve ASSERT error handling and reporting

whoenig:
* Motors are now disabled after an assert. This is in particular
  important if the watchdog is disabled.
* ASSERT information is now stored for FreeRTOS related issues
  (out of memory, stack overflow). Note that the stack overflow
  detection has to be enabled manually for release builds.
* HardFault asserts now include more detailed information, e.g.,
  the program counter. One can use GDB to find the source code
  line that triggered a hard fault.

### DIFF
--- a/src/drivers/interface/motors.h
+++ b/src/drivers/interface/motors.h
@@ -229,6 +229,8 @@ void motorsDeInit(const MotorPerifDef** motorMapSelect);
  */
 bool motorsTest(void);
 
+void motorsDisable(void);
+
 /**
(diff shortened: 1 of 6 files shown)


krichardsson:
Nice, thanks!

krichardsson:
@whoenig I just realized that there is already a function for turning off the motors. In power_distribution_stock.c, there is powerStop() function that does the same thing as the new powerStop() in motors.c.

whoenig:
Sorry about that - see #550 for a fix.

PR #553, opened 2020-02-14
drop Python 2.x requirement for building as the scripts work fine with Python 3.x

ntamas:
This PR modifies the Makefile so that it does not require Python 2.x for the build process any more. Python 2.x has reached its end of life in Jan 2020, and the build process works just fine with Python 3.x, so I don't think there's a need to force the user to install Python 2.x.

(I have recently uninstalled Python 2.x from my Mac, that's how I stumbled upon this).


krichardsson:
Merged
Thanks!

PR #554, opened 2020-02-18
fix log types of motorPower.m[1-4]

ntamas:
We have stumbled upon these mistyped log variables while trying to port Mike Hamer's controller to the head revision of the firmware. The old types probably caused no harm so far because the logged values did not go out the `uint16_t` range, but it's best be on the safe side.


krichardsson:
Merged, thanks!

ntamas:
@krichardsson I think this may have been closed _without_ merging after all; I cannot see this in the latest revision and Github says that the PR is closed without merging.

krichardsson:
Sorry about that, my bad! Now it is merged :-)

PR #558, opened 2020-02-24
uart link

s9meboeh:
Enable CRTP communication via Uart.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 # CrazyFlie's Makefile
 # Copyright (c) 2011,2012 Bitcraze AB
-# This Makefile compiles all the objet file to ./bin/ and the resulting firmware
+# This Makefile compiles all the object file to ./bin/ and the resulting firmware
 # image in ./cfX.elf and ./cfX.bin
 
 CRAZYFLIE_BASE ?= ./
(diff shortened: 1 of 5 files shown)


knmcguire:
Hi! 

From the first sight this looks good but we would need to check if this works properly, which we will try to verify this week.

Is there a reason why the CRTP through usb or crazyradio does not work out for you?

s9meboeh:
Hello Kimberly,

I am from the team of the Saarland University that extended the Drone
for Wifi communication with the Raspberry Pi, thus we needed Uart to the
Raspberry Pi.
I am still struggeling with the toolbelt for the Crazyflie library,
otherwise I would already have created a pull request for this one too
and send you an Mail with further information.
Probably it makes sense to wait with the review till you also have this
counterpart, I hope to get it done today or tomorrow.

Best regards,
Marlene

knmcguire:
Code looks good and is all behind a build flag so it is not really affecting anybody else using it. 

A small word of concern is that, with the flag on, communication via the crazyradio is not possible anymore. For the USB and Radiolink, there is an automatic switch, which would be nice to implement the uart in there as well. But we will consider this as a possible future improvement, and it is all safely behind a build flag anyway :)

Issue #536, opened 2020-01-18
Unable to execute high-level commands immediately after low-level commands have been sent


jpreiss:
In my own work and in others' (https://github.com/bitcraze/crazyflie-firmware/issues/452, https://github.com/USC-ACTLab/crazyswarm/issues/143), we have often hit the issue that it is impossible to instantaneously switch from streaming low-level setpoints to high-level commands. The most common desired sequence is:

1. Take off in high-level mode
2. Hand over to some control algorithm running on PC that generates streaming setpoints
3. Land (possibly with `goTo` before) in high-level mode

In the following:

https://github.com/bitcraze/crazyflie-firmware/blob/438f33eaf0972c642745ef5b5d7c895465b0f9ca/src/modules/src/commander.c#L90-L93

we never evaluate the high-level setpoint until the most recent low-level setpoint is old (the high-level commander never modifies `setpoint->timestamp`).

The current system seems built for the use case where the high-level commands coming from the PC are incorrect/unsafe, and we want to override them with manual control inputs even when the PC is sending fresh incorrect high-level commands continuously. That makes sense, but maybe we could add a param or radio packet to override it?

hanzheteng:
From a user's perspective, I would say this feature is in high demand.

krichardsson:
@jpreiss I think I would prefer a packet that resets the timeout. Something like "I'm done with low level setpoints and want to use the hight level commander"

jpreiss:
@krichardsson that sounds good to me. Looking at the snippet of `commander.c` above, the behavior upon `COMMANDER_WDT_TIMEOUT_SHUTDOWN` depends on `bool enableHighLevel` which, as far as I can tell, is only controllable via the param system. If that is correct, then it seems most natural to call the packet something like `notifySetpointsEnding`, i.e. to avoid mentioning the high level commander explicitly.

A simple way to implement the desired behavior would be to overwrite `setpoint->timestamp` with zero when this packet is received. This way, if we do in fact need manual control to take over due to erroneous high-level commands later, we don't need to do anything more complicated than sending a fresh low-level setpoint.

Unfortunately, I can't see how to do that in the current architecture. `crtp_commander.c` can only interact with the core commander module by calling `commanderSetSetpoint()`, but that function writes the current `xTaskGetTickCount()` into `setpoint->timestamp`.

Any design thoughts?

jpreiss:
[...]

PR #560, opened 2020-02-27
#536 ability to enable high-level commands after low-level commands

jpreiss:
Extensive discussion of the code design is in #536. This code is tested on real hardware. Confirmed that we can:

- Take off in high-level mode
- Fly using streaming setpoints (I tested `velocity_world` and `full_state`)
- Re-enable high-level mode
- Immediately land or do a goTo

I was not able to test manual override because I was not sure how. However, based on the design I am confident it will work.

### DIFF
--- a/src/modules/src/crtp_commander.c
+++ b/src/modules/src/crtp_commander.c
@@ -27,6 +27,7 @@
 
 #include "crtp_commander.h"
 
+#include "cfassert.h"
 #include "commander.h"
 #include "crtp.h"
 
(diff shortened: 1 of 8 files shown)


src/modules/src/crtp_commander.c
```diff
@@ -55,8 +103,19 @@ static void commanderCrtpCB(CRTPPacket* pk)
...
+      commanderSetSetpoint(&setpoint, COMMANDER_PRIORITY_CRTP);
+      break;
+    case 1: {
+        uint8_t metaCmd = pk->data[0];
+        if (metaCmd < nMetaCommands && (metaCommandDecoders[metaCmd] != NULL)) {
+          metaCommandDecoders[metaCmd](pk->data + 1, pk->size - 1);
+        }
+      }
+      break;
+    }
```
krichardsson: Named constants makes it easier to understand what the cases are handling. Maybe something like 
```
switch (pk->channel) {
  case SET_SETPOINT_CHANNEL:
    // ...
    break;
  case META_COMMAND_CHANNEL:
    // ...
    break;
  default:
    // Do nothing
    break;
}
```

and a default for completeness.

jpreiss:
@krichardsson Thank you for the review! I added the channel `enum` and got rid of the `DEBUG_PRINT`s. 

Following the style of `crtp_commander_generic.c`, I also added comments in `crtp_commander.c` with a numbered "to-do" list for adding a new command, and changed the comments from `//` to `/* */` style.

krichardsson:
Merged. Thanks!

PR #563, opened 2020-03-08
Kyle app

Kpellham:


krichardsson:
Hi! 
The intention of the app layer is to enable users to write their own apps with an [out of tree build](https://www.bitcraze.io/docs/crazyflie-firmware/master/build_instructions/#out-of-tree-build) to separate the app from the firmware. The idea is that you store the source code of your app in your own repo and pull in the Crazyflie firmware as needed. For this reason I will close this pull request.

Please reopen this pull request if I misunderstood the situation.

PR #571, opened 2020-04-03
Doc for creating a new system task

jpreiss:
I needed to explain this procedure to collaborators, so I figured it would be useful to write it up as a public doc. I started from the framework of `howto.md`.

I hope this is useful. I am happy to rewrite parts or add/correct information as needed.

@Zhenghao-Dai can you try to follow these steps and make sure I didn't leave anything out?

### DIFF
--- a/docs/development/systemtask.md
+++ b/docs/development/systemtask.md
@@ -1,0 +1,263 @@
+---
+title: Adding a new system task
+page_id: systemtask
+---
+
+This howto describes how to create a new system task.
+A FreeRTOS task is similar to a thread on a standard operating system.
+It has its own function call stack, and can be preempted by the RTOS scheduler.
+Typical hazards and techniques of shared-memory multithreaded programming apply.
+
+A new task is most often needed when adding a fundamentally new subsystem
+to the Crazyflie firmware.
+
+If the new subsystem needs to do
+significant computation in response to inputs such as CRTP radio messages, it
+should receive those inputs on a queue and perform the computation within
+its own task instead of blocking the radio task.
+In this example, we will set up the skeleton for such a new subsystem.
+
+Development environment
+-----------------------
+
+You should have the
+[crazyflie-firmware](https://github.com/bitcraze/crazyflie-firmware) and
+[crazyflie-clients-python](https://github.com/bitcraze/crazyflie-clients-python)
+cloned in the same folder. If you are using the [Bitcraze
+VM](https://github.com/bitcraze/bitcraze-vm) this is already the case.
+
+The Crazyflie firmware should be on Master branch.
+
+For the rest of the howto you will work in the crazyflie-firmware
+project.
+
+
+Implementing the task
+---------------------
+
+High level tasks (those that do not directly talk to hardware) usually go in
+the `modules` directory. We will walk through the sections needed in a
+new file `modules/src/example.c`.
+
+
+First, include the necessary system header files:
+
+``` {.c}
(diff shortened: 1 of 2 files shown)


krichardsson:
Great work!
We should have more of this type of documentation. Thanks!

I have added some comments in the doc for minor updates that I found.

docs/development/systemtask.md
```diff
@@ -0,0 +1,263 @@
...
+that need to last as long as the firmware is running.
+Later in this howto, we will need to edit another file to call the `Init()`.
+
+``` {.c}
+static bool isInit = false;
+
+void exampleTaskInit() {
+  inputQueue = STATIC_MEM_QUEUE_CREATE(inputQueue);
+  // TODO
+  STATIC_MEM_TASK_CREATE(exampleTask, exampleTask, KALMAN_TASK_NAME, NULL, KALMAN_TASK_PRI);
```
krichardsson, line 94: Copy/paste from the kalman task.
EXAMPLE_TASK_NAME and EXAMPLE_TASK_PRI would be better

docs/development/systemtask.md
```diff
@@ -0,0 +1,263 @@
...
+  return isInit;
+}
+```
+
+
+Now we come to the main task function.
+It is usually an infinite loop that waits on inputs from something.
+In our case, we wait on inputs from the queue.
+
+Our `xQueueReceive` call waits for up to 1000 ticks - one second.
```
krichardsson, line 114: It might be a good idea to mention that ```portMAX_DELAY``` will wait forever and only release the lock when the queue is not empty. I think this is a common use case.

docs/development/systemtask.md
```diff
@@ -0,0 +1,263 @@
...
+Target info: stm32 (0xFF)
+Flash pages: 1024 | Page size: 1024 | Buffer pages: 10 | Start page: 16
+1008 KBytes of flash available for firmware image.
+
+Flashing 1 of 1 to stm32 (fw): 161867 bytes (159 pages) ..........10..........10..........10..........10..........10..........10..........10..........10..........10..........10..........10..........10..........10..........10..........10.........9
+Reset in firmware mode ...
+$
+```
+
+Since our task doesn't do anything, there is nothing to observe when running
```
krichardsson, line 262: If you want to you could add a DEBUG_PRINT() to the new task to show that it is running

docs/development/systemtask.md
jpreiss: New constants created in 'config.h' in d3c0a97.

docs/development/systemtask.md
jpreiss: Switched from 1000ms to `portMAX_DELAY` timeout value and added explanation.

docs/development/systemtask.md
jpreiss: Added in d3c0a97.

jpreiss:
I added the link in `menu.yml`. When I run `tb docs 8080` the link and the new page show up, but it seems like the whole served site is missing any CSS styling. Guessing that is an unrelated issue.

knmcguire:
Hi James.

That is fine, it is not supposed to show the bitcraze website styling. But at least now you can check if the markdown files and menu are correct and to your liking before pushing it.  And once this is accepted and pushed, we need to deploy the web again and then you will be able to see it on the actual website.

krichardsson:
Merged. Thanks!

Issue #567, opened 2020-03-29
Simple onboard collision avoidance for multiple Crazyflies


jpreiss:
In multi-quadrotor applications like mapping and search/rescue, where the task requires the robots to spread out, the risk of collisions is inherently low. We want to ensure collision avoidance, but heavy centralized planning methods that handle maze-like environments and dense formations (like @whoenig's work) are overkill. A simpler solution is some kind of reactive onboard collision avoidance, where each quadrotor attempts to move towards its goal position in a straight line, but reacts appropriately if another quadrotor is near.

We considered implementing something like this as library functionality in [Crazyswarm](https://github.com/USC-ACTLab/crazyswarm), but implementing it in the firmware has significant benefits:

- Most importantly: makes the functionality available to a larger set of Crazyflie users.
- Minimizes communication bandwidth - only need to send one packet with the goal position and any algorithm control parameters.
- Minimizes computational load on the base station PC.
- Keeps the Crazyswarm control script simple.

We would like to contribute our implementation back to the main firmware project.  I created this issue to get some design feedback before sending any PR.

# Algorithms

One suitable algorithm is [buffered Voronoi cells (BVC)](https://www.youtube.com/watch?v=rrDmcE94OQw).  Another possibility is [optimal reciprocal collision avoidance (ORCA)](https://www.youtube.com/watch?v=zmTgj4SKftc).  ORCA gets better results when spatial contention is high, but it requires knowing the velocity of neighbors, and imposes stricter constraints on the kind of control laws that can be followed.  Buffered Voronoi Cells only constrains that each robot should stay within its cell during the planning horizon.

I already have a prototype of BVC implemented in C and tested: https://github.com/bitcraze/crazyflie-firmware/compare/master...jpreiss:collision_avoidance. Most the core math is useful for generic applications so it has been implemented in https://github.com/jpreiss/cmath3d. I did not use the model-predictive optimal control from the BVC paper.  I use a simple projection of the desired direction vector into the Voronoi cell polytope.  It runs in less than 1ms on my PC, so I believe it will be fast enough to run at an appropriate rate on the Crazyflie's STM32.

# Software Design

## High-level commands

Currently there are only two fundamental control modes in the Crazyflie firmware:

- Follow a piecewise polynomial trajectory from the high-level commander.
- Follow streaming low-level setpoints from the PC.

[...]

PR #574, opened 2020-04-24
add peer_localization to store the position of other crazyflies

Zhenghao-Dai:
#567 In our discussion about implementing the collision avoidance, @krichardsson suggests that we put the position of other crazyflies in its own module. @jpreiss

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -167,7 +167,7 @@ PROJ_OBJ += range.o app_handler.o static_mem.o
 
 # Stabilizer modules
 PROJ_OBJ += commander.o crtp_commander.o crtp_commander_rpyt.o
-PROJ_OBJ += crtp_commander_generic.o crtp_localization_service.o
+PROJ_OBJ += crtp_commander_generic.o crtp_localization_service.o peer_localization.o
 PROJ_OBJ += attitude_pid_controller.o sensfusion6.o stabilizer.o
 PROJ_OBJ += position_estimator_altitude.o position_controller_pid.o
 PROJ_OBJ += estimator.o estimator_complementary.o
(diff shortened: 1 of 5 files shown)


krichardsson:
Thanks for the pull request!

It looks good, but I have a couple minor comments:

1. The type "struct otherPosition:"
    1. We tend to typedef structs, see for instance https://github.com/bitcraze/crazyflie-firmware/blob/master/src/modules/interface/stabilizer_types.h#L57
    1.  We also try to name types based on the module to get some sort of "name space"  (previous is not a good example :-)), for instance this type should maybe be called peerLocalizationOtherPosition_t. There are no clear guidelines on naming, sorry!
    1. You could use the type point_t (in the example above), it already contains a timestamp if you want to
    1. It is probably enough to use a uint32_t for the timestamp

2. The size of the NUM_MAX_NEIGHBORS is fairly large considering that this functionality probably will be used by a limited number of users, in combination with that we are starting to be low on RAM. I'm not sure what a "normal" use case is for this module? What would make sense?

jpreiss:
Re. 1: @Zhenghao-Dai can you update the pull request? I think we can use the `point_t` as suggested. No need to add a redundant type.

Re. 2: I'm not sure about `NUM_MAX_NEIGHBORS`. I think 10 would be enough for most uses.

Users can always recompile and flash the firmware to support larger teams, but that gets more and more tedious as the number increases. Something like the dynamic memory allocation currently used to support long polynomial trajectories would be ideal.

Would it make sense to have a `peerLocalizationNumNeighbors` param that triggers a dynamic memory allocation when it changes? The default value could be zero. I think it would be reasonable to assume that this param will only change once after firmware bootup, so we can avoid worrying about memory fragmentation.

@whoenig any thoughts?

Zhenghao-Dai:
Updated

jpreiss:
@Zhenghao-Dai I think this CI build failure is related to copying the comments from Zoom chat, like we saw before. Try retyping the comments yourself.

Zhenghao-Dai:
Fixed and passed @jpreiss

krichardsson:
Merged. Thanks!

Issue #576, opened 2020-05-03
Inter ranging between multiple crazyflies by using the Loco decks


shushuai3:
This issue will discuss a new feature of inter two-way-ranging between multiple crazyflies, which will lead to relative localization and control between multiple crazyflies without relying on the loco positioning nodes on the ground.

knmcguire:
Thanks @shushuai3 ! This seems like it could be combined with the PR #574 as indicated in #567. Could you tell a bit how you would like to implement it? Then we can comment as well on what is handy in the firmware

- Would this function be a part of the LPS-deck driver?
- Is this considered to be always on or should it be activated by a parameter or flag?
- How many crazyflies should communicate with eachother

keffklau:
Hello Kimberly, hello shushuai3,

such a function would be really great. We also plan to make use of inter-agent range measurements for improving navigation quality. However, we would like to use CF as mobile anchors in addition to the fixed mounted LPS nodes. So we would prefer a solution combining both distance measurements TDoA3 with anchors, ToA/TWR in between Crazyflies.

In order to combine those measurements, state&covariance information concerning the navigation filters would have to be exchanged. Is there a possibility to encode those within the UWB messages since this would reduce time-drifts between measurement and filter data? 

- We would like to pass the inter-agent range measurements (and agent data) to the navigation filter but the measurement itself could be done within the deck driver. 

- We would like to be able to switch it on and off (in order to show the effect on navigation quality)

- It's important for us, that this approach does not cause issues with the flow deck, since we want to combine information

- From our point of view, CFs should randomly initiate range measurements with each other so the number of CFs can be arbitrary

@shushuai3 I am not sure, whether this is in line with your objectives, hopefully there is no contradiction

krichardsson:
Hi all!

krichardsson:
Hi all!

The idea of (sometimes) ranging between Crazyflies and have them act as anchors from time to time, is something we discussed when we wrote TDoA3 (internally we usually call it TDoA4). 
In my opinion, the best way to implement such a functionality would be to reuse the TDoA3 protocol, which would make it compatible with anchors running TDoA3. In the TDoA modes, anchors are constantly ranging with each other and the TDoA part is a "side effect", so one way to look at this is that the only thing that needs to be added is that Crazyflies sometime also may transmit packets.

[...]

PR #580, opened 2020-05-08
inter-ranging between multiple crazyflies using loco decks

shushuai3:
To achieve two-way-ranging between multiple crazyflies:
1. set the address of each crazyflie ending by E5, E6, E7, ...
2. set the NUM_CFs in lpsTwrInCFs.h to the actual number of crazyflies
3. make LPS_TWRINCFS_ENABLE=1

The ranging distances can be seen from ranging.distance0, ranging.distance1, ... in the cfclient.

This pull request is related to issue #576

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,7 @@ PLATFORM          ?= cf2
 LPS_TDMA_ENABLE   ?= 0
 LPS_TDOA_ENABLE   ?= 0
 LPS_TDOA3_ENABLE  ?= 0
+LPS_TWRINCFS_ENABLE ?= 0	
 
 
 # Platform configuration handling
@@ -225,6 +226,11 @@ endif
 
 ifeq ($(LPS_TDMA_ENABLE), 1)
 CFLAGS += -DLPS_TDMA_ENABLE
+endif
+
+ifeq ($(LPS_TWRINCFS_ENABLE), 1)
+PROJ_OBJ := $(filter-out lpsTwrTag.o,$(PROJ_OBJ))
+PROJ_OBJ += lpsTwrInCFs.o
 endif
 
 ifdef SENSORS
--- a/src/deck/drivers/interface/lpsTwrInCFs.h
+++ b/src/deck/drivers/interface/lpsTwrInCFs.h
@@ -1,0 +1,28 @@
+#ifndef __LPS_TWR_TAG_H__
+#define __LPS_TWR_TAG_H__
+
+#include "locodeck.h"
+#include "libdw1000.h"
+
+#include "mac.h"
+
+#define LPS_TWR_POLL    0x01  
+#define LPS_TWR_ANSWER  0x02
+#define LPS_TWR_FINAL   0x03
+#define LPS_TWR_REPORT  0x04
+
+#define LPS_TWR_TYPE 0
+#define LPS_TWR_SEQ 1
+
+#define NUM_CFs 3
+
+extern uwbAlgorithm_t uwbTwrTagAlgorithm;
+
+typedef struct {
+  uint8_t pollRx[5];
+  uint8_t answerTx[5];
+  uint8_t finalRx[5];
+  uint16_t distance; // mm
+} __attribute__((packed)) lpsTwrInterCFsReportPayload_t;
+
+#endif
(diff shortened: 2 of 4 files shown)


shushuai3:
This pull request is related to issue #576

knmcguire:
Thanks @shushuai3! Now it is clear on what you meant. We will look at it and discuss it internally with @krichardsson on a strategy how to implement this in a generic way for the firmware. We will update you once we know more.

krichardsson:
Thanks for your pull request!

We think inter drone ranging would be an interesting addition to the Crazyflie software stack.
Having that said, we also have to consider a number of properties of new code such as:
* Is the solution generic enough to be used by more users?
* Is the API flexible and easy to use?
* Can the code be maintained and tested? When we add code to the system, we also commit to maintain it for the future.
* Will the code work together with other parts of the system such as the python lib, clients and other hardware components.
* Memory and CPU consumption
* Stability
* Coding style

In this case we have found that this pull request needs modifications to be accepted, and our main concerns are:

1. The implementation replaces the TWR driver with the inter drone ranging driver using a compile flag.
This breaks the API for the python lib/client and could cause confusion (and possibly crashes) for the user. The compile flag also makes this harder to test and maintain, as we would have to recompile the firmware for regression testing during releases.
We would prefer a solution where the functionality is added as a new mode to the existing LPS modes, or as an extension
to one of them.

1. The current solution is intended to be used with 5 Crazyflies running the same firmware, but is not compatible with other LPS modes (as far as we can see, correct us if we are wrong). It would be nice to have interoperability with other parts of the eco system, such as anchors. The current solution uses the same protocol as TWR (we assume) with fixed time slots for TX/RX. A more flexible and scalable approach would be to utilize the TDoA3 protocol (stil doing
two way ranging) that would provide interoperability with anchors and Crazyflies using "normal" TDoA3, as well as scalability in numbers.

1. The output from the ranging is exposed as log variables. This is one way to do it, but it would be nice to be able to maybe route it to the estimator as well even though that could be a future addition.

We are more than happy to answer any questions you may have or discuss possible solutions to help you reach an implementation we can accept. Another way forward could possibly also be that we make a rough "sketch" implementation to get you stared in the right direction, but we're not sure when that would be. Please let us know your thoughts.

shushuai3:
@krichardsson Thanks for your review of my pull request! First I would like to clarify the properties of the inter-drone ranging code.
- Generality. The current code only supports a fully connected topology. But a function for arbitrary topology is achievable.
- API. Currently, the ranging variables can be called by other files. A similar API like estimatorEnqueueDistance would be more flexible and easy to use.
- The code can be maintained just like other LPS methods.
- Up to now, the code has no conflicts with the python lib, clients, and other hardware.
- Memory and CPU consumption is similar to that of two-way ranging in theory.
- The inter-drone ranging can be run infinitely. It may stop if two drones are too far away.
- Code style approximates to that of two-way ranging.

As for your main concerns:
1. The inter-drone ranging will be reformed as a new LPS algorithm though there are some invisible errors to be fixed.
2. TDoA3 sounds very useful to the interoperability. As it could be a large modification and may take much time, I am wondering if it could be put in a future commit? If it is, the code can be revised soon. Otherwise, more time is needed to test the TDoA3.
3. A similar way of 'estimatorEnqueueDistance' can be added.

krichardsson:
Hi and sorry for the late reply!

I have done some more thinking and this feature (#576) can be fairly complex depending on which way we want to go, I'm not sure what the best solution is. 
Your initial proposal, has advantages in terms of few dependencies to other modules, but is also a bit limited in functionality and other users will need more features. 
Other options could be to extend TDoA3 in various ways which would give more functionality and flexibility, but also tend to have effects on other parts of the system (client, library, anchors and so on) that will require more work in other repositories.

One problem with changing the implementation later (like you suggest to add TDoA3 later) is that it breaks compatibility. Someone may have started to add functionality on top of the code that breaks when the underlying functionality changes. We try to avoid it if possible. 

Maybe the best way forward is to continue the discussion in #576 to try to figure out what features that are desired?

shushuai3:
Thanks for your suggestions. I agree with you. Let's discuss the potential features in #576 first.

knmcguire:
Hi! I will close this PR due to inactivity and since we are not able to merge this into the current firmware. No worries, it will not be lost, we just want to clean up a bit :) I'll keep the issue #576 so we can talk development there, but for a interranging feature, there needsto be a new pull request

PR #581, opened 2020-05-11
break stabilizer_types dependency on ARM

jpreiss:
Compiling, testing, and debugging bits of `crazyflie-firmware` code on x86 is very useful [1]. To compile the code for #567, we need to include `stabilizer_types.h`, because the collision avoidance function operates on a `setpoint_t`. However, `stabilizer_types.h` depends on `lighthouse_geometry.h`, which includes an ARM math library.

`lighthouse_geometry.h` declares a function that accepts an `arm_math.h` type by pointer. Unfortunately, we cannot forward declare the pointed-to type because it's an anonymous struct of form
  
    typedef struct {
       ....
    } name;

These typedefs cannot be forward-declared.

It turns out that the only part of `lighthouse_geometry.h` that `stabilizer_types.h` needs is some trivial typedefs of fixed-size arrays. I felt the lowest impact fix would be to repeat those typedefs in `stabilizer_types.h`.

If the repetition is not acceptable, other options would be to put the typedefs in some other dependency-free file that `lighthouse_geometry.h` and `stabilizer_types.h` can both include, or to rewrite `lighthouse_geometry.h` to use `math3d.h` types.

The `controller_indi` implementations transitively depended on `arm_math.h`'s 32-bit pi constant. I replaced it with the one from `math3d.h`.

[1] For past modules like `pptraj.h`, the modules only depend on `math3d.h` and the standard library, so this problem doesn't exist.

### DIFF
--- a/src/modules/interface/stabilizer_types.h
+++ b/src/modules/interface/stabilizer_types.h
@@ -29,7 +29,6 @@
 #include <stdint.h>
 #include <stdbool.h>
 #include "imu_types.h"
-#include "lighthouse_geometry.h"
 
 /* Data structure used by the stabilizer subsystem.
  * All have a timestamp to be set when the data is calculated.
@@ -241,6 +240,13 @@ typedef struct {
 } yawErrorMeasurement_t;
 
 /** Sweep angle measurement */
+
+// Forward-declare math types from lighthouse_geometry.h to avoid transitive
+// dependency on arm_math.h. Necessary to allow compiling this file for PC,
+// which is useful for testing.
+typedef float vec3d[3];
+typedef float mat3d[3][3];
+
 typedef struct {
   uint32_t timestamp;
   vec3d* baseStationPos;
--- a/src/modules/src/controller_indi.c
+++ b/src/modules/src/controller_indi.c
@@ -24,6 +24,7 @@
  */
 
 #include "controller_indi.h"
+#include "math3d.h"
 
 static float thrust_threshold = 300.0f;
 static float bound_control_input = 32000.0f;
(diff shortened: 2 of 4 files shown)


krichardsson:
Good point! 
stabilizer_types.h is used in may files and should probably have as few (no?) dependencies as possible.
I think your changes are fine, but you can take it even further. The definition in stabilizer_types.h should not be a forward declaration, it should be The declaration. you can simply remove the old one from lighthouse_geometry.h and include stabilizer_types.h instead.

It would be interesting to get your x86 build into the build server some how to catch this type of dependencies as early as possible.

As a side note, we could probably clean up the types a bit. For instance, I think there are lots of similar vector types that maybe could be harmonized.

krichardsson:
Slack or google meetings?

jpreiss:
Was this message meant for me? It looks like there is a missing context...

jpreiss:
If we are moving the declaration to `stabilizer_types.h`, maybe we should change the name to reduce confusion with `vector_t` and `struct vec`. How about `array3` and `array33`?

I agree unifying vector types would be great for future work. There's also `Axis3f` in `imu_types.h`.

The x86 build is part of the Crazyswarm project. Right now we only build `pptraj.h`, `planner.h`, and `math3d.h`. We use SWIG to generate Python bindings for all of the structs and functions in those files. Since we want to simulate multiple Crazyflies in one process, we are limited to modules that don't use any global variables. It's pretty self-contained in https://github.com/USC-ACTLab/crazyswarm/tree/master/ros_ws/src/crazyswarm/scripts/pycrazyswarm/cfsim.

jpreiss:
Hmm, somehow `kalman_core.c` transitively includes `lighthouse_geometry.h` and uses `vec3d`. Maybe renaming is not worth it.

krichardsson:
Maybe wait with the renaming

krichardsson:
Merged. 
It looks like you "force-pushed" to the pull request. Not sure if that contained some changes, at least they were not visible from my side :-)

jpreiss:
Oops, I guess I should not be trying to rewrite git history at 1:40 AM! Doesn't matter - the force pushed version was supposed to make the exact change you made.

PR #583, opened 2020-05-26
traj_eval_zero() in pptraj

jpreiss:
This is a trivial change. Not used anywhere in the firmware yet, but it is needed (via SWIG wrapper) for the upcoming Crazyswarm-based collision avoidance test script  (#567).


krichardsson:
Merged!

PR #584, opened 2020-05-27
updated links in documentation

chronoB:
I was looking through the documentation and noticed that the links for the images weren't working on github so I updated them. 
While doing it I also fixed some tables and some external links.
I also added a paragraph with an internal link because I remembered that I was stuck at that point when reading the documentation on my own and that link would have helped me. 

I hope that is useful and wanted and correct.

### DIFF
--- a/docs/building-and-flashing/flashing.md
+++ b/docs/building-and-flashing/flashing.md
@@ -27,7 +27,7 @@ Warning: if multiple Crazyflies within range are in bootloader mode the result i
 
 ### Automatically enter bootloader mode
 
-* Add the address of the crazyflie to the [`config.mk`](/building-and-flashing/configure_build/) file, for instance `CLOAD_CMDS = -w radio://0/80/2M`
+* Add the address of the crazyflie to the [`config.mk`](/docs/building-and-flashing/configure_build.md) file, for instance `CLOAD_CMDS = -w radio://0/80/2M`
 * Make sure the Crazyflie is on
 * In your terminal, run `make cload`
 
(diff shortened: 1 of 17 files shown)


krichardsson:
Hi and thanks for the contribution!

There is a slight problem with your changes in relation to our systems, but also a possibility of improvement. I'll need to give you some background to explain what I mean.

Our web site (https://github.com/bitcraze/bitcraze-website) is generated from markdown using Jekyll with some custom plugins and processing to bake it all into a working web.
Last year we started to move documentation from our wiki to the various repos we have (to get the documentation closer to the code), and settled on a structure where we put the new documentation in the docs folder using markdown. The markdown files are pulled from the repos into the web build and massaged to work with the rest of the content (you'll find it here https://www.bitcraze.io/documentation/repository/, for instance https://www.bitcraze.io/documentation/repository/crazyflie-firmware/master/).

You can also read about the docs here https://www.bitcraze.io/development/contribute/repo-doc-guidelines/

The structure of internal links and paths are designed to work on our web and unfortunately not on github, why your changes would break the web (this is the problem mentioned earlier). We (I) did not really think about making it github compatible when we set it up (makes me feel a bit stupid now), but this would be a really nice feature (this is the possible improvement). It will require some thinking and tweaking on our side to make it work though, and i'm not sure when it will happen. I'll leave the this pull request open until that is fixed.

You mentioned that you also fixed something that explained a problem you had when being stuck. It might be a good idea to issue a second pull request with this change so I can merge that sooner.

Thanks!
- Kristoffer

chronoB:
Hey Kristoffer,

thanks for the comprehensive explanation!
I already thought that there would be some problem just because the links were so consistent. So thanks for the background on that! It's always nice to learn about something new especially when it's backed by such an extensive explanation and links.
I will see that i will add a second pr. 
Have a great day!

krichardsson:
I have been looking at adding support for this type of linking, but it will probably take a while before it is done. I'll close this PR for now.

PR #585, opened 2020-05-28
added link to configure_build.md for clarification

chronoB:
I added a reference to the configure_build.md because i remembered that i had some trouble finding the correct information when I was working with this part of the docs. 
So I thought it could help to have it there.


krichardsson:
Thanks!

Issue #592, opened 2020-06-12
Lighthouse reflection handling


dolfje:
I'm checking to reduce the reflection problems with the crazyflie.

I already coded and looked to support multiple sweeps per frame and check the difference between the previous frame. This as thought that a reflection will normally make the sweep somewhere totally different. This didn't worked as smooth as thought. It improves, but if it suddenly cuts to the other sweep, it is a big problem.

So I have another thought, normally reflection will always reduce the strength of the signal. So my question: Is it possible to get the signal strength from the fpga? I tried to understand the scala code, but I fear that you can't get the strength as it is a digital pin?

krichardsson:
Hi!

Are you working with lighthouse 1 or 2?

The signal strength is unfortunately not available.

There is another flavour of this in the kalman filter where we have an outlier filter that tries to remove samples that seems to be wrong, base on the current estimated position.
https://github.com/bitcraze/crazyflie-firmware/blob/master/src/modules/src/kalman_core.c#L612

dolfje:
I'm working with lighthouse 1 (2 bs) and at the moment 18 crazyflies. Looking forward to upgrade to the bolts eventually.

Filtering in the Kalman filter is definitely handy, but doesn't solve this use case with fading led lights and/or reflections. e.g. if you have a reflection through a window, usually all sensors will pick up the reflection and thus if you only have one bs coverage at that location, then you have lost all location tracking. So then no angles will be provided to the kalman filter.

The signal strength would be the easiest way.

But I have an update in my mind for the selection of the angle in the same sweep frame.  I think my problem is that the premise: 'sweeps are far from the normal sweep is bogus' is incorrect. In my test location is the y-axis probably very close, so that jumped probably.

dolfje:
After some more bugfixes, I have a fix for reflection handling. It can support up to 3 reflections. But it uses the knowledge that the sweep will not be in the neighborhood of the current one. It can start only when it has only one sweep in the frame and will fail if it loses track of the correct sweep.

krichardsson:
See pull request #594

PR #594, opened 2020-06-19
#592: Lighthouse reflection handling

dolfje:
When only one sweep is detected in the frame, then nothing will be adapted. But if multiple sweeps are detected, this is probably because of a reflection of the sweep. In that case we check what sweep is closest to the previous calculated angle and choose that. 

caveat: once the sweep is lost (so the sweep isn't seen once), then we have no fallback and can only reset when only one sweep is in the frame anymore. So this algo works as longs as he has vision of the bs. Cannot be started when the system already has already multiple sweeps.

### DIFF
--- a/src/modules/src/lighthouse/lighthouse_core.c
+++ b/src/modules/src/lighthouse/lighthouse_core.c
@@ -196,10 +196,10 @@ static void usePulseResult(pulseProcessor_t *appState, pulseProcessorResult_t* a
 
     switch(estimationMethod) {
       case 0:
-        usePulseResultCrossingBeams(appState, angles, basestation, axis);
+        usePulseResultCrossingBeams(appState, angles, basestation);
         break;
       case 1:
-        usePulseResultSweeps(appState, angles, basestation, axis);
+        usePulseResultSweeps(appState, angles, basestation);
         break;
       default:
         break;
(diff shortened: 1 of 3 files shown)


krichardsson:
Thanks for the pull request!
We will unfortunately not have time to look at before vacations, but we'll check it out when we're back.

krichardsson:
Hi!

I have looked at this and I did unfortunately not manage to setup an environment with reflections that caused any problems with the standard firmware. I assume your PR solves problems in your environment though, so I think it would be a nice addition to the system.
I have a slight modification request though. The `timestampsHistory` is a member of the `pulseProcessorBaseStationMeasuremnt_t` type which is used by both the V1 and V2 implementations. It would be nicer to keep the reflection stuff in V1 code only. As far as I can see, the `timestampsHistory` array is only used in the `filterReflections()` function (correct me if I'm wrong). I guess the reason for including it in `pulseProcessorBaseStationMeasuremnt_t` mainly is to be able to log it. Correct?
If the logging is removed, I assume it could even be a local array in `filterReflections()`?

dolfje:
Hi, It is definitely helping in my environment. My lighthouse is located next to a window, so that is reflected. I could of course move it away from there. But I try to have the hardest environment in testing, so that it is better in real life.

I will have no time in the coming weeks, but will try to adapt your modifications. Although the array cannot be local in filterReflections only (the values are recorded in storeSweepData and then used in filterReflections). A global array in the v1 file would work.

krichardsson:
Looking forward to the modifications :-)

krichardsson:
Hi! Do you still intend to finalize this PR or should we close it?

dolfje:
Wow times fly.
I still intend to finalize it. But I can open a new PR when needed.
As I guess the base code has also been adapted a lot...

krichardsson:
Yes, the code base is moving quite a bit now as we are trying to finalize a first official version of the lighthouse system
I'll close the PR for now

Thanks!

Issue #595, opened 2020-06-19
Add ledMode to view valid lighthouse angles


dolfje:
When I'm debugging the lighthouse code, it would be easy to view how much sweeps are registered by the angles. Also when visiting a new performance room, it would be easy to go through the room and can get a sense of what areas are problematic for the lighthouse setup (reflections, lh range, problems with smoke in theater)

There it would be nice to have a visual indication for the receiving angles on the drone itself. And that you then can go around the place and see the led change. So when we have 16 good sweeps (2 bs * 4 sensors * 2 axis), the led ring shows green and when we don't receive anything, the led ring shows red.

PR #596, opened 2020-06-19
#595 Add ledMode to view valid lighthouse angles

dolfje:
This will add a new ledmode that will show how many lighthouse angles there are valid. So green if it receives 2 bs * 2 axis * 4 sensors. And red if it doesn't receive any angles. 

For that, I had to change the getUartFrame to be non-blocking and add a lot of sensor value invalidation to make sure that in all cases the data was adapted to be correct.

### DIFF
--- a/src/deck/drivers/src/ledring12.c
+++ b/src/deck/drivers/src/ledring12.c
@@ -708,6 +708,29 @@ static void rssiEffect(uint8_t buffer[][3], bool reset)
 }
 
 /**
+ * An effect that shows the status of the lighthouse.
+ *
+ * Red means 0 angles, green means 16 angles (2 basestations x 4 crazyflie sensors x 2 sweeping directions).
+ */
+static void lightHouseEffect(uint8_t buffer[][3], bool reset)
+{
+  static int validAnglesId = -1;
+  
+  // lazy initialization of the logging variables
+  if (validAnglesId == -1) {
+    validAnglesId = logGetVarId("lighthouse", "validAngles");
+  }
+
+  uint16_t validAngles = logGetUint(validAnglesId);
+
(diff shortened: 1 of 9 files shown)


krichardsson:
Thanks for the pull request!

I like this idea (as well)!

I have some comments though that needs to be fixed before we can merge the PR.

This implementation is related to LH V1 only, which is fine, but I would like a solution that can be expanded easily to LH2 as well. We should also take into account that LH2 will support more than 2 base stations in the future, and for this reason I suggest the following:

1. Do all processing that is related to LH1 in the the pulse_processor_v1.c file.
1. Add a clean API (a function or an exposed variable) to pulse_processor.c file where other modules (the ledring12.c driver in this case) can query the pulse processor for the current reception quality. Pulse_processor.c would get the data from the appropriate active implementation (LH v1 or v2).
1. It is probably better that the reception quality is something like a float in the range 0.0-1.0, or an int 0 - 100, or 0 - 255. This would remove the knowledge about base stations and sensors from the led ring driver, and support future extensions.
1. Don't use logGetUint() in the LED ring driver. I can see that it has been done before, but I think it is better to expose a variable or function in the pulse_processor instead. The drawback with logGetUint() is that errors are not caught in compile time.
1. IMHO there are a bit too many things going on in the lighthouseCoreTask, but that could always be debated.

Thanks!

krichardsson:
Merged.

* Fixed some broken unit tests.
* Removed the quality measurement for LH2. It did not seem to work and it is likely that the implementation will move in the near future, it is probably easier to reimplement it when LH2 is more stable.

Thanks!

krichardsson:
I will also add the LED effect to the client

Issue #579, opened 2020-05-07
Memory allocation for heavyweight optional modules


jpreiss:
I understand that memory pressure in the firmware is rising. #574 peer_localization reserved memory for a hard-coded maximum number of peers. #567 will need another O(n) allocation for the collision avoidance algorithm. The high-level commander also reserves around 4KB for uploaded polynomial trajectories.

We will probably see more contributions of memory-heavy but optional subsystems in the future. How will we deal with this? (Aside from releasing Crazyflie 3 with 32MB of ram 😉) As far as I can tell, the two main options are:

### Dynamic allocation
Modules init with no memory allocated. If the user sends an appropriate command over the radio (perhaps a param change, or a new kind of packet), the module can allocate memory.

I see `pvPortMalloc` a few places in the codebase, but it does not seem to be part of the standard style. If it helps, we could avoid the complexity of heap fragmentation and free lists by imposing that all optional modules can only allocate memory once and never free it.

**Benefits:**
- Modules consume almost no memory until they are used.
- Supports e.g. a lab where Student A wants Module 1 on Monday and Student B wants Module 2 on Tuesday. They can do experiments without constantly reflashing firmware.

**Drawbacks:**
- Code is more complex and harder to reason about.
- Anything pushed from compile time to runtime is extra difficult because you can't debug while flying.
- Lost optimization opportunities because addresses are no longer constants.

### Recompiling firmware
Currently, each module tries to statically allocate enough memory to satisfy the typical user. This cannot last as more modules are added. Instead, modules could statically allocate the minimal amount of memory needed to be useful. For example, peer localization with ~2 other Crazyflies instead of 10, trajectory uploads of ~4 pieces instead of 32. Then, users who want to use the module more extensively can increase its allocation and recompile. Or, entire modules could be conditionally compiled.

**Benefits:**
- Code is more amenable to static analysis.
- Fewer runtime state changes.
- No fallback or error handling for failed allocations.

**Drawbacks:**
- Forces otherwise Python-level users to get involved with C.
- Flashing firmware becomes very tedious for large number of Crazyflies.

For me, it is not obvious which one is better. Is there some prior discussion about this?

krichardsson:
Yes, memory is starting to become an issue (also see #546), and as you mention there are a few new (nice) modules on the way that requires bigger chunks of data.

[...]

PR #598, opened 2020-06-19
fixes #597: Add ledmode to support an uploaded light track

### LINKED ISSUE #597: Add ledmode to support an uploaded light track (like position trajectory, but for light)
When doing a performance with the drone, it would be nice to also adapt the lights on timing. Just like when you set a trajectory of positions, a sequence of lights that the drone should go through. You can upload it just like a trajectory and then when enabling that ledmode, it start to play that sequence.

dolfje:
This adds a ledmode to upload a led sequence that you want to display on the drone. Whenever you enable that ledmode it will go through that uploaded sequence.

### DIFF
--- a/src/deck/drivers/interface/ledring12.h
+++ b/src/deck/drivers/interface/ledring12.h
@@ -11,19 +11,26 @@
 #endif
 
 #ifndef LEDRING_TIME_MEM_SIZE
-#define LEDRING_TIME_MEM_SIZE 50
+#define LEDRING_TIME_MEM_SIZE 10
 #endif
 
-typedef struct timings
+extern uint8_t ledringmem[NBR_LEDS * 2];
+
+typedef struct __attribute__((packed)) timing
 {
   uint8_t duration;       // How long this color should be show in parts of 1/25s. So 25 will show the color 1s, before going to the next color.
   uint8_t color[2];       // What color should be shown in RGB565 format.
-  unsigned leds:5;        // What led should be show, 0 equals all leds.
-  bool fade;              // Fade from previous colour to this colour during the full duration.
-  unsigned rotate:2;      // Speed of the rotation, number of seconds per revolution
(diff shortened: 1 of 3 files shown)


krichardsson:
Thanks for the pull request!

I like the idea of a sequencer for the LED ring and I think the implementation looks OK, but there are some minor things I would like to fix before accepting it.

There is a group of things are related to the size of the ledringtimingsmem. 

First of all it is better to use a constant/define, especially since the magic number 3000 is duplicated in two files.

Secondly I would like to have a smaller default size of the memory. RAM is a limited and shared resource and as long as we use static memory allocation for this type of applications (see #579 for more info) I think it is better to keep it fairly small. I think 50 events could be a reasonable size, it would enable users to try it out, but those that want to do more advanced sequences will have to modify the firmware. 

You could set the size through a define that can be set from the config.mk file or command line at build time to make it easy to build a FW with larger sequence memory without modifying the code.

I noted that you are using 3 bytes for each event in the ledtimings type. The compiler will word align the structs and use 4 bytes/event anyway, so if you want to you can use the fourth byte for something useful.

BTW we don't require a github issue for a pull request. The pull request it self is an issue and we're happy to use that as it is. There is no harm in creating an issue first though :-)

Thanks!

dolfje:
The feature is now code-complete and we use the extra bit to support fading, rotation and choosing what leds that should change.
Also an update to the library: https://github.com/bitcraze/crazyflie-lib-python/pull/158

knmcguire:
I would like to try this out for review, but it is probably better to first fix the python lib first :) I have a comment on that pull request (bitcraze/crazyflie-lib-python#158)

dolfje:
The lib is already fixed, was just some styling issues (length of the line to big)

krichardsson:
Merged

I fixed a problem with the memory mapping. The OW_FIRST_ID must be the last ID in the list for it to work, which might not be obvious.

Thanks!

PR #628, opened 2020-10-15
Collision avoidance

jpreiss:
Implements onboard collision avoidance algorithm following the design detailed in #567. Here is [a video handling a high-conflict scenario](https://drive.google.com/file/d/16SLjO568eADFKynYasulYrEGXV6cmfHd/view?usp=sharing).

Positions of other Crazyflies **on the same radio channel** are obtained from the peer localization service (#574). Collision avoidance uses the buffered Voronoi collision avoidance (BVCA) method [1]. Collision avoidance is enabled via the parameter system, so no new radio packet is needed.

Supports any position- or velocity-based control mode. I tested with high-level `goTo`, low-level `cmdPosition`, and low-level `cmdVelocityWorld`. All modes work on real hardware performing the same maneuver shown in the video.

The **Mellinger controller does not work** with collision avoidance as of now, because it does not behave gracefully when the setpoint is far from the current position [2]. However, the PID controller handles it well. I did not test the INDI controller.

The corresponding Crazyswarm update is in the branch https://github.com/jpreiss/crazyswarm/tree/bvca.

I am having trouble rebasing this branch for some reason - **please use "squash and merge"** to collapse the commits.

@whoenig @Zhenghao-Dai 

---

[1] Zhou, Dingjiang, et al. "Fast, on-line collision avoidance for dynamic vehicles using buffered voronoi cells." IEEE Robotics and Automation Letters 2.2 (2017): 1047-1054.

[2] In my opinion, we should consider the Mellinger controller as an aggressive trajectory-tracking controller, not a general-purpose controller.


krichardsson:
Merged. Thanks!

PR #648, opened 2020-11-16
Added crtpCommanderHighLevelIsTrajectoryDefined()

ntamas:
This PR extends the public API of the high-level commander with a function that allows the app layer to determine whether the trajectory at a given trajectory slot is defined.

This could be used by apps that are supposed to start flying a trajectory that is to be uploaded by the user to wait until the trajectory is indeed defined by the user via CRTP commands.


krichardsson:
Merged. Thanks!

PR #651, opened 2020-11-18
Added takeoff/land with average velocity commands to public API of the HL commander

ntamas:
The public API of the HL commander missed the entry points for the "takeoff with average velocity" and "land with average velocity" commands. Now they are there.


krichardsson:
Sorry missed those, thanks!

PR #657, opened 2020-11-25
added isCharging(), isBatteryLow() and isChargerConnected() to PM API

ntamas:
This PR adds three new functions to the power management API to make the list of query functions more complete.


krichardsson:
Merged. Thanks!

PR #658, opened 2020-11-25
added feature to trigger bright flashes on the LED ring to identify a drone in a swarm

ntamas:
As discussed with @krichardsson in email.

This PR adds a new parameter to the LED ring that allows the user to trigger five bright white flashes on the LED ring, irrespectively of what mode the LED ring is set to. The light signal can be triggered by setting the parameter to a non-zero value; the drone sets it back to zero when the signal was processed.

The purpose of this PR is to make it easier to work with multiple Crazyflie drones in a swarm as it can be used to draw attention to a particular drone from a ground control station; for instance, it can help to identify which (physical) drone has to be replaced before a flight if we see on the GCS that the battery voltage is too low.

In email, we were talking about using a different name for the parameter as the feature could be generalised as an "identification" signal that other decks or the built-in LED may also respond to. I'll just submit the PR as is, but feel free to rename the parameter to something more appropriate if you find a better name for it. (Naming is hard).

### DIFF
--- a/src/deck/drivers/src/ledring12.c
+++ b/src/deck/drivers/src/ledring12.c
@@ -961,6 +961,45 @@ Ledring12Effect effectsFct[] =
   lighthouseEffect,
 };
 
+/********** Light signal overriding **********/
+
+static struct {
+  uint8_t trigger;
+  uint32_t triggeredAt;
+  uint8_t active;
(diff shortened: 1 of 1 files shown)


krichardsson:
Merged. Thanks!
I'll change the parameter name in a separate commit

PR #659, opened 2020-11-25
disable MISC_VALUE_UPDATED CRTP messages on-demand with a compile-time switch

ntamas:
As discussed with @krichardsson .

This PR allows the user to prevent the Crazyflie from dispatching MISC_VALUE_UPDATED messages on the CRTP param port when a parameter value is updated from within the firmware (i.e. not from the GCS via a CRTP command). It can help to reduce radio traffic generated by certain Crazyflie apps if the app updates many parameter values frequently; for instance, when the app drives the color of the LED ring in a tight loop.

This is an advanced feature and it breaks synchronization of parameter values between a GCS and the drone if the GCS relies on local caching of parameter values.

### DIFF
--- a/src/modules/src/param.c
+++ b/src/modules/src/param.c
@@ -728,7 +728,9 @@ void paramSetInt(paramVarId_t varid, int valuei)
       break;
   }
 
+#ifndef SILENT_PARAM_UPDATES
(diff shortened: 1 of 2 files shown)


krichardsson:
Merged. Thanks!

PR #660, opened 2020-11-26
dev/robust kf

Williamwenda:
An M-estimation based robust kalman filter update for UWB TDOA2. Need a proper initialization in kalman_core.c (initialX,Y,Z), so that the estimator can reject measurement outliers during takeoff. Hyperparameters for the robust cost functions are the sigmas in function GM_UWB and GM_state (in kalman_core).


knmcguire:
Hi! Sorry for the late request but is this implementation based on a paper? Could you share that one with us?

Williamwenda:
Hi McGuire,

The paper is still under the process of ICRA21-RAL and the robust KF is
only a part of the content. Therefore, I would like to
share the document I summarized for robust (M-estimation) KF instead, which
provides the details about
the M-estimation KF derivation.

The document is attached. Hope it will help understand the code I
implemented.

Kind regards,

Wenda

knmcguire:
Hi!

The paper didn't attach to the pull request comment unfortunately, but if you can send it to contact@bitcraze.io? I will try to look at tomorrow.

Williamwenda:
Hi McGuire,

I have sent the document to contact@bitcraze.io. Please check.

Wenda

knmcguire:
Thanks! I've received it. Its good to see the theory for the implementation.

I've understood that this pull request is not to be merged but meant to demonstrate how you have implemented it? I'll try to give some suggestions on how to improve this:

**Robust EKF**
kalmanCoreRobustUpdateWithTDOA() seems to be one which implements the code from your document. And it seems to check out, although I would  suggest try to add the equestions of the pseudocode in some kind of latex format as well for more clarity. can see that its also doing the the full update too, which is handled by the previous tdoa update code with scalarUpdate. Is this because it is not compatible for what you want to do? Maybe it would be good to make a seperate function called vectorUpdate to make it more generic and shorten the kalmanCoreRobustUpdateWithTDOA (its quite big). 

**Math help  functions**
I've noticed that there are a couple of matrix help functions in kalman_core now. There is actually already an header file with many matrix functions already called cf_math.h or math_3d.h. Maybe it would be good to reuse functions there or copy the functions there if they are missing. I do see that you are using the arm_math. so that is good :) We have put some of our own arm math functions in cf_math.h. Its good to know that their function copy over the source matrix...

**Iteration**
Currently this is the line for the iteration:
`        for (int iter = 0; iter < 2; iter++){
`
The number 2 stands for the maximum iteration however hardcoding like this would not be good. You could consider turning it into a parameter. 

**Unit tests**
I see that you are using unit tests to test the matrix multiplications. This is actually something that we don't really have but should definitely add as well. We have in the folder 'test' several unit test functions, so after this has been implemented we should definitely add this to that to make sure that future implementation does not break yours. However, this is a last step in the implementation I think. 

_These might not have that much to do with the kalman filter implementation but if you wanted to merge it as well._
**Measurement types**
Currently I see a lot of positionYawMeasurement or posvelyaw in the stabilizer_type.h. I think it might be better to include euler angles and velocity to the current poseMeasurement_t structure instead. I know that technically Pose is usually only position and orientation but I think we can stretch the definition a little more here. 

However, I do have a comment on the kalmanupdates that might argue to actually split it up instead...

**Kalman updates**
So what I see in the code is that in crtp_broadcast_service.c in getExtPosVelYawBC(), first position, then velocity
and then yaw is collected and then all sent with estimatorKalmanEnqueuePosVelYaw() to the kalman filter measurement models where they are all handled at the same time kalmanCoreUpdateWithPosVelYaw() in kalman_core.c. 

What I would propose is to use smaller update functions instead. So in getExtPosVelYawBC() in crtp_broadcast_service.c (_italic_ what needs to be made new for this):
1 - Get position, put it in a positionMeasurement_t struct and send it through estimatorKalmanEnqueuePosition().... This already exist so nothing else needs to be done here.
2 - Get velocity, put it in a _velocityMeasurement_t_ struct and send it through  _estimatorKalmanEnqueueVelocity()_. Then through _kalmanCoreUpdateWithVelocity()_ do the scalar update. (the content of this can be copied from kalmanCoreUpdateWithPosVelYaw)
3 - Get the yaw, put it in a _yawMeasurement_t_ struct and sent it through  _estimatorKalmanEnqueueYaw()_. Then through _kalmanCoreUpdateWithYaw()_ do the scalar update, of which kalmanCoreUpdateWithPosVelYaw can also be used as inspiration.

User smaller functions for the updates will also make it more generic and will make it easier for future users to use it as well, then if instead you have an update for all possible measurement combinations: pose, pos + vel, pos + yaw, vel + pitch.... anyway you get the idea.

This also means that  would need to think about splitting up the kalmanCoreUpdateWithPose, but that is maybe just a thought for cleaning up later, not for your implementation.

Williamwenda:
Thanks for reviewing the code! This pull request is intended to be merged
finally as a robust estimation method, especially for UWB sensors.
As discussed with Kristoffer before, I created this pull-request so that
you can test the estimation performance in your lab setup. I have tested
this code in my lab (UWB) setups and it works correctly. If everything
works well in your lab, I can improve the code so that it can be directly
merged into the master branch (under your guidance).

A few answers to the questions you posted:
(1) The kalman filter updated function is based on *Joseph form covariance
update, *which is slightly different than the conventional update
equations. My derivation of the robust KF is based on the
conventional Kalman filter derivation. Therefore, I implemented the entire
update function in the code. The Joseph form covariance update function
might work for robust KF as well, but need to be further justified.
(2) I am glad to put the Math help functions to kalman_core.c, adjust the
iteration time into a parameter, and further clean up the code.
(3) Do not worry about the Crazyradio communication code. That is the
customized code for our lab to send commands and Vicon measurements from
our ground station. It can be omitted when merging to the master.

Finally, if this code can be merged into the master branch, we would like
to add a title in front to indicate that the code is implemented by our lab
following our paper.

Please let me know how the code performs in your lab and how would you want
to improve the code for merging.

Thanks,

Wenda

knmcguire:
Hi! Sorry we had to find some time to test this in our flightlab. Here are two videos of the performances of a CF2.1 with only an LPS deck

**Our EKF (ROBUST = false)**
https://user-images.githubusercontent.com/11061260/106160694-63a1eb80-6186-11eb-9447-a5b9ef0689a4.mp4

**New EKF (ROBUST = true)**
https://user-images.githubusercontent.com/11061260/106160781-7ddbc980-6186-11eb-92ae-1bf33ed865b6.mp4

Before making another pull request, I would say to keep discussing indeed with kristoffer first. Internally we are also trying to figure out how to setup the kalman filter code to be able to add this nicely.

Williamwenda:
Thank you for testing the code. I will discuss with Kristoffer and try to organize the code better.

krichardsson:
Hi! I suppose this PR is replaced by #707? I'll close this, please let me know if I missed something and it should remain open.

PR #665, opened 2020-12-10
fix variance calculation

Leonana69:
The old variance calculation is wrong, it should be v=E(x^2) - (E(x))^2

### DIFF
--- a/src/hal/src/sensors_bmi088_bmp388.c
+++ b/src/hal/src/sensors_bmi088_bmp388.c
@@ -726,13 +726,13 @@ static void sensorsCalculateVarianceAndMean(BiasObj* bias, Axis3f* varOut, Axis3
     sumSq[2] += bias->buffer[i].z * bias->buffer[i].z;
   }
 
-  varOut->x = (sumSq[0] - ((int64_t)sum[0] * sum[0]) / SENSORS_NBR_OF_BIAS_SAMPLES);
-  varOut->y = (sumSq[1] - ((int64_t)sum[1] * sum[1]) / SENSORS_NBR_OF_BIAS_SAMPLES);
-  varOut->z = (sumSq[2] - ((int64_t)sum[2] * sum[2]) / SENSORS_NBR_OF_BIAS_SAMPLES);
+  meanOut->x = (float) sum[0] / SENSORS_NBR_OF_BIAS_SAMPLES;
+  meanOut->y = (float) sum[1] / SENSORS_NBR_OF_BIAS_SAMPLES;
+  meanOut->z = (float) sum[2] / SENSORS_NBR_OF_BIAS_SAMPLES;
 
-  meanOut->x = (float)sum[0] / SENSORS_NBR_OF_BIAS_SAMPLES;
-  meanOut->y = (float)sum[1] / SENSORS_NBR_OF_BIAS_SAMPLES;
-  meanOut->z = (float)sum[2] / SENSORS_NBR_OF_BIAS_SAMPLES;
+  varOut->x = sumSq[0] / SENSORS_NBR_OF_BIAS_SAMPLES - meanOut->x * meanOut->x;
+  varOut->x = sumSq[1] / SENSORS_NBR_OF_BIAS_SAMPLES - meanOut->y * meanOut->y;
+  varOut->x = sumSq[2] / SENSORS_NBR_OF_BIAS_SAMPLES - meanOut->z * meanOut->z;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


Leonana69:
Also, the variance function in stabilizer.c is also wrong.

tobbeanton:
Nice find, thanks! 
There are threshold values depending on the wrong calculations that would be effected and these needs to be looked at together. The effected values are:
```
// Variance threshold to take zero bias for gyro
#define GYRO_VARIANCE_BASE              10000
#define GYRO_VARIANCE_THRESHOLD_X       (GYRO_VARIANCE_BASE)
#define GYRO_VARIANCE_THRESHOLD_Y       (GYRO_VARIANCE_BASE)
#define GYRO_VARIANCE_THRESHOLD_Z       (GYRO_VARIANCE_BASE)
```
So before we can accept this pull request the threshold values needs to be re-tuned. What we used as a guide before was that it should not be possible to go under the threshold while holding the crazyflie in your hand. It would have to be placed on a stable surface.

Leonana69:
> Nice find, thanks!
> There are threshold values depending on the wrong calculations that would be effected and these needs to be looked at together. The effected values are:
> 
> ```
> // Variance threshold to take zero bias for gyro
> #define GYRO_VARIANCE_BASE              10000
> #define GYRO_VARIANCE_THRESHOLD_X       (GYRO_VARIANCE_BASE)
> #define GYRO_VARIANCE_THRESHOLD_Y       (GYRO_VARIANCE_BASE)
> #define GYRO_VARIANCE_THRESHOLD_Z       (GYRO_VARIANCE_BASE)
> ```
> 
> So before we can accept this pull request the threshold values needs to be re-tuned. What we used as a guide before was that it should not be possible to go under the threshold while holding the crazyflie in your hand. It would have to be placed on a stable surface.

I made some test, the value will be less than 100 if the drone is placed stably. If held by the hand, the value will be more than 1000. You can try it, I think 100 works fine.

knmcguire:
Hi! Just a quick message about this, we are already on our christmas holiday so probably we are not going to look at this PR until January. Happy Holidays!

whoenig:
I had a look and I am not sure how the new calculation is different. We currently have:

sum(x^2) - (sum(x))^2/num

The new equation uses:

  sum(x^2)/num - (sum(x))^2/num^2
= sum(x^2) - (sum(x))^2/num

Thus, the computed variance values should be identical. Therefore, the threshold should stay the same, too. I agree that the new computation is a bit closer to the textbook definition, so we could accept this PR if the threshold is not going to be changed and the additional debug print will be removed.

knmcguire:
Hi! There are currently conflicting files with this pull request... As the calculation is not different to what already is implemented then I'll close this pull request. Please reopen if there is anything new!

bttner:
I see that this issue is closed but it is actually true that the current variance calculation is wrong. @whoenig simply did a calculation mistake as (with the exception when num = 1; != means not equal)

sum(x^2)/num - (sum(x))^2/num^2 = [sum(x^2) - (sum(x))^2/num] / num != sum(x^2) - (sum(x))^2/num

Attached you can find a Python script that compares the calculation with Numpy and shows that one needs to divide again by the number of samples to get the uncorrected sample variance. Thus, @Leonana69 is right.

To give an example. The new variance threshold (_GYRO_VARIANCE_BASE_) for _sensors_bmi088_bmp388.c_ would be 20 instead of 10,000, which is much more reasonable considering that the range of the sensor is -32,767 to 32,767. 

Well, I can help to do the appropriate changes if it is decided to reopen the issue.

[variance.zip](https://github.com/bitcraze/crazyflie-firmware/files/7260732/variance.zip)

whoenig:
Good point (and very embarrassing for me!). I also think this should be re-opened and fixed. This is also related to other inconsistencies, see #849. It would have more scientific value if it would be in real units (i.e., rad²/s²), though. Otherwise, we are just off by a linear scaling in an arbitrary unit and could fix this simply by documenting the code.

bttner:
Another way of fixing it would be to call it _pseudo inverse_ or something along these lines. Otherwise, it is very misleading, at least it was for me.

jonasdn:
It is not possible to re-open this because the form has been removed, anyone up for recreating this?  @bttner ?

knmcguire:
Ah sorry I totally missed this. This was before the handy notification feature of github so this got lost in my emails.

Yes let's reopen it! but since the original PR is gone, the diff still exists. The original PR had merge conflicts anyway.

knmcguire:
btw @jonasdn, @Leonana69 was the OP of this PR. But I can start a new one

jonasdn:
Got for it. Just reference this PR from the new one.

PR #667, opened 2020-12-21
App API of high-level commander now returns error codes

ntamas:
This PR updates the high-level commander functions that are part of the app API to return the error code that the underlying `handleCommand()` call returns so the app can detect if a high-level commander function failed.


knmcguire:
Hi! I tried out the code and just compiled  the apps (app_api and swarm demo) that  use these functions to see if they break, and they didn't. So I'll merge it! Thanks and happy holidays!

PR #671, opened 2021-01-13
sensfusion6: Rename quaternion components for clarity.

### LINKED ISSUE #654: Quaternion component naming inconsistency
In `sensfusion6.c` [here](https://github.com/bitcraze/crazyflie-firmware/blob/master/src/modules/src/sensfusion6.c#L52) and [here](https://github.com/bitcraze/crazyflie-firmware/blob/master/src/modules/src/sensfusion6.c#L260) the real term of the quaternion is variously named `w` and `q0`.

But in `quaternion_t` in `stabilizer_types.h` [here](https://github.com/bitcraze/crazyflie-firmware/blob/master/src/modules/interface/stabilizer_types.h#L70),  `w` is overlaid on `q3`.

So far as I can tell, the only accesses to `quaternion_t` use `.w`, `.x` etc, so maybe the union with `.q0` etc isn’t needed?

whoenig:
Fixes #654.

Notes:
* This is just search & replace.
* Later on, we can look into using math3d and its quaternion definition & helper functions, although perhaps that is too much effort for 'legacy' code.

### DIFF
--- a/src/modules/src/sensfusion6.c
+++ b/src/modules/src/sensfusion6.c
@@ -49,10 +49,10 @@
   float integralFBz = 0.0f;  // integral error terms scaled by Ki
 #endif
 
-float q0 = 1.0f;
-float q1 = 0.0f;
-float q2 = 0.0f;
-float q3 = 0.0f;  // quaternion of sensor frame relative to auxiliary frame
+float qw = 1.0f;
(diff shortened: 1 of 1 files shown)


krichardsson:
Thanks!

PR #673, opened 2021-01-22
fix: sys.canfly variable should start from 'true' until we detect that the USB cable is connected

ntamas:
The Crazyflie has a neat log variable named `sys.canfly` that can be used to determine whether the USB cable is connected. Unfortunately, it starts from `false` and it does not get updated until the state of the power management subsystem changes at least once. The PM state starts from "battery", so in order to make `sys.canfly` consistent with the detected PM state, it should start from `true` and not `false` (as it is now).

This PR tries to be polite and not set `sys.canfly` to `true` until the system is started. Once it is started, it sets `sys.canfly` to `true` and enters the PM loop immediately; if the USB charger happens to be connected at boot, the first iteration will update `sys.canfly` to `false` so this is not a problem.

### DIFF
--- a/src/hal/src/pm_stm32f4.c
+++ b/src/hal/src/pm_stm32f4.c
@@ -315,6 +315,8 @@ void pmTask(void *param)
   pmSetChargeState(charge500mA);
   systemWaitStart();
 
+  systemSetCanFly(true);
+
   while(1)
   {
     vTaskDelay(100);


whoenig:
This seems like a great catch, but this also shows that there is another underlying issue: the actions on state-change are not executed if the CF powers up with the battery attached. I am not sure if that is intended or not. If not, another fix would be to add `invalid` to the `PMStates enum` and start out in the invalid state for `pmStateOld`.

ntamas:
I couldn't decide it either, that's why I went for the surgical solution with the most minimal change I could think of.

Actually, most of the logic related to the PM state switches affects at least three parts of the system (the LED sequences, the sound submodule and the "can fly" state variable), but it seems to be completely independent of the hardware that the Crazyflie is running on, so it should not be in an STM32F4-specific power management module; a better long-time solution for the PM module would be to provide events that other modules can subscribe to and act accordingly when the PM state changes. But this is definitely a larger chunk of work.

ataffanel:
I agree that there is much to do for the PM, but this is a nice fix so lets merge it. Thanks!

I have created a ticket to discuss how we want to improve the PM in the future: #718

Issue #681, opened 2021-02-05
CTRP stack might lose packets


whoenig:
The current CRTP stack implementation has several potential places in the firmware, where packets can be lost if an internal queue is full, because return codes are not always checked or blocking calls are not used. This issue is for tracking changes that improve this handling in the STM firmware.

ataffanel:
Thanks for the issue and the PRs! I renamed the title and ticket to reflect that packets are lost in the CRTP stack and not in safelink (safelink only protects the radio packets, and so far it works).

jonasdn:
Is this still an active issue? @ataffanel @whoenig ?

whoenig:
We can close this. The remaining work/bug is captured in #703.

PR #682, opened 2021-02-05
Enable uart-syslink asserts in release builds.

whoenig:
We need to be notified if this is a problem during real operation. If
we ever see cases where this is hit, proper flow control between NRF51
and STM32 are needed.

Part of fixes for issue #681.


ataffanel:
Thanks for the PR! Looks good, I am merging it.

PR #685, opened 2021-02-06
Switch crtp service and platform service to be tasks

whoenig:
These services respond to request and therefore should use the blocking
version of sendPacket that waits if the outgoing queue is full.

There are two other files that use the CRTP callback functionality
(commander and location service), but neither of them requires blocking
calls.

Part of fixes for issue #681.


ataffanel:
Thanks! merged.

PR #687, opened 2021-02-06
Use blocking enqueue calls for radiolink and crtp to avoid packet loss

whoenig:
Part of fixes for issue #681.


src/hal/src/radiolink.c
```diff
@@ -159,7 +159,8 @@ void radiolinkSyslinkDispatch(SyslinkPacket *slp)
   if (slp->type == SYSLINK_RADIO_RAW)
   {
     slp->length--; // Decrease to get CRTP size.
-    xQueueSend(crtpPacketDelivery, &slp->length, 0);
+    // Block, since we should never drop a packet
+    xQueueSend(crtpPacketDelivery, &slp->length, portMAX_DELAY);
```
ataffanel, line 163: Blocking here will make syslink silently ignore/drop packets sent by the nRF51. We should not block and assert if there is not space in the queue instead. Later, we should implement some flow control in syslink to block packet from making it into the STM32 and make the radio stop acking packets which will block packet from being sent to the Crazyflie, this will effectively implement flow control down to the Crazyradio.

ataffanel (changes requested):

PR #696, opened 2021-02-12
Add crtpsrv.echoDelay parameter

whoenig:
When echoDelay is greater than 0, a slow CRTP task can be simulated for
system tests of the link.

Helps debugging/testing of issue #681.


ataffanel:
Looks good! Merging!

Issue #699, opened 2021-02-18
Revamp CRC32 implementation


ataffanel:
We currently have two implementation of CRC32 in the Crazyflie and both have major problems:
 - crc.h: Does not compute a CRC compatible with the "standard" zlib crc32 and only works on memory buffers, cannot work on stream
 - crc_bosch.h: Works on stream and is generating the same crc32 as zlib but is not thread-safe, require every user to initialize a lookup table, and has a pretty hard to use API.

My proposal would be to use the core of the bosh implementation but thread-safe, with static lazy-initialized lookup table and with a more standard API:
```
 - void crcInitContext(crcContext_t *context)
 - void crcUpdate(crcContext *context, const char* data, uint32_t size)
 - uint32_t crcOut(const crcContext *context)
```

PR #706, opened 2021-02-25
Fix uSD CRC regression introduced in refactoring 9eb330c

whoenig:
Part of issue #699.

Tested with the provided CF_Functions.py decoding library part of the firmware tree.


ataffanel (approved):
Looks good!

PR #707, opened 2021-02-25
Pr/robust tdoa

Williamwenda:
An M-estimation based robust kalman filter update for UWB TDOA localization. Need a proper initialization in kalman_core.c (initialX,Y,Z), so that the estimator can reject measurement outliers during takeoff. Hyperparameters are the sigmas in the weight functions 'GM_UWB' and 'GM_state'. A general guideline for hyperparameter tuning is provided.
 
The code is tested to be working with 8 UWB anchors (TDOA2 mode). A testing video is linked below.
[video](https://drive.google.com/file/d/1PmnZhmQWFWryNk5wImV7jsyzR1ulc4k9/view?usp=sharing)


krichardsson:
Hi and thanks for the pull request!

I have limited access to our flight lab (due to covid) and it will probably taks some days (weeks) until I have the opportunity to try it out.

I had a quick browse through the code and it seems as there are a number of files that are changed but that are not related to the contributed functionality (CF_functions.py, some png files and lighthouse.bin). Please remove them from the pull request to make it clear what functionality that is added. 

I will check the code out a bit more and get back to you as soon as possible if I have any more comments.

Thanks!

Williamwenda:
We have restored the unrelated changes. Please check. Thanks

krichardsson:
Looks much better, thanks!

I'll try it out next time I'm in our flight lab

krichardsson:
Hi Wenda! 

I have tested your pull request and I think it needs some more work, some minor things that I can easily fix as a part of the merge, but I also found some stability issues that are more serious. 

1. Minor: the default initial position should be (0, 0, 0), it has been changed to (1.5, 0, 0)
2. Minor: the tdoa update is called instead of flow update (looks like a copy/paste error)
3. Major: I sometimes get an assert in cf_math.h, line 64. It seems to be a matrix inverse that is not OK. I do not know which matrix inv it is. I think I mainly could see this problem when the estimate was way off.
4. Major: the filter seems to diverge pretty easily. An example is
    1. I put the CF on the floor at (0, 0, 0) and let the filter converge
    2. Lift the CF up and move it to (-1.5, 1.5, 1.5) 
    3. Quite often the estimated position starts to drift off after a few seconds
    4. As I write this I realize that it might be related to that I move the CF by hand? Some other code in the EKF has special handling for when the CF flies or not.

Apart from the above, the code looks good :-)

krichardsson:
Hi Wenda! 

I have tested your pull request and I think it needs some more work, some minor things that I can easily fix as a part of the merge, but I also found some stability issues that are more serious. 

1. Minor: the default initial position should be (0, 0, 0), it has been changed to (1.5, 0, 0)
2. Minor: the tdoa update is called instead of flow update (looks like a copy/paste error)
3. Major: I sometimes get an assert in cf_math.h, line 64. It seems to be a matrix inverse that is not OK. I do not know which matrix inv it is. I think I mainly could see this problem when the estimate was way off.
4. Major: the filter seems to diverge pretty easily. An example is
    1. I put the CF on the floor at (0, 0, 0) and let the filter converge
    2. Lift the CF up and move it to (-1.5, 1.5, 1.5) 
    3. Quite often the estimated position starts to drift off after a few seconds
    4. As I write this I realize that it might be related to that I move the CF by hand? Some other code in the EKF has special handling for when the CF flies or not.

Apart from the above, the code looks good :-)

krichardsson:
I checked the CPU usage also
Normal TDoA: 8% CPU for the kalman filter
Robust TDoA: 17% CPU for the kalman filter

It is obviously using more CPU, but it is not too bad I think.

Williamwenda:
Hi Kristoffer,

Thanks for testing the code.
(1) The robust KF needs a proper initialization at the beginning so that we can reject the outlier measurements. But I guess we can do it by sending parameters to 1) reset the estimation in estimator_kalman.c, 2) set the initial x,y,z in kalman_core.c.
(2) Sorry for this mistake. In our lab, we use a slightly different crtp protocol so I tested the code first in our setup and then made the changes in the default firmware. Please fix this small error when testing. I will also update the pr commit for this issue.

(3)&(4)
First, this code should be used when the state is predicted by IMU measurements. If the CF is moved by hand, the dynamic model is different, then the predicted state would be different. 
In the estimator, we weighted the predicted state and the UWB measurement updates. By setting the hyperparameters 'sigma' in functions GM_UWB and GM_state, I trusted the predicted states more than the UWB measurements. With wrong predicted states, the UWB measurements will be rejected as outliers. Without any correction, the estimator will diverge.

This is in contrast to the conventional EKF as all the measurements are not rejected beforehand. With an accurate external sensor (like a motion capture system), the drone can have a very good estimation relying only on measurements. 

For the matrix inverse, it only happens in line 173 of 'mm_tdoa_robust.c'. Each time, I do Cholesky decomposition for the predicted covariance matrix and then invert the lower triangular matrix. I didn't see this error much in my lab, but I can do some tests next week. If the matrix inverse would be a problem when the states are far off, we can do some protections for that, like adding small values on the diagonal entries, etc.

In general, please test the code when the CF is commanded to fly and see if (3)&(4) still happens. 

It's great to see the CPU usage for the robust KF is not that heavy. 

One small issue. In my PC, when I clone the repo., the PNG files and the carriage return line feed (CRLF) in CF_functions.py, which seems to be written in a Windows OS, are automatically be replaced by a line feed (LF). That's the reason why the initial PR has those unrelated changes. One solution in my PC is to add  "*.png binary" in the .gitattribute file. Yet that is also a change and I pushed it in the PR. I am not sure how common this issue is and you can decide to keep this change or not.

Thanks,

Wenda

Williamwenda:
Some related links for the CRLF and LF errors are:
[git issue](https://github.com/puma/puma/issues/1672), [solution](https://github.com/puma/puma/pull/1715/commits/778d4610e7c14921c77a6913ddfcbcca2312d097)

krichardsson:
Thanks for the clarifications!

(1) and (4) My suggestion is that we add a parameter to switch between the normal and robust flavour of the filter and that the default setting is to use the normal filter. This will allow the filter to converge in most cases and users can then switch to the robust variation without problem. I can add this parameter when merging if you like. 

(2) No worries

(3) 
> If the matrix inverse would be a problem when the states are far off, we can do some protections for that, like adding small values on the diagonal entries, etc.

Yes, I think this is important. The failed matrix inverse triggers an assert (in the current implementation) which will re-boot the Crazyflie in mid-air.

krichardsson:
Thanks for the updates!

I was hoping that I would have the time to try it out yesterday when I was at our office, but unfortunately I didn't. I'll do it as soon as I'm back at the office again.

krichardsson:
Merged

I did some manual fixes:
1. The .gitattributes file was a bit weird and I removed the png line. There was an extra space that made git believe all files were binary
2. Added a parameter to switch between the "normal" and robust flavours

We are working on making a release and the plan is to incorporate this PR. 

Thanks for your hard work!

PR #708, opened 2021-02-28
FatFS: Fix Clang build (typedef redefinition with different types)

concatime:
Hi.
I’m trying to build the firmware using Clang.
I faced one trivial issue:
```
In file included from ./src/deck/drivers/src/usddeck.c:48:
In file included from ./src/drivers/interface/fatfs_sd.h:10:
In file included from ./src/drivers/interface/diskio.h:9:
./src/lib/FatFS/integer.h:31:23: error: typedef redefinition with different types ('unsigned long' vs 'uint32_t' (aka 'unsigned int'))
typedef unsigned long   DWORD;
                        ^
./src/lib/FatFS/ff.h:48:21: note: previous definition is here
typedef uint32_t    DWORD;  /* 32-bit unsigned integer */
                    ^
```
The issue is that both `ff.h` and `integer.h` define the same types.
This patch fixes it.

### DIFF
--- a/src/lib/FatFS/ff.h
+++ b/src/lib/FatFS/ff.h
@@ -33,29 +33,7 @@ extern "C" {
 #endif
 
 
-/* Integer types used for FatFs API */
-
-#if defined(_WIN32) /* Main development platform */
-#define FF_INTDEF 2
-#include <windows.h>
-typedef unsigned __int64 QWORD;
-#elif (defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L) || defined(__cplusplus)  /* C99 or later */
-#define FF_INTDEF 2
-#include <stdint.h>
-typedef unsigned int  UINT; /* int must be 16-bit or 32-bit */
-typedef unsigned char BYTE; /* char must be 8-bit */
-typedef uint16_t    WORD; /* 16-bit unsigned integer */
-typedef uint32_t    DWORD;  /* 32-bit unsigned integer */
-typedef uint64_t    QWORD;  /* 64-bit unsigned integer */
-typedef WORD      WCHAR;  /* UTF-16 character type */
-#else   /* Earlier than C99 */
-#define FF_INTDEF 1
-typedef unsigned int  UINT; /* int must be 16-bit or 32-bit */
-typedef unsigned char BYTE; /* char must be 8-bit */
-typedef unsigned short  WORD; /* 16-bit unsigned integer */
-typedef unsigned long DWORD;  /* 32-bit unsigned integer */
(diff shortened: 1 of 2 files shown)


whoenig:
FatFS is external code (see http://elm-chan.org/fsw/ff/00index_e.html). Therefore, it would be best to fix those issues upstream (by posting in the FatFS user forum I think) and then just update the version that is part of the Crazyflie firmware. Perhaps your issue has even been fixed already, since the version in this repository is a few years old.

While accepting the change here is possible, it will be destroyed the next time we upgrade FatFS, so this is not going to be a long-term solution.

concatime:
It seems that the issue was introduced in version `R0.13c`:
```
R0.13c (October 14, 2018)
  Supported stdint.h for C99 and later. (integer.h was included in ff.h)
```

See https://github.com/abbrev/fatfs/commit/f2bf030bfd94c55d72d507c493be63f89e2a9f5b#diff-54ec3d3ec0749a2a213fb4aff3999e6e3035fedf153e004ebe06ccbc3c820914.

As you can see [here](https://github.com/abbrev/fatfs/commit/f2bf030bfd94c55d72d507c493be63f89e2a9f5b#diff-33786978971856478469fbe0c2c86d31d97f680c457fad8b92569a90c70258dd), `integer.h` was completely dropped.

1. https://github.com/bitcraze/crazyflie-firmware/commit/46cf8db280be1c9988c6176829e8fcd3b737e4d4#diff-5093b86e1ec53c03f45175e0fd1d05130c877849813a00c7affee5532ca1926c upgraded `ff.c`, `ff.h` and `ffunicode.c` from `R0.12b` to `R0.14`.
2. Not all files were upgraded. https://github.com/bitcraze/crazyflie-firmware/blob/master/src/lib/FatFS/00history.txt#L269 is still at `R0.12b`.
3. `integer.h` was dropped in `R0.13c`, but is still preset in the repo because, as I said, not every files were upgraded.

So please, is it possible to do a proper upgrade? My PR is not relevant anymore.

Thank you.

concatime:
Also, https://github.com/ARM-software/CMSIS_5/commit/9b5b7561f98cb39f5407959d99c98bdd6eb0fa78 and https://github.com/ARM-software/CMSIS_5/commit/f46ec37262bc4fd82b6c91d24198cc3b4a499fe3 are required for Clang. Are you okay if I submit a PR to update the legacy CMSIS 4.5.0 to 5.7.0? I tested the build and the update process is quite trivial.

whoenig:
For FatFS, I created a new bug (#713, I think I'll get to this soon). I don't see a reason for not upgrading CMSIS.

PR #710, opened 2021-03-02
Makefile: Error out on missing submodules

jonasdn:
If you, like me, fail to read the README.md properly you will get a bit of a cryptic error when trying to build.

```
   $ make
   CLEAN_VERSION
   make[1]: *** No rule to make target 'list.o', needed by 'cf2.elf'.  Stop.
   make: *** [Makefile:380: build] Error 2
```

 This does not make it super clear that it is the missing submodules that is at fault.
 This PR adds a check for the vendor submodules and will produce an error if missing, like so:

```
   $ make
   Makefile:383: ***
    The submodules does not seem to be present, consider fetching them by:
    $ git submodule init
    $ git submodule update
    .  Stop.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -370,8 +370,23 @@ ifeq ($(SHELL),/bin/sh)
   COL_RESET=\033[m
 endif
 
-#################### Targets ###############################
+# This define n-thing is a standard hack to get newlines in GNU Make.
+define n
 
+
+endef
+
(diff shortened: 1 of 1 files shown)


jonasdn:
Another way to tackle this would be to just fetch the submodules if they are not there on `make`. Since a call to `make`implies that a user wants to build the firmware. And they can't without the submodules.

PR #716, opened 2021-03-03
I2C improvements

whoenig:
This PR includes several I2C improvements, mostly related to initialization and timing (w.r.t. to new compilers). For review it might be easiest to look at the 4 different commits in sequence, rather than all file changes at once.


tobbeanton:
The never ending I2C problem solving :-). It all looks good but would be nice with some scope picture showing the new timing of the bus unlocking before we merge it.

tobbeanton:
image
It now looks good and the bus unlocking fix was a nice catch.

PR #717, opened 2021-03-03
eprintf: Argument fmt should be constant

concatime:
This is required, otherwise, you would get, in C++ code, lot of:
```
warning: ISO C++11 does not allow conversion from string literal to 'char *' [-Wwritable-strings]
  DEBUG_PRINT("Waiting for activation ...\n");
              ^
```

More information [here](https://stackoverflow.com/a/20944858).

`printf`’s fmt argument is [constant too](https://linux.die.net/man/3/printf).

Thank you.

### DIFF
--- a/src/utils/interface/eprintf.h
+++ b/src/utils/interface/eprintf.h
@@ -29,7 +29,7 @@ typedef int (*putc_t)(int c);
  * @param[in] ... Parameters to print
  * @return the number of character printed
  */
-int eprintf(putc_t putcf, char * fmt, ...) 
+int eprintf(putc_t putcf, const char * fmt, ...)
     __attribute__ (( format(printf, 2, 3) ));
 
 /**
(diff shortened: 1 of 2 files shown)


whoenig:
Thanks! Would be interesting to learn more about your overall clang use-case/story at some point. It certainly improves code quality, but perhaps there is another motivation as well?

concatime:
> Thanks! Would be interesting to learn more about your overall clang use-case/story at some point. It certainly improves code quality, but perhaps there is another motivation as well?

My main goal was to add CMake as a build system. I have a _profound_ hate for GNU/BSD Makefiles (I’m fine with POSIX Makefile). It feels like a stockpile of hacks/glue code the whole time. Then, I wanted to add both GCC and Clang as backends to CMake. Finally, I wanted to to use C++/libstdc++ in a crazyflie project for university. I am 100% open to share my build system.

whoenig:
Sounds awesome - please keep us in the loop with cmake. The added advantage would also be that it is much more portable across platforms (if done right). We certainly have talked before about using CMake rather than a Makefile, but never got around actually working on it.

PR #726, opened 2021-03-11
Active marker deck docs: clarify difference between modes 1 and 2

whoenig:


ataffanel:
Looks good, thanks! Merging.

PR #727, opened 2021-03-11
Split BMI088 part of sensors_bmi088_bmp388 into backends

jonasdn:
Instead of having one complete driver for each BMI088 interface we split the BMI088 part into a I2C and a SPI backend. This way we have one driver implementing functionality and handling LOG and PARAM and the backends take care of interface details.

New structure:

 * sensors.c => tells us which init function (SPI/I2C) to use in ...
   * sensors_bmi088_bmp388.c => which defers bmi088 interface and device init to ...
     * sensors_bmi088_[i2c|spi].c => which sets up device information.

Two files are removed:
  `sensors_bmi088_spi_bmp388.c`
  `sensors_bmi088_spi_bmp388.h`

Three files are added:
  `sensors_bmi088_i2c.c`     - unique i2c device init
 ` sensors_bmi088_spi.c`     - unique spi functions and device init
  `sensors_bmi088_common.h`  - shared function definitions

Closes: bitcraze/crazyflie-clients-python#364


tobbeanton (commented):
Looks good, finally no duplication of code! 
I was thinking if we should somewhere print out what interface is used as this is removed now. I guess there is a chance someone creates a platform with the wrong driver selected and then it will be easier if that is printed. Like "BMI088: Using I2C interface".

jonasdn:
> Looks good, finally no duplication of code!
> I was thinking if we should somewhere print out what interface is used as this is removed now. I guess there is a chance someone creates a platform with the wrong driver selected and then it will be easier if that is printed. Like "BMI088: Using I2C interface".

Thanks for reviewing! Yes, that sounds useful. I added a commit.

whoenig (approved):
In general this looks great! The comments below are maybe more general things that we should discuss moving forward. They shouldn't prevent you from merging this PR.

src/hal/interface/sensors_bmi088_common.h
```diff
@@ -23,22 +23,18 @@
...
 
-#endif // __SENSORS_BMI088SPI_BMP388_H__
\ No newline at end of file
+void        bmi088_ms_delay(uint32_t period);
+
+bstdr_ret_t bmi088_burst_read(uint8_t dev_id, uint8_t reg_addr,
+                              uint8_t *reg_data, uint16_t len);
+
+bstdr_ret_t bmi088_burst_write(uint8_t dev_id, uint8_t reg_addr,
+                               uint8_t *reg_data, uint16_t len);
```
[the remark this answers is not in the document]
jonasdn: Yes, agreed.

tobbeanton:
The Bolt is not booting for some reason. Only the blue M3 LED turns on and then nothing more. I'm investigating it.

src/hal/src/sensors_bmi088_spi.c
```diff
@@ -0,0 +1,360 @@
...
+  // Enable SPI DMA requests
+  SPI_I2S_DMACmd(BMI088_SPI, SPI_I2S_DMAReq_Tx, ENABLE);
+  SPI_I2S_DMACmd(BMI088_SPI, SPI_I2S_DMAReq_Rx, ENABLE);
+
+  // Enable peripheral to begin the transaction
+  SPI_Cmd(BMI088_SPI, ENABLE);
+
+  // Wait for completion
+  // TODO: Better error handling rather than passing up invalid data
+  xSemaphoreTake(spiTxDMAComplete, portMAX_DELAY);
```
tobbeanton, line 154: When this line is executed the chip just resets. I have no idea why. The spiTxDMAComplete looks valid.

whoenig: It looks like those are initialized after DMA is enabled. Could it be that we can some sort of transactions before the semaphore is fully initialized?

tobbeanton: I don't think it is a reset but rather a task switch back to itself... It is all very strange and then it is most often memory related.

tobbeanton: Found it! It was the DMA interrupts that was missing.

tobbeanton:
I added the missing DMA interrupt functions to src/hal/src/sensors_bmi088_spi.c. It all worked well then and merged the PR.

jonasdn:
Thanks @tobbeanton !

PR #728, opened 2021-03-11
Lighthouse: Allow to set system type at runtime

ataffanel:
System type (Lighthouse V1 or V2) can be set with a parameter. The new
setting is stored in persistent storage and loaded on next statups.
When changing types, the current configuration (geometry and
calibration) is cleared from RAM.

By default lighthouse V2 is used if no setting exists in storage.


jonasdn (approved):
Look good, just some nits.

src/modules/src/lighthouse/lighthouse_core.c
```diff
@@ -171,6 +168,40 @@ void lighthouseCoreLedTimer()
...
+    case lighthouseBsTypeV1:
+      pulseProcessorProcessPulse = pulseProcessorV1ProcessPulse;
+      break;
+    case lighthouseBsTypeV2:
+      pulseProcessorProcessPulse = pulseProcessorV2ProcessPulse;
+      break;
+    default:
+      // Do nothing if the type is not in range, stay on the previous processor
+      return;
+      break;
```
jonasdn, line 183: nit: maybe remove the `break;`?  It might trigger unreachable code on some linters?

PR #731, opened 2021-03-13
Avoid double initialization of external interrupts.

whoenig:
Interrupts are initialized and distributed in exti.c. Deck drivers
should not re-initialize those interrupts.

Note that this change implicitly changes the priority of the locodeck
interrupt from NVIC_VERY_HIGH_PRI to NVIC_MID_PRI (as defined in
nvicconf.h).

This change also adds NVIC_ClearPendingIRQ calls in exti.c for
correctness.

### DIFF
--- a/src/deck/drivers/src/locodeck.c
+++ b/src/deck/drivers/src/locodeck.c
@@ -7,7 +7,7 @@
  *
  * Crazyflie control firmware
  *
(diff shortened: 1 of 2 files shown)


jonasdn (approved):
nice cleanup!

PR #732, opened 2021-03-15
Remove information about last control command from state estimator

whoenig:
The state estimator should only get raw sensor measurements as input.
Previously, the last control command was needed to determine if the CF
is flying or not. Since #723 (introduction of the supervisor module),
this is handled in a better way.


jonasdn (approved):
Looks good!

PR #733, opened 2021-03-16
Lighthouse crossing beam: improve result quality

whoenig:
The previous code reported the position of a single sensor in
lighthouse.{x,y,z}, rather than the position w.r.t. the center of mass
of the Crazyflie. This new code now averages over all sensors.

The lighthouse.delta variable gives an indication of the quality of the
result. This indication does not correlate with absolute accuracy (when
measured using a motion capture system as ground truth). However, it
provides a helpful indication on whether lighthouse.{x,y,z} is valid
(deltaLog>0) or not (deltaLog == 0).

### DIFF
--- a/src/modules/src/lighthouse/lighthouse_position_est.c
+++ b/src/modules/src/lighthouse/lighthouse_position_est.c
@@ -218,22 +218,25 @@ static float sweepStd = 0.0004;
 static float sweepStdLh2 = 0.001;
 
 static vec3d position;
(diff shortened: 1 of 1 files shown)


knmcguire:
I checked it and it looks good! Now the logging and workspace variable are nicely separated. The delta log is indeed handy to see if it is a valid measurement. I still see half a second outages but that is for another fix.

PR #734, opened 2021-03-17
Event trigger accuracy improvements.

whoenig:
* introduces version 2 of the recorded binary file, which uses a
  microsecond timestamp, rather than a millisecond timestamp. Older
  version-1 files can still be decoded in the same way as before.
* introduces usd.canLog readonly parameter to check if uSD card is
  inserted and configuration file is valid.
* support for more payload variables for event triggers.
* reduction of data loss when starting/stopping logging.

### DIFF
--- a/docs/userguides/decks/micro-sd-card-deck.md
+++ b/docs/userguides/decks/micro-sd-card-deck.md
@@ -52,7 +52,7 @@ The `config.txt` file will be read only once on startup, therefore make sure tha
 
 ## Data Analysis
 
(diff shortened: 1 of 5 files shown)


jonasdn (approved):
Looks good!

A thought, does enableLogging need to be declared as volatile? Could the compiler do something fishy?

whoenig:
I don't think the volatile is needed - at least we don't have it in other parts of the firmware that uses parameters for similar purposes.

Issue #722, opened 2021-03-08
State Estimator Sensor Push Redesign


whoenig:
The current design of the state estimators has multiple issues:

1. IMU sensor data is not pushed, but pulled by each state estimator. This has weird side effects, for example, the barometer logging variables are not valid if the state estimator does not need it. It also makes it harder to log the data that is put into the filter.
2. Each new state estimator needs to provide functions to enqueue sensor data, as well as keep track of the enqueued data separately. Most (advanced) state estimators will have similar implementations of such functionality.
3. The current way of enqueuing data in the Kalman filter loses the order of the sensor data, since separate queues for sensors are used. In general, the order might be important, especially for slower filters.

A possible solution to address these issues is:

* In `estimator.c` add a new unified queue:

```
struct measurement_s {
measurement_type_e type;
union {
  distanceMeasurement_t distance;
  positionMeasurement_t position;
  // ...
} data;
};
static xQueueHandle measurementsQueue;
STATIC_MEM_QUEUE_ALLOC(measurementsQueue, 50, sizeof(struct measurement_s));
```
* In `estimator.c`, we can now fill definitions for all the enqueue* functions to just push data in the queue. In all state estimators, we simply extract items from the queue and ignore measurements that we do not want to handle. There is no need anymore to define additional enqueue* functions in each state estimator. This solves both problems 2) and 3).
* In the various sensor implementations, we push the measurements to the stateEstimator, similar to the LH deck. Note that we need to do that in the sensor implementation directly, because the barometer is updated at a different frequency. This fixes issue 1).
* In `stabilizer.c`, we always call `sensorsAcquire(&sensorData, tick);` to ensure that current logging variables are updated. In the future, we should eliminate the sensorData struct entirely, but this would be the scope of a separate issue.

whoenig:
@ataffanel @krichardsson It would be nice if you can comment on this design, before I work on an actual implementation of it.

krichardsson:
Looking good! Pushing the IMU data the same was as other sensors would be nice and we could get rid of some nasty dependencies as well.

One property that is lost in this design is "prioritisation" between sensor types. Even though it is not utilised in a structured way today, it is possible to set the queue length for different sensors. If one sensor pushes lots of data (too much) into the queue today, it will be discarded while other sensor data still will be used as it is pushed to other queues. In the suggested design we might discard the "wrong" sensor data instead. One may argue that we should have a different mechanism to throttle sensor data anyway.

[...]

Issue #616, opened 2020-09-11
Barometer logging does not work with the kalman estimator


krichardsson:
Sensor reading is not handled the same way for all estimators. When using the kalman estimator the barometer data is not passed on to the logging variable and plotting in the client shows 0.

**How to reproduce:**
1. Switch to the kalman estimator using a parameter or a deck (LPS for instance)
2. Set up logging for baro.pressure in the client

**Expected result:**
The current air pressure, around 1000

**Actual result:**
0

whoenig:
This is related to #722.

whoenig:
This works now with the latest master (tested using CF 2.1 with flow deck installed).

PR #735, opened 2021-03-17
Unify state estimator sensor data queues and move them to estimator.c

whoenig:
* IMU data is now pushed to the estimator, rather than pulled for
  consistency.
* The many different estimator queues are now unified into one
  queue of a dynamic union type.
* The latest sensorData is still filled in stabilizer for health check
  and logging.

Addresses issues #722 and #616.

### DIFF
--- a/src/hal/interface/sensors_mpu9250_lps25h.h
+++ b/src/hal/interface/sensors_mpu9250_lps25h.h
@@ -7,7 +7,7 @@
  *
  * Crazyflie control firmware
  *
- * Copyright (C) 2018 Bitcraze AB
+ * Copyright (C) 2018-2021 Bitcraze AB
  *
  * This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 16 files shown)


jonasdn (approved):
Very nice cleanup!

+365 −556 :1st_place_medal:

krichardsson:
Would it be possible to push the IMU data from the stabilizer loop instead? It would avoid duplication of code in all sensor implementations.

whoenig:
The challenge is the barometer, which is not read at the same frequency as the Gyro and Accelerometer. Currently, this logic is in each sensor implementation and the frequency could technically vary between different implementations. So in order to move the logic to the stabilizer, we would need some way to detect if a value actually changed (e.g., a counter). I feel that code duplication here might be more code overall, but is more consistent w.r.t. to the rest of the code and easier to read.

PR #743, opened 2021-03-22
Fix regression that caused invalid data in uSD card log files.

whoenig:
This regression was introduced with 303f0736298d7ec7151ac23ebb8bb9599e019334,
where the timestamp was changed to 64bit (8 Bytes), but the dataSize
check remained for 4 bytes. Some of the ringBuffer_push calls can fail
in that case, causing invalid events in the buffer.

This fix corrects the dataSize computation from 4 bytes to sizeof(ticks)


jonasdn (approved):
The numbers were to magical

PR #745, opened 2021-03-25
Pull request for the TWR robust Kalman filter

Williamwenda:
This pull request adds the robust Kalman filter for UWB TWR measurements.
The robust Kalman filter weights between (1) the states predicted from IMU measurements and (2) the noisy IMU measurements.
A good initial state should be sent to the CF in the beginning to state the estimation.

Please test the code and feel free to contact me if you have any questions.

Thanks


krichardsson:
Merged
Thanks!

krichardsson:
For some reason this pull request is not marked as merged even though it is

Williamwenda:
That's fine. My git version is a bit old which causes many weird problems. 
I have updated it recently. Hope further PRs will not have these issues.

PR #747, opened 2021-03-25
Update development docs part 2

jonasdn:


docs/development/systemtask.md
```diff
@@ -256,6 +253,15 @@ Add this to the Makefile, after the end of the `Modules` block:
...
 
+And to make sure that the DEBUG_PRINT from your task is shown in the console you can change:
+
+```{.make}
+DEBUG             ?= 0
+```
+In the Makefile to:
+```{.make}
+DEBUG             ?= 1
+```
```
whoenig, line 264: Those are actually shown even in release builds... Also, the "official" way would be to edit config.mk, not the Makefile. I suggest removing this part again.

whoenig (changes requested):

docs/development/systemtask.md
jonasdn: Gotcha!

krichardsson:
Looking good

PR #749, opened 2021-03-26
Fix issue where logGetVarId returns a groupId rather than a varId.

whoenig:
If the variable is called the same as the group, the current function
returns the group, rather than the actually variable. For example
"motion.motion". This makes it impossible to log the variable.


ataffanel (approved):
Good finding, this was nasty!

Issue #730, opened 2021-03-13
Deck Drivers: Add correct usage of used GPIO pins


whoenig:
A number of our current deck drivers have code like this:

```
  .usedGpio = 0,  // FIXME: set the used pins
```

For example: locodeck, lighthouse, etc. This should be replaced by the correct pin usage, e.g., `DECK_USING_IO_1`, as described in the documentation.

PR #750, opened 2021-03-26
Add usedGpio and usedPeriph to deck drivers

jonasdn:
Includes:

- locodeck: Add usedGpio and usedPeriph
- flowdec_v1v2: Add usedGpio and usedPeriph
- multiranger: Add usedGpio
- aideck: Add usedGpio and usedPeriph
- zranger1 zranger2: Add usedGpio
- gtgps: Add usedPeriph
- aidecktest: Add usedGpip and usedPeriph
- lighthouse: Add usedGpio and usedPeriph

Fixes: #730


whoenig (commented):
I think we need a discussion here what the difference is between GPIO and peripherals. In particular, if we use pins for peripheral purposes, do we still list those as part of GPIO? One example is I2C: do we list a) SDA and SCL in usedGpio, b) DECK_USING_I2C in usedPeriph, or c) both? The current PR actually has a mix of all 3 behaviors. I personally prefer b), i.e., only list a pin in usedGpio if it is used for a non-peripheral purpose. Then, in the automatic compatibility check, we can trigger an error if multiple boards try to use UART (but not if multiple try to use I2C).

ataffanel:
> One example is I2C: do we list a) SDA and SCL in usedGpio, b) DECK_USING_I2C in usedPeriph, or c) both? 

I also think b) is the one we should go for. A good example of why is the SPI port: two decks can use the SPI port at the same time, if they where to declare the SPI GPIOs it would be a conflict since they would both use the same GPIO in push-pull mode. However, the chip select pin needs to be declared as a GPIO and checked since two deck cannot use the same chip select at the same part.

One thing we are currently missing in the check we do for decks is that we are not marking GPIO taken by different busses: we would not detect a conflict if one deck using SPI and another one declaring one of the SPI line as used GPIO.

I really like the idea of this PR, we should finally fill up these fields and then (in a later step) implement a proper runtime incompatibility check. think it should be modified so that pin declared in GPIO are only the pins that are accessed by the deck driver directly, not the one access by a bus driver that can be declared in .usedPeriph.

jonasdn:
> > One example is I2C: do we list a) SDA and SCL in usedGpio, b) DECK_USING_I2C in usedPeriph, or c) both?
> 
> I also think b) is the one we should go for. A good example of why is the SPI port: two decks can use the SPI port at the same time, if they where to declare the SPI GPIOs it would be a conflict since they would both use the same GPIO in push-pull mode. However, the chip select pin needs to be declared as a GPIO and checked since two deck cannot use the same chip select at the same part.
> 
> One thing we are currently missing in the check we do for decks is that we are not marking GPIO taken by different busses: we would not detect a conflict if one deck using SPI and another one declaring one of the SPI line as used GPIO.
> 
> I really like the idea of this PR, we should finally fill up these fields and then (in a later step) implement a proper runtime incompatibility check. think it should be modified so that pin declared in GPIO are only the pins that are accessed by the deck driver directly, not the one access by a bus driver that can be declared in .usedPeriph.

An issue would be that if one deck uses SPI and another deck uses an alternate pinmux and utilizes A3/A4 analog-in or more generic the PA6/PA7 pins for other stuff than SPI (that is legal?). How do we catch that without having the SPI pins in usedGpios?

jonasdn:
I have forced pushed an update to this PR that only sets usedGpio for pins used directly by driver, not by peripherals.

jonasdn:
Maybe we should alter how we define the `usedPeriph` macros, so that it would be easier to detect conflicts with pins used?

```
/* Used GPIO */
#define DECK_USING_PC11 (1<<0)
#define DECK_USING_PC10 (1<<1)
#define DECK_USING_PB7  (1<<2)
#define DECK_USING_PB6  (1<<3)
#define DECK_USING_IO_1 (1<<4)
#define DECK_USING_IO_2 (1<<5)
#define DECK_USING_IO_3 (1<<6)
#define DECK_USING_IO_4 (1<<7)
#define DECK_USING_PA2  (1<<8)
#define DECK_USING_PA3  (1<<9)
#define DECK_USING_PA5  (1<<10)
#define DECK_USING_PA6  (1<<11)
#define DECK_USING_PA7  (1<<12)

/* Used peripherals */
[...]
#define DECK_USING_SPI    (DECK_USING_PA5 | DECK_USING_PA6 | DECK_USING_PA7)
[...]
#define DECK_USING_I2C    (DECK_USING_PB7 | DECK_USING_PB6)
```

whoenig:
> Maybe we should alter how we define the `usedPeriph` macros, so that it would be easier to detect conflicts with pins used?

That's a very elegant idea and I take it that you answered your earlier question on how to detect a conflict yourself then. There is only one corner case I can think of: UART can only be used by at most one deck.

src/deck/drivers/src/lighthouse.c
```diff
@@ -89,7 +89,8 @@ static const DeckDriver lighthouse_deck = {
   .pid = 0x10,
   .name = "bcLighthouse4",
 
-  .usedGpio = 0,  // FIXME: set the used pins
+  .usedGpio = 0,
+  .usedPeriph = DECK_USING_UART1 | DECK_USING_I2C,
```
whoenig, line 93: I think LH is currently using only UART. I2C is a possibility in the future.

src/deck/drivers/src/usddeck.c
```diff
@@ -1013,8 +1013,8 @@ static const DeckDriver usd_deck = {
     .vid = 0xBC,
     .pid = 0x08,
     .name = "bcUSD",
-    .usedGpio = DECK_USING_MISO|DECK_USING_MOSI|DECK_USING_SCK|DECK_USING_IO_4,
-    .usedPeriph = DECK_USING_SPI,
+    .usedGpio = DECK_USING_IO_4,
+    .usedPeriph = DECK_USING_SPI | DECK_USING_SPI,
```
whoenig, line 1017: Not wrong, but a bit odd:-)

src/deck/drivers/src/locodeck.c
```diff
@@ -580,7 +580,13 @@ static const DeckDriver dwm1000_deck = {
...
   .name = "bcDWM1000",
 
-  .usedGpio = 0,  // FIXME: set the used pins
+#ifdef LOCODEC_USE_ALT_PINS
+  .usedGpio = DECK_USING_IO_1 | DECK_USING_IO_2 | DECK_USING_IO_3,
+#else
+   // (PC10/PC11 is UART1 TX/RX)
+  .usedGpio = DECK_USING_IO_1 | DECK_USING_PC10 | DECK_USING_PC11,
+#endif
+  .usedPeriph = DECK_USING_SPI,
```
whoenig, line 589: from what I remember, this also uses timers?

src/deck/interface/deck_core.h
```diff
@@ -89,7 +77,17 @@ typedef struct deck_driver {
   uint8_t pid;
   char *name;
 
-  /* Periphreal and Gpio used _dirrectly_ by the driver */
+  /*
+   * Periphreal and Gpio used _dirrectly_ by the driver.
```
whoenig, line 81: maybe install a spell checker? (and be prepared for the shock when you browse existing code).

whoenig (changes requested):

src/deck/interface/deck_core.h
jonasdn: Squiggly.

src/deck/drivers/src/usddeck.c
jonasdn: Looks legit :)

src/deck/drivers/src/locodeck.c
jonasdn: I cannot find any calls to TIM_functions ... you got any pointers on where to look?

src/deck/drivers/src/lighthouse.c
jonasdn: Ok.

jonasdn:
Pushed commits to resolve issues found by @whoenig

knmcguire:
Let's try to resolve this, as this has been lingering a bit to long. @whoenig are you okay with the changes? and @jonasdn is this PR still valid? 

There is a merge conflict that needs to be fixed first though...

whoenig:
Yes, this looks good to me!

knmcguire:
Thanks @whoenig ! The PR still says that there are requested changes but then I'll merge it :)

jonasdn:
> Yes, this looks good to me!

You are still _technically_ requesting changes to this PR though, which prohibits merging. Is it ok to merge?

whoenig (approved):

whoenig:
Better now:-)?

jonasdn:
> Better now:-)?

Much better, thank you : )

PR #751, opened 2021-03-29
Add Eventtriggers for kalman filter enqueue functions.

whoenig:
Things we should discuss:

1. Adding all information to payload for IMU, rather than relying on logging variables?
2. Inclusion of stddev?
3. compressed quaternion vs. uncompressed?


whoenig:
The discussion resulted in the following decisions:
1. We keep the payload as minimal as possible, to allow for maximum flexibility.
2. No, since that is part of the measurement model.
3. This is solved by 1: the user can choose if they want to log the compressed or uncompressed state.

This PR was tested by krichardsson in the flight lab, collecting TDOA data.

PR #752, opened 2021-04-08
Adding support for Renode emulation

X-N-C:
Added a workaround that enables DMA.
Unfortunately the DMA does not get enabled automatically when reading I2C messages. The workaround does not seem to affect the actual Crazyflie when run on the hardware.

Added debug messages to easier pinpoint where (if any) self tests fail or if the self tests pass.


src/drivers/src/i2c_drv.c
```diff
@@ -498,6 +498,7 @@ static void i2cdrvEventIsrHandler(I2cDrv* i2c)
       // Enable the Transfer Complete interrupt
       DMA_ITConfig(i2c->def->dmaRxStream, DMA_IT_TC | DMA_IT_TE, ENABLE);
       I2C_DMACmd(i2c->def->i2cPort, ENABLE); // Enable before ADDR clear
+      DMA_Cmd(i2c->def->dmaRxStream, ENABLE); // Workaround for emulation...
```
jonasdn, line 501: All information about this workaround will be lost to time, only myths and legends will remain. Could you possibly add a, somewhat, more detailed comment about why this was needed or what problems you saw without it?

In other words, something like
```
//
//
//
//
DMA_Cmd(i2c->def->dmaRxStream, ENABLE);
```
Rather than:
```
DMA_Cmd(i2c->def->dmaRxStream, ENABLE); //
```

jonasdn (changes requested):
Thanks!

jonasdn:
Perhaps to avoid the merge commit in this PR you could go:

`$ git rebase origin/master` and update it with `$ git push -f ... `

src/modules/src/system.c
```diff
@@ -201,23 +201,38 @@ void systemTask(void *arg)
...
   pass &= soundTest();
+  DEBUG_PRINT("sound %s\n", pass ? "[OK]" : "[FAIL]");
   pass &= memTest();
+  DEBUG_PRINT("mem %s\n", pass ? "[OK]" : "[FAIL]");
   pass &= watchdogNormalStartTest();
+  DEBUG_PRINT("watchdogNormalStart %s\n", pass ? "[OK]" : "[FAIL]");
   pass &= cfAssertNormalStartTest();
+  DEBUG_PRINT("cfAssertNormalStart %s\n", pass ? "[OK]" : "[FAIL]");
   pass &= peerLocalizationTest();
+  DEBUG_PRINT("peerLocalization %s\n", pass ? "[OK]" : "[FAIL]");
```
ataffanel, line 230: These are a lot of prints on the console. I have 2 worries about it:
 - It is a lot more data printed at startup which is going to accumulate in the console buffer. It should still be ok but the risk of adding too much is to overflow the bugger
 - More important, the fact that if one test fails, all the subsequent one will be shown as failed which can be very confusing to the use.

One proposal would be to instead only show the failed tests:
```
if (systemTest() == false) {
  pass = false;
  DEBUG_PRINT("system [FAIL]");
}
```

It is more code but it would alleviate both my concerns. Another idea would be to show the printout only in debug mode (this is what we do for the deck driver, we are much more verbose in debug mode).

ataffanel (changes requested):

ataffanel (approved):
Thanks for the changes, looks good!

nirge:
I think these changes should also be changed in your other repository @ https://github.com/bitcraze/renode/tree/crazyflie/src

PR #753, opened 2021-04-09
Enable development of deck drivers in Rust

jonasdn:
With these commits we enable someone to write a deck driver using Rust.

The result of the example driver:
```
#![no_std]
#![feature(extended_key_value_attributes)]

use rust_drivers::*;

#[no_mangle]
pub extern "C" fn rustExampleInit(_info: *mut DeckInfo) {
    console_print!(b"Hello from rustExampleDriver init\n\0".as_ptr() as *const cty::c_char);
}

#[no_mangle]
pub extern "C" fn rustExampleTest() -> bool {
    console_print!(b"Hello from rustExampleDriver test\n\0".as_ptr() as *const cty::c_char);
    true
}

#[no_mangle]
static rust_example: RustDeckDriver = RustDeckDriver {
    name: DriverName(b"rustExampleDriver\0".as_ptr() as *const cty::c_char),
    init: rustExampleInit,
    test: rustExampleTest,
    usedPeriph: 0,
    usedGpio: 0,
    requiredEstimator: 0,
    requiredLowInterferenceRadioMode: false,
    vid: 0xDE,
    pid: 0xAD,
    memoryDef: ConstDeckMemDef(core::ptr::null() as *const DeckMemDef_t),
};

deck_driver!(rust_example, RustExampleDriver);
```
And forcing rustExampleDriver using `DECK_FORCE` is:
hello

The rust project in `src/deck/core/rust` uses `bindgen` to generate bindings for the headers specified and the driver itself lives in `src/deck/drivers/rust_example`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,5 @@ current_platform.mk
 
 /docs/.jekyll-metadata
 docs/.jekyll-cache
+
+Cargo.lock
--- a/src/deck/core/rust/.cargo/config
+++ b/src/deck/core/rust/.cargo/config
@@ -1,0 +1,2 @@
+[build]
+target = "thumbv7em-none-eabihf"
--- a/src/deck/core/rust/Cargo.toml
+++ b/src/deck/core/rust/Cargo.toml
@@ -1,0 +1,13 @@
+[package]
+name = "rust_drivers"
+version = "0.1.0"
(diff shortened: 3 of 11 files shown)


knmcguire:
This one is a bit old, but would it perhaps be better to have an RUST app layer example instead of a deck driver?

PR #758, opened 2021-04-20
remove unused checkStops variable from stabilizer.c, make inToOutLatency static

ntamas:
This PR cleans up two minor issues in `stabilizer.c` that I have stumbled upon while reviewing changes from the previous version of the firmware; in particular:

* it removes the now-unused `checkStops` variable; it used to be exposed as a log variable before the battery test was introduced, but not any more, hence the variable is not needed. (It is written but never read).

* it makes `inToOutLatency` static so it is not visible from other compilation units


krichardsson:
Thanks!

PR #759, opened 2021-04-22
motors: Document battery compensated thrust

jonasdn:
Data for this comes from:
  https://wiki.bitcraze.io/misc:investigations:thrust

### DIFF
--- a/src/drivers/src/motors.c
+++ b/src/drivers/src/motors.c
@@ -85,6 +85,59 @@ static uint16_t motorsConv16ToBits(uint16_t bits)
   return ((bits) >> (16 - MOTORS_PWM_BITS) & ((1 << MOTORS_PWM_BITS) - 1));
 }
 
+// We have data that maps PWM to thrust at different supply voltage levels.
+// However, it is not the PWM that drives the motors but the voltage and
+// amps (= power). With the PWM it is possible to simulate different
+// voltage levels. The assumption is that the voltage used will be an
+// procentage of the supply voltage, we assume that 50% PWM will result in
+// 50% voltage.
+//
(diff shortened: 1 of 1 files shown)


src/drivers/src/motors.c
```diff
@@ -231,10 +231,53 @@ void motorsSetRatio(uint32_t id, uint16_t ithrust)
...
+    // => p = -0.00062390   0.08835522   0.06865956
+    //
+    // And to get the PWM as a percentage we would need to divide the
+    // Voltage needed with the Supply voltage.
+    //
     if (motorMap[id]->drvType == BRUSHED)
     {
       float thrust = ((float)ithrust / 65536.0f) * 60;
-      float volts = -0.0006239f * thrust * thrust + 0.088f * thrust;
+      float volts = -0.0006239f * thrust * thrust + 0.088f * thrust + 0.06866;
```
jonasdn, line 280: Was the constant left out for a reason? Is it negligible?

tobbeanton: Yes the constant was left out as we need zero thrust to output zero.

tobbeanton:
Nice clarifying comment. What about breaking it out to its own function as well?

jonasdn:
> Nice clarifying comment. What about breaking it out to its own function as well?

I tried that and force-pushed, what do you think?

krichardsson (commented):
Great to document this stuff and move missing info from the wiki! 

Would it be better to put it in the repo docs instead? This is also only valid for the "normal" Crazyflie, how do we handle the Bolt and BigQuad decks? I suppose we will add more info about those when we finalize them.

PR #764, opened 2021-04-27
Add tools/build/elf_sanity.py

jonasdn:
This script uses pyelftools to check the elf output file.

In this first version it will make sure that the added parameters are
not duplicates and does not exceed the max length of 25 characters.

This is hard to do in compile-time with macro magic.

### DIFF
--- a/tools/build/check_elf
+++ b/tools/build/check_elf
@@ -2,4 +2,4 @@
 set -e
 
 scriptDir=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
-python3 ${scriptDir}/elf_sanity.py ${scriptDir}/../../cf2.elf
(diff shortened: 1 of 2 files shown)


krichardsson:
Nice!
I think it should be added to the master build script ```tools/build/build``` to run it in CI and so on as well.

krichardsson:
We have long discussed some form of documentation of log and params, could an extension of this functionality possibly be a part of that? For instance generating lists of logs and params to be able to see what is available and what has been added/removed between releases?

jonasdn:
> We have long discussed some form of documentation of log and params, could an extension of this functionality possibly be a part of that? For instance generating lists of logs and params to be able to see what is available and what has been added/removed between releases?

This is possible, the what and how should be discussed.

ataffanel:
This is a long needed addition. Looks great!

PR #767, opened 2021-04-30
Enable an estimator to be added by out of tree build

knmcguire:
There is now a define called  EXTERNAL_ESTIMATOR which adds extra estimator functions that will be filled in by an out of tree build of another estimator. This is mostly for other people to tryout their estimator without worrying to constantly merging the latest firmware.

### DIFF
--- a/docs/_data/menu.yml
+++ b/docs/_data/menu.yml
@@ -29,3 +29,4 @@
     - {page_id: systemtask}
     - {page_id: memory_management}
     - {page_id: dfu}
+    - {page_id: oot}
--- a/docs/development/oot.md
+++ b/docs/development/oot.md
@@ -1,0 +1,33 @@
+---
+title: Out of tree build
+page_id: oot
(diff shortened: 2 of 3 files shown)


ataffanel (changes requested):
This looks good!

What about documenting somewhere that adding "CFLAGS += -DEXTERNAL_ESTIMATOR" to the makefile and implementing the right functions enables implementation of an external estimator.

knmcguire:
Yeah good point. I will make a document in development called out of tree build, that this can be a subtopic of as well.

knmcguire:
By the way, isn't it enough to have EXTERNAL_ESTIMATOR = 1 in the OOT make file?

knmcguire:
Changed the name to out of tree build instead of external (since people probably will confuse it with an estimator running on python on a computer)

ataffanel (approved):
Looks good.

PR #770, opened 2021-04-30
OOT estimator fixes

knmcguire:
There was actually quite a lot more that needed to be done to really get the out of tree build of the estimator to work... But maybe it can serve as an example that the Kalman filter could maybe be even be more separated?

I just need to have this verified first before merging


krichardsson (approved):
Looking good!

PR #777, opened 2021-05-11
Add log- and param core macros

jonasdn:
- log.h: Add LOG_ADD_CORE macro
- param.h: Add PARAM_ADD_CORE macro

### DIFF
--- a/src/deck/drivers/src/activeMarkerDeck.c
+++ b/src/deck/drivers/src/activeMarkerDeck.c
@@ -258,8 +258,8 @@ PARAM_ADD_CORE(PARAM_UINT8, right, &requestedId[3])
 /**
  * @brief Off(0), pwm(1), modulated(2) or qualisys(3)
  *
- * Mode       | value         |     comment
- * ___________|_______________|_______________________________
+ *    Mode    |    Value      |         Comment
+ * -----------|---------------|-------------------------------
  * OFF        |   0           |    Always off
  * PWM        |   1           |    Always on, PWM modulated
  * MODULATED  |   2           |    Switching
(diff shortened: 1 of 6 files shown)


src/deck/drivers/src/activeMarkerDeck.c
```diff
@@ -211,12 +211,80 @@ static const DeckDriver deck_info = {
...
+ * @brief Off(0), pwm(1), modulated(2) or qualisys(3)
+ *
+ * Mode       | value         |     comment
+ * ___________|_______________|_______________________________
+ * OFF        |   0           |    Always off
+ * PWM        |   1           |    Always on, PWM modulated
+ * MODULATED  |   2           |    Switching
+ * QUALISYS   |   3 (default) |    Qualisys Active Marker mode
+ *
+ * Off mode
```
jonasdn, line 268: missing ###

knmcguire:
looks good to me! So does this mean that the regular  PARAM_ADD is considered a development add now?

And if the TYPE contain a bit for CORE params and logs, I guess we need something on the CFLib side to be able to take care of this right?

jonasdn (commented):
> looks good to me! So does this mean that the regular PARAM_ADD is considered a development add now?
> 
> And if the TYPE contain a bit for CORE params and logs, I guess we need something on the CFLib side to be able to take care of this right?

Yes, that would be the idea! If we want to show in the client that something is core we could look at the type yes! And we could add something to the script that check the elf-file here in the firmware and make sure they are documented, later on.

Adding something in cflib to make it easy to check if a param or log variable is core seems like a nice addition.

knmcguire:
Ah yeah oke that sounds good.

I guess what we will need to check for now if the CFlib side won't break with this addition once we merge it

jonasdn:
Switch to PR here: https://github.com/bitcraze/crazyflie-firmware/pull/780

PR #778, opened 2021-05-17
Pr lps doc

Williamwenda:
This PR adds the document of LPS system (convention Kalman filter and robust Kalman filter). Also, this PR unifies the titles of robust kalman filter.  

Please check and let me know if there are any suggestions.


krichardsson:
Thanks!

PR #780, opened 2021-05-18
Add and document core parameters and log variables

jonasdn:


src/deck/drivers/src/usddeck.c
```diff
@@ -1047,9 +1047,19 @@ PARAM_ADD_CORE(PARAM_UINT8 | PARAM_RONLY, bcUSD, &isInit)
...
 PARAM_GROUP_STOP(deck)
 
+/**
+ * The micro SD card deck is used for on-board logging of data to a micro SD card.
+ */
 PARAM_GROUP_START(usd)
-PARAM_ADD(PARAM_UINT8 | PARAM_RONLY, canLog, &initSuccess)
-PARAM_ADD(PARAM_UINT8, logging, &enableLogging) /* use to start/stop logging*/
+/**
+ * @brief Non zero if logging is possible, 0 indicates there might be a problem with the logging configuration.
```
jonasdn, line 1055: nit: I have used "Nonzero", I am not sure what is correct of "Non zero", "Non-zero" or "Nonzero" but we should probably use the same everywhere.

src/modules/src/collision_avoidance.c
```diff
@@ -345,12 +345,50 @@ LOG_GROUP_START(colAv)
...
 PARAM_GROUP_START(colAv)
-  PARAM_ADD(PARAM_UINT8, enable, &collisionAvoidanceEnable)
 
-  PARAM_ADD(PARAM_FLOAT, ellipsoidX, &params.ellipsoidRadii.x)
-  PARAM_ADD(PARAM_FLOAT, ellipsoidY, &params.ellipsoidRadii.y)
-  PARAM_ADD(PARAM_FLOAT, ellipsoidZ, &params.ellipsoidRadii.z)
+  /**
+   * @brief Nonzero to enable collision avoidance
+   *
+   * Used to enable or disable the collion avoidance module.
```
tobbeanton, line 374: Small spelling mistake "collion"

src/modules/src/power_distribution_stock.c
```diff
@@ -162,7 +162,17 @@ PARAM_ADD_CORE(PARAM_UINT16, m4, &motorPowerSet.m4)
...
+ */
 PARAM_GROUP_START(powerDist)
+/**
+ * @brief Motor thrust to set at idle (default: 0)
+ *
+ * This is often needed for brushless motors as
+ * it takes time to start up the motor. Then a
+ * common value is between 3000 - 6000.
+ */
 PARAM_ADD(PARAM_UINT32, idleThrust, &idleThrust)
```
jonasdn, line 176: This should be `PARAM_ADD_CORE` ?

src/modules/src/stabilizer.c
```diff
@@ -313,9 +313,22 @@ void stabilizerSetEmergencyStopTimeout(int timeout)
...
+ */
 PARAM_ADD(PARAM_UINT8, estimator, &estimatorType)
+/**
+ * @brief Controller type Any(0), PID(1), Mellinger(2), INDI(3) (Default: 1)
+ */
 PARAM_ADD(PARAM_UINT8, controller, &controllerType)
+/**
+ * @brief If set to nonzero will turn of power
+ */
 PARAM_ADD(PARAM_UINT8, stop, &emergencyStop)
```
jonasdn, line 332: `PARAM_ADD_CORE` ?

src/deck/drivers/src/flowdeck_v1v2.c
```diff
@@ -264,10 +264,23 @@ LOG_ADD(LOG_UINT8, squal, &currentMotion.squal)
...
 
+/**
+ * Settings and parameters for handling of the flowdecks
+ * measurments
+ */
 PARAM_GROUP_START(motion)
+/**
+ * @brief Nonzero to turn on adaptive standard devivation estimation (default: 0)
+ */
 PARAM_ADD(PARAM_UINT8, disable, &useFlowDisabled)
```
jonasdn, line 275: It seems disputed if this should be core or not? In the sheet you mark it as core. Should be discussed?

src/modules/src/collision_avoidance.c
jonasdn: thanks, will fix!

src/modules/src/power_distribution_stock.c
knmcguire: good one, I'll change it

src/deck/drivers/src/flowdeck_v1v2.c
knmcguire: Ah it was supposed to be a core one. It's a bit iffy though, since you don't want regular users to mess with this, but for many that doing research in estimators this is a very handy parameter. Technically I would say it is core. 

I also saw that I accidently switched the docs so I fixed that too

src/modules/src/system.c
```diff
@@ -349,17 +349,48 @@ void vApplicationIdleHook( void )
...
+
 PARAM_GROUP_STOP(cpu)
 
 PARAM_GROUP_START(system)
-PARAM_ADD(PARAM_INT8 | PARAM_RONLY, selftestPassed, &selftestPassed)
+
+/**
+ * @brief All tests passed when booting
+ */
+PARAM_ADD_CORE(PARAM_INT8 | PARAM_RONLY, selftestPassed, &selftestPassed)
```
tobbeanton, line 387: selftestPassed parameter has before been possible to set to 1 to "force" a system start. Therefore I suggest it not being read only. Possibly also add to the description something like:
"All tests passed when booting. If 0 can be set to 1 to force a system start"

PR #782, opened 2021-05-19
updated variable for config.mk to make use of the ifdef, previous entry had no effect

Spordoz:


krichardsson:
Merged. Thanks!

knmcguire:
hi! I just added another fix. Now it actually was sending the result of the crossing beam to the estimator if lighthouse_as_groundtruth was enabled... which is the opposite of what we wanted :)

PR #783, opened 2021-05-19
update lighthouse handling variable declaration.

Spordoz:
LIGHTHOUSE_AS_GROUNDTRUTH needs to be defined to have an influence, previous entry had no effect


krichardsson:
Merged. Thanks!

PR #784, opened 2021-05-20
Mark log variables as core and add documentation

krichardsson:
Document log variables and select a set of variable to be considered core logs. Core logs will be part of the stable API.

PR #788, opened 2021-05-26
Rechecked INDI controller for mistakes/consistency

MaxKemmeren:
As I am working with the INDI controller for my Thesis I checked the code for mistakes, and also made it more consistent in both style and units. 

Kimberley is working on cleaning the code and asked for any changes, so here they are.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -332,6 +332,7 @@ CFLAGS += $(PROCESSOR) $(INCLUDES)
 
 
 CFLAGS += -Wall -Wmissing-braces -fno-strict-aliasing $(C_PROFILE) -std=gnu11
+# CFLAGS += -O0 -Wmissing-braces -fno-strict-aliasing $(C_PROFILE) -std=gnu11 #Use this compiler during debugger, as it has a different optimizer so you can better track variables
 # Compiler flags to generate dependency files:
 CFLAGS += -MD -MP -MF $(BIN)/dep/$(@).d -MQ $(@)
 #Permits to remove un-used functions and global variables from output file
--- a/src/modules/interface/controller_indi.h
+++ b/src/modules/interface/controller_indi.h
@@ -37,23 +37,35 @@
 
 #define ATTITUDE_UPDATE_DT    (float)(1.0f/ATTITUDE_RATE)
 
-// these parameters are used in the filtering of the angular acceleration
+// these parameters are used in the filtering of the angular acceleration [Hz]
 #define STABILIZATION_INDI_FILT_CUTOFF 8.0f
 
 // the yaw sometimes requires more filtering
 #define STABILIZATION_INDI_FILT_CUTOFF_R STABILIZATION_INDI_FILT_CUTOFF
 
-// these parameters are used in the filtering of the angular acceleration
-#define STABILIZATION_INDI_G1_P 0.0066146f
+// Control effectiveness coefficients values Volodscoi
+#define STABILIZATION_INDI_G1_P 0.0066146f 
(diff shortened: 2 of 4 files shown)


knmcguire:
Hi Max! I'll check this contribution today in our flight lab. In the mean time, could you also add those comments and  @brief's for the logs as you did with the params? Especially since there are more logs added to this PR than before and we just added documentation of logs in #784.

MaxKemmeren:
Yes no problem, will add those comments today or tomorrow.

knmcguire:
Great! Sounds like a plan

 In the mean time, I've tested it in a lab setting and it seems to have very much improved! However, it seems that the default setting of indi (only innerloop indi and PID outer) seems not to work properly. With the outerloop also set on INDI, it works good. Perhaps we should put the outerloop indi as default and just remove the PID option completely?

As a sidenote, I'm wondering you or others have compared the INDI controller  with the standard PID controller? We don't have one of those big wind fans ourselves, but it would be nice to be able showcase the strength of INDI.

MaxKemmeren:
Ah okay, have not tested the INDI inner and PID outer combination, so might be some issues with signs that might make it not fly. So we can either investigate the signs in the PID and see if they match the signs of the inner INDI loop or just remove the option. 

Furthermore there are some papers on INDI which mostly compare it to the PID controller, e.g.
-Cascaded Incremental Nonlinear Dynamic Inversion Control for MAV Disturbance Rejection, by Ewoud J.J. Smeur, Guido C.H.E. de Croon and Qiping Chu

Moreover there is this video of the INDI and PID controller with the big fan setup: https://www.youtube.com/watch?v=VddwU5CabS8

knmcguire:
Hi Max,

Thanks for the citation! I am already familiar with the work with the INDI on the bebop but what i was trying to say, if it has been compared on the crazyflie itself ;)

MaxKemmeren:
Until now I have not seen any comparison on the Crazyflie itself, as the paper about the implementation of the INDI on the Crazyflie was focused on the implementation and not the performance itself.

knmcguire:
Thanks for adding the docs!
Could also the PID outerloop be removed if it can not be fixed? Or at least that default option of the outerLoopActive to be true?

By the way I also noticed that there is currently no information about the INDI controller in the [the documentation
](https://github.com/bitcraze/crazyflie-firmware/blob/master/docs/functional-areas/sensor-to-control/controllers.md). It doesn't have to be elaborate but just for people to know what it is so they know it exists. This information can be part of this pull request too as the documentation exists in the firmware repo.

MaxKemmeren:
I changed the boolean to true so it is always activated. More to the end of my thesis work I can always look into the combination inner INDI and outer PID to see what signs might cause the problems.

For the other documentation, I will talk to Ewoud upcoming monday again and discuss it with him. As he is sort of Master INDI it might be better to let him write a small summary.

MaxKemmeren:
Hi Kimberley, for now nothing will be changed until the control effectiveness and gains are confirmed to be perfect. If those are confirmed to be working I will make a new pull request, those are not very large changes, but have an impact on the performance. 

For the documentation I talked to Ewoud and he will arrange that, I will see him tomorrow and can thus ask when he plans to write the documentation about the INDI. 

Greetings 
Max

knmcguire:
Alright, I'll just wait untill you guys talked. I just want to make sure that there is nothing to be added more to the PR before I merged it!

knmcguire:
I'll merge it for now so if you want to add anything new that is totally fine as a new PR!

PR #791, opened 2021-06-01
param: fix paramGetInt() and paramSetInt()

ntamas:
This PR fixes `paramGetInt()` and `paramSetInt()` (and related functiosn that call these). The implementation broke after the introduction of the "core parameter" flag in the parameter type.

I think that the old `paramSetInt()` routine was not able to manipulate read-only params because of the presence of the "read only" flag. I updated the implementation of `paramSetInt()` to keep it this way even after masking the param type field with `PARAM_BYTES_MASK` - I hope that is okay.

### DIFF
--- a/src/modules/src/param.c
+++ b/src/modules/src/param.c
@@ -612,7 +612,7 @@ int paramGetInt(paramVarId_t varid)
 
   ASSERT(PARAM_VARID_IS_VALID(varid));
 
-  switch(params[varid.ptr].type)
+  switch(params[varid.ptr].type & PARAM_BYTES_MASK)
   {
     case PARAM_UINT8:
       valuei = *(uint8_t *)params[varid.ptr].address;
@@ -687,7 +687,10 @@ void paramSetInt(paramVarId_t varid, int valuei)
   pk.data[2] = (varid.id >> 8) & 0xffu;
   pk.size=3;
 
-  switch(params[varid.ptr].type)
+  if (params[varid.ptr].type & PARAM_RONLY)
+    return;
+
+  switch(params[varid.ptr].type & PARAM_BYTES_MASK)
   {
     case PARAM_UINT8:
       *(uint8_t *)params[varid.ptr].address = (uint8_t) valuei;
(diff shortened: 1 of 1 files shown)


ntamas:
Hmm, okay, sorry, this is not enough; `PARAM_BYTES_MASK` does not cover the part of the param type byte that encodes whether the param is an int or a float, neither does it cover the signedness flag.

Would it be okay to add a `PARAM_TYPE_MASK` with a value of `0xF`? This should match the signedness, int/float and size bits in the param type byte but not the "core" or "readonly" flags.

ntamas:
I think this version works now, and I did not want to get into a more elaborate refactoring, but I think that in the mid-term there should be macros like `IS_PARAM_READ_ONLY()`, `IS_PARAM_CORE()`, `GET_PARAM_TYPE()` and so on to make the code more readable. I can work on that in another PR if needed.

jonasdn:
> I think this version works now, and I did not want to get into a more elaborate refactoring, but I think that in the mid-term there should be macros like `IS_PARAM_READ_ONLY()`, `IS_PARAM_CORE()`, `GET_PARAM_TYPE()` and so on to make the code more readable. I can work on that in another PR if needed.

Thank you Tamás! Thans sounds great!

jonasdn:
Thanks for catching this, we had missed this angle :(

The `logGetInt()` / `logGetFloat()` functions suffer from this as well, right?

ntamas:
> The `logGetInt()` / `logGetFloat()` functions suffer from this as well, right?

I haven't checked, but most likely they do suffer from this as well. We don't use `logGetInt()` and `logGetFloat()` in our firmware fork, but from a cursory look at the source code it seems like the same problems are present there.

ntamas:
> The `logGetInt()` / `logGetFloat()` functions suffer from this as well, right?

I haven't checked, but most likely they do suffer from this as well. We don't use `logGetInt()` and `logGetFloat()` in our firmware fork, but from a cursory look at the source code it seems like the same problems are present there.

PR #794, opened 2021-06-10
converting wiki documentation to the doc folder

knmcguire:
The [wiki](https://wiki.bitcraze.io/) still contains a quite some pages that need to be converted to the githubrepo.

- [ ] Software related information with decks
- [ ] Deck hacks?

### DIFF
--- a/docs/userguides/decks/buzzer-deck.md
+++ b/docs/userguides/decks/buzzer-deck.md
@@ -1,0 +1,26 @@
+---
+title: Buzzer Deck
+page_id: buzzer-deck
+---
+
+## Changing the sounds 
+
+
+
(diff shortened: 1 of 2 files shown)


jonasdn (commented):
nice!
Couldn't this just be converted or added to the docs for the param group and params?
Since param is the API entry point anyway?

knmcguire:
I know that this is indeed the eventual plan, but this way at least all is in the repo docs and not lingering around in the wiki. Especially since the apidocs are not out yet. We need to convert the existing deck pages anyway in the end

knmcguire:
There is not really anything to add to the decks on the firmware side so I'll merge this now!

Issue #798, opened 2021-06-20
usd unable to log


easy98:
Hi,

My usd is not able to log any logging variables due to the assert problem here(https://github.com/bitcraze/crazyflie-firmware/blob/b448553725634aa83d25b356a7de12ba61647377/src/deck/drivers/src/usddeck.c#L504). I use the latest firmware.

> SUMMARY
========

PARAMETERS
 * /crazyflieTypes/CF21SingleMarker/batteryVoltageWarning: 3.8
 * /crazyflieTypes/CF21SingleMarker/batteryVoltateCritical: 3.7
 * /crazyflieTypes/CF21SingleMarker/bigQuad: False
 * /crazyflieTypes/CF21SingleMarker/dynamicsConfiguration: 0
 * /crazyflieTypes/CF21SingleMarker/firmwareParams/ctrlMel/i_range_m_z: 1500
 * /crazyflieTypes/CF21SingleMarker/firmwareParams/ctrlMel/i_range_xy: 2.0
 * /crazyflieTypes/CF21SingleMarker/firmwareParams/ctrlMel/i_range_z: 0.4
 * /crazyflieTypes/CF21SingleMarker/firmwareParams/ctrlMel/kR_xy: 70000
 * /crazyflieTypes/CF21SingleMarker/firmwareParams/ctrlMel/kR_z: 60000
 * /crazyflieTypes/CF21SingleMarker/firmwareParams/ctrlMel/kd_omega_rp: 200
 * /crazyflieTypes/CF21SingleMarker/firmwareParams/ctrlMel/kd_xy: 0.2
 * /crazyflieTypes/CF21SingleMarker/firmwareParams/ctrlMel/kd_z: 0.4
 * /crazyflieTypes/CF21SingleMarker/firmwareParams/ctrlMel/ki_m_z: 500
 * /crazyflieTypes/CF21SingleMarker/firmwareParams/ctrlMel/ki_xy: 0.05
 * /crazyflieTypes/CF21SingleMarker/firmwareParams/ctrlMel/ki_z: 0.05
 * /crazyflieTypes/CF21SingleMarker/firmwareParams/ctrlMel/kp_xy: 0.4
 * /crazyflieTypes/CF21SingleMarker/firmwareParams/ctrlMel/kp_z: 1.25
 * /crazyflieTypes/CF21SingleMarker/firmwareParams/ctrlMel/kw_xy: 20000
[...]

PR #799, opened 2021-06-21
usddeck: Add eventriggerType_fp16

jonasdn:
Fixes: #798

### DIFF
--- a/src/deck/drivers/src/usddeck.c
+++ b/src/deck/drivers/src/usddeck.c
@@ -885,6 +885,9 @@ static void usdWriteTask(void* prm)
               case eventtriggerType_float:
                 typeChar = 'f';
                 break;
+              case eventtrigerType_fp16:
+                typeChar = 'e';
+                break;
               default:
                 ASSERT(false);
               }
--- a/src/modules/interface/eventtrigger.h
+++ b/src/modules/interface/eventtrigger.h
@@ -40,6 +40,7 @@ enum eventtriggerType_e {
     eventtriggerType_int16  = 5,
     eventtriggerType_int32  = 6,
     eventtriggerType_float  = 7,
+    eventtrigerType_fp16    = 8,
 };
 typedef float float_t;
 


whoenig (approved):
Do we actually have variables of this type? Your code changes look good, but I am worried about proper support on the host side. In particular, does struct.decode do the right thing and is the firmware actually using the IEEE 754 binary16 standard? Supporting this datatype on a C++ host might also be challenging, as this is none of the standard types in C(++).

jonasdn:
> Do we actually have variables of this type? Your code changes look good, but I am worried about proper support on the host side. In particular, does struct.decode do the right thing and is the firmware actually using the IEEE 754 binary16 standard? Supporting this datatype on a C++ host might also be challenging, as this is none of the standard types in C(++).

Good questions!

Python supports this since I think 3.6: https://docs.python.org/3/library/struct.html#format-characters, and I believe it should work. And according to @ataffanel this should work (https://forum.bitcraze.io/viewtopic.php?p=22023&sid=311d5a4e46d027554d6e04312f573ebd#p22023).

I think both clang and gcc has extensions to support this. But! Your objections are fair. Perhaps instead of this PR we make it not assert and instead warn in some way?

whoenig:
My (bold) suggestion would have been to just remove FP16 entirely from the firmware - less to test & maintain... I don't remember seeing it being used inside the firmware and I've also not really seen the feature being used to log data with a different type (e.g., logging a float32 variable as FP16 to fit more inside a single log block.)

ataffanel:
There is currently no variable that uses FP16 in the firmware memory, this is only converted to be transferred in the log packets.

My intention was to implement IEEE 754 binary16 when I implemented FP16 in the Crazyflie. From my tests, the python implementation works fine with it. I do not remember being too careful about rounding and subnormalized representation is not handled, but for the purpose of compressing floating point representation it works well. For reference, the implementation is there: https://github.com/bitcraze/crazyflie-firmware/blob/a70c357a7c03b84a0caebe6d31a5fb0ccc40ff2b/src/utils/src/num.c#L48-L62.

Now, when it comes to the SD-card deck, it may be that fp16 is not that interesting since we have a lot more bandwidth. Though if it can be handled rather than failing it would be more coherent and less surprising if anyone tries to use it.

GCC support for fp16 was added right after I made my own implementation, it might also be an idea to use it in the Crazyflie so that we can be a bit more confident with interoperability.

ataffanel:
> My (bold) suggestion would have been to just remove FP16 entirely from the firmware - less to test & maintain... I don't remember seeing it being used inside the firmware and I've also not really seen the feature being used to log data with a different type (e.g., logging a float32 variable as FP16 to fit more inside a single log block.)

It is implemented in the Crazyflie lib now. I was implemented to make this packet possible using a single log block:
https://github.com/bitcraze/crazyflie-lib-python/blob/master/examples/basiclog.py#L75-L84

I do think that it is a useful functionality when it comes to the radio since we have such a limited space in each log packet.

whoenig:
For this particular case, I think it makes more sense to use `pm.vbatMV` (https://github.com/bitcraze/crazyflie-firmware/blob/6b04364847c5a2d6cbcb695395152e655c11b1fd/src/hal/src/pm_stm32f4.c#L421).

But going back to the topic at hand: if fp16 is to remain supported, then this PR looks good.

PR #800, opened 2021-06-22
log: Use TYPE_MASK to avoid problems with LOG_CORE

jonasdn:
Fixes: #798

### DIFF
--- a/src/modules/src/log.c
+++ b/src/modules/src/log.c
@@ -70,7 +70,7 @@ static const uint8_t typeLength[] = {
   [LOG_FP16]   = 2,
 };
 
-#define TYPE_MASK (0x0f)
+#define LOG_TYPE_MASK (0x0f)
 
 typedef enum {
   acqType_memory = 0,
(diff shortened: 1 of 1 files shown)


whoenig (approved):
Minor suggestion: move logGetType to the header file, mark it as inline, and then use the function throughout, rather than having to rely on not forgetting the TYPE_MASK. Perhaps TYPE_MASK should also be renamed to LOG_TYPE_MASK.

jonasdn:
> Minor suggestion: move logGetType to the header file, mark it as inline, and then use the function throughout, rather than having to rely on not forgetting the TYPE_MASK. Perhaps TYPE_MASK should also be renamed to LOG_TYPE_MASK.

Thanks!

The problem with moving `logGetType` to the header is that it references the global struct `logs` which point directly to the memory area of log structures. I am not sure we want that in a header file.

I pushed a commit that makes more use of `logGetType` and marks it inline, but does not move it to the header.

whoenig (approved):
Nice!

PR #802, opened 2021-06-23
Altitude hold baro

Spordoz:
Improved Altitude hold assisted mode using only the barometer and the complementary filter.
Using more of the barometer data in the estimation which results in a noisier estimation but better response for the PID.
Moved the D-filter from only the D component of the PID to the entire output. This is done to counteract the noise generated by the increased barometer influence which causes the P to also be noisy. Both the filter and PID are reset on initiating the assisted mode. Tweaked PID vallues and D-filter cutoff frequency for VZ to increase performance.


Spordoz:
Still need to remove binary files

PR #803, opened 2021-06-23
Barometer based altitude hold

Spordoz:
Improved Altitude hold assisted mode using only the barometer and the complementary filter.
Using more of the barometer data in the estimation which results in a noisier estimation but better response for the PID.
Moved the D-filter from only the D component of the PID to the entire output. This is done to counteract the noise generated by the increased barometer influence which causes the P to also be noisy. Both the filter and PID are reset on initiating the assisted mode. Tweaked PID vallues and D-filter cutoff frequency for VZ to increase performance.

These PID changes do have an influence on the unassisted thrust output as well, might need to put them behind a define flag.

### DIFF
--- a/src/modules/src/pid.c
+++ b/src/modules/src/pid.c
@@ -25,6 +25,10 @@
  * pid.c - implementation of the PID regulator
  */
 
+#ifdef IMPROVED_BARO_Z_HOLD
+#define PID_FILTER_ALL
+#endif
+
 #include "pid.h"
 #include "num.h"
 #include <math.h>
@@ -66,7 +70,7 @@ float pidUpdate(PidObject* pid, const float measured, const bool updateError)
     output += pid->outP;
 
     float deriv = (pid->error - pid->prevError) / pid->dt;
-    #ifdef IMPROVED_BARO_Z_HOLD
+    #ifdef PID_FILTER_ALL
       pid->deriv = deriv;
     #else
       if (pid->enableDFilter){
@@ -93,7 +97,7 @@ float pidUpdate(PidObject* pid, const float measured, const bool updateError)
     pid->outI = pid->ki * pid->integ;
     output += pid->outI;
     
-    #ifdef IMPROVED_BARO_Z_HOLD
+    #ifdef PID_FILTER_ALL
       //filter complete output instead of only D component to compensate for increased noise from increased barometer influence
       if (pid->enableDFilter)
       {
--- a/src/modules/src/pid.c
+++ b/src/modules/src/pid.c
@@ -66,13 +66,15 @@ float pidUpdate(PidObject* pid, const float measured, const bool updateError)
     output += pid->outP;
 
     float deriv = (pid->error - pid->prevError) / pid->dt;
-    /*if (pid->enableDFilter)
-    {
-      pid->deriv = lpf2pApply(&pid->dFilter, deriv);
-    } else {
+    #ifdef IMPROVED_BARO_Z_HOLD
       pid->deriv = deriv;
-    }*/
-    pid->deriv = deriv;
+    #else
+      if (pid->enableDFilter){
+        pid->deriv = lpf2pApply(&pid->dFilter, deriv);
+      } else {
+        pid->deriv = deriv;
+      }
+    #endif
     if (isnan(pid->deriv)) {
       pid->deriv = 0;
     }
@@ -90,18 +92,22 @@ float pidUpdate(PidObject* pid, const float measured, const bool updateError)
 
     pid->outI = pid->ki * pid->integ;
     output += pid->outI;
-
-    //filter complete output instead of only D component to compensate for increased noise from increased barometer influence
-    if (pid->enableDFilter)
-    {
-      output = lpf2pApply(&pid->dFilter, output);
-    }
-    else {
-      output = output;
-    }
-    if (isnan(output)) {
-      output = 0;
-    }
+    
+    #ifdef IMPROVED_BARO_Z_HOLD
+      //filter complete output instead of only D component to compensate for increased noise from increased barometer influence
+      if (pid->enableDFilter)
+      {
+        output = lpf2pApply(&pid->dFilter, output);
+      }
+      else {
+        output = output;
+      }
+      if (isnan(output)) {
+        output = 0;
+      }
+     #endif
+      
+    
 
     // Constrain the total PID output (unless the outputLimit is zero)
     if(pid->outputLimit != 0)
--- a/src/modules/src/pid.c
+++ b/src/modules/src/pid.c
@@ -66,15 +66,17 @@ float pidUpdate(PidObject* pid, const float measured, const bool updateError)
     output += pid->outP;
 
     float deriv = (pid->error - pid->prevError) / pid->dt;
-    if (pid->enableDFilter)
+    /*if (pid->enableDFilter)
     {
       pid->deriv = lpf2pApply(&pid->dFilter, deriv);
     } else {
       pid->deriv = deriv;
-    }
+    }*/
+    pid->deriv = deriv;
     if (isnan(pid->deriv)) {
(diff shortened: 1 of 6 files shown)


src/modules/src/pid.c
```diff
@@ -66,15 +66,17 @@ float pidUpdate(PidObject* pid, const float measured, const bool updateError)
     output += pid->outP;
 
     float deriv = (pid->error - pid->prevError) / pid->dt;
-    if (pid->enableDFilter)
+    /*if (pid->enableDFilter)
```
tobbeanton, line 69: Should this code be removed. Commented code is always hard to understand.

src/modules/src/position_controller_pid.c
```diff
@@ -98,9 +100,9 @@ static struct this_s this = {
 
   .pidVZ = {
     .init = {
-      .kp = 25,
-      .ki = 15,
-      .kd = 0,
+      .kp = 3.0f,
```
tobbeanton, line 103: New tuning of the VZ PID. This needs to be tested in other positioning systems as well as it will effect those. I suggest testing the new values in Loco and Lighthouse systems

tobbeanton (commented):
Mainly more testing of the VZ tuning is needed.

src/modules/src/position_controller_pid.c
Spordoz: Using Lighthouse: Very slowly descends on no input while using the assisted mode (less than a centimeter per second), and slightly overshoots on release of height control input (keeps going for about 5-10 cm). Have not been able to test it using the loco system

knmcguire:
@Spordoz if you are unsure about the tuning gains of PID Z, please put them within a compile flag: like this

```
#ifdef IMPROVED_BARO_Z_HOLD
   //your pid gains
#else
   //current PID gains 
#endif
```

How it is currently performing in lighthouse does not seem like this is desirable as you are describing, so we would not like to have this gains as standard, or find a way to change the gains nicely if the baro is used.

knmcguire:
Hi! Just a reminder on this pull request. Could you please implement the suggested changes as we noted before?

src/modules/src/pid.c
Spordoz: has been uncommented and added back in with the addition of ifdefs

Spordoz:
Suggested changes have been added, please check them to make sure they are added correctly, but it should be good

src/modules/src/pid.c
```diff
@@ -90,18 +92,22 @@ float pidUpdate(PidObject* pid, const float measured, const bool updateError)
...
-      output = lpf2pApply(&pid->dFilter, output);
-    }
-    else {
-      output = output;
-    }
-    if (isnan(output)) {
-      output = 0;
-    }
+    
+    #ifdef IMPROVED_BARO_Z_HOLD
```
matejkarasek, line 96: Filtering the entire PID output can be useful also in other situation... 
Maybe better to use a different define name?

e.g. PID_FILTER_ALL

and then
#ifdef IMPROVED_BARO_Z_HOLD
#define PID_FILTER_ALL
#endif

src/modules/src/position_estimator_altitude.c
```diff
@@ -100,8 +104,12 @@ static void positionEstimateInternal(state_t* estimate, const baro_t* baro, cons
       filteredZ = (state->estAlphaAsl       ) * state->estimatedZ +
                   (1.0f - state->estAlphaAsl) * baro->asl;
     }
-    // Use asl as base and add velocity changes.
-    state->estimatedZ = filteredZ + (state->velocityFactor * state->velocityZ * dt);
+    #ifdef IMPROVED_BARO_Z_HOLD
+      state->estimatedZ = filteredZ;
+    #else
```
matejkarasek, line 109: Maybe remove the original formula entirely? Or is it used in other situations?
The velocity changes do not seem to help with the estimation...

src/modules/src/position_estimator_altitude.c
```diff
@@ -51,7 +51,11 @@ static struct selfState_s state = {
   .estimatedZ = 0.0f,
   .velocityZ = 0.0f,
   .estAlphaZrange = 0.90f,
-  .estAlphaAsl = 0.90f,
+  #ifdef IMPROVED_BARO_Z_HOLD
```
matejkarasek, line 54: I do not see a reason for keeping the old value, it seems to be only used for baro-based estimation

src/modules/src/position_estimator_altitude.c
```diff
@@ -100,8 +104,12 @@ static void positionEstimateInternal(state_t* estimate, const baro_t* baro, cons
       filteredZ = (state->estAlphaAsl       ) * state->estimatedZ +
                   (1.0f - state->estAlphaAsl) * baro->asl;
     }
-    // Use asl as base and add velocity changes.
-    state->estimatedZ = filteredZ + (state->velocityFactor * state->velocityZ * dt);
+    #ifdef IMPROVED_BARO_Z_HOLD
+      state->estimatedZ = filteredZ;
+    #else
+      // Use asl as base and add velocity changes.
+      state->estimatedZ = filteredZ + (state->velocityFactor * state->velocityZ * dt);
```
matejkarasek, line 111: Maybe remove the original formula entirely? Or is it used in other situations?
The velocity changes do not seem to help with the estimation...

src/modules/src/position_estimator_altitude.c
```diff
@@ -51,7 +51,11 @@ static struct selfState_s state = {
   .estimatedZ = 0.0f,
   .velocityZ = 0.0f,
   .estAlphaZrange = 0.90f,
-  .estAlphaAsl = 0.997f,
+  #ifdef IMPROVED_BARO_Z_HOLD
+    .estAlphaAsl = 0.90f,
+  #else
+    .estAlphaAsl = 0.997f,
```
matejkarasek, line 57: I do not see a reason for keeping the old value, it seems to be only used for baro-based estimation

knmcguire: Yeah you are right, let's up it to the new value. I don't think it will matter that much for the rest of the estimation as for the zranger

src/modules/src/pid.c
knmcguire: Yeah true, I agree

knmcguire:
Also You should add the new ifdefs to [config.mk](https://github.com/bitcraze/crazyflie-firmware/blob/master/tools/make/config.mk.example), so people knows it exists!

src/modules/src/position_estimator_altitude.c
Spordoz: In that case i will leave this as is

src/modules/src/pid.c
Spordoz: has been done

knmcguire:
Hi! It's double checked, flight tested, merged, and I've added the compile flag to the config.mk.example list too. It seemed that in our office it flew pretty good but perhaps it needs some more tuning of the Z axis pid controller, but definitely much better than it was before.

Thanks!

ntamas:
@knmcguire I'm not 100% sure whether it's a bug or not, but it seems like `positionControllerResetAllfilters()` resets the cutoff frequency of the Z velocity lowpass filter to `ZVELOCITY_LPF_CUTOFF_FREQ` even if `IMPROVED_BARO_Z_HOLD` is not defined. It's suspicious because `positionControllerInit()` uses `POSITION_LPF_CUTOFF_FREQ`. Is this intentional?

matejkarasek:
Well spotted @ntamas, the pid reset is wrong...
There should be an #ifdef as well

@knmcguire Can make a new PR to fix this.

knmcguire:
Ah yeah that's right! missed that. Thanks @ntamas !

Yeah @matejkarasek sounds good. I'll keep an eye on it and accept it quickly

PR #807, opened 2021-06-25
Added optional transmit channel for raw lighthouse data

jdavidberger:
This change allows one to set a parameter and start streaming off the raw light data. This is slightly different and in a more raw state than the port and channel which transmits angle data. It is also packed in such a way to allow two full frames of light data per packet, so in theory a single lighthouse will consume ~50p/s. 

The transmit is non blocking, so if the bandwidth is there it works and if it isn't, it just moves on to the next without endangering the stability of on-board tracking. The computational overhead should also pretty insignificant at these data rates. 

I have a branch in [libsurvive](https://github.com/cntools/libsurvive/blob/feature/crazyflie_driver/src/driver_crazyflie.cc) that has the receiver logic for reference.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -263,7 +263,7 @@ PROJ_OBJ += filter.o cpuid.o cfassert.o  eprintf.o crc32.o num.o debug.o
 PROJ_OBJ += version.o FreeRTOS-openocd.o
 PROJ_OBJ += configblockeeprom.o
 PROJ_OBJ += sleepus.o statsCnt.o rateSupervisor.o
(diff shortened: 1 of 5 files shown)


knmcguire:
There are some checks that are failing so you might want to take a look at that.

On the first sight, this seems to be fine! But I also want @ataffanel to look at this, who will be back next week.

ataffanel:
Hi, Sorry for the delayed review, this looks great I am merging it, thanks!

Do you by any change have a simple python script showing how to use this new functionality that we could put in the [python lib](https://github.com/bitcraze/crazyflie-lib-python/tree/master/examples/lighthouse) example folder?

jdavidberger:
Not right not but possibly later. I need to merge the libsurvive in; and then potentially update the pysurvive bindings -- currently they are autogenerated and I think I need to update how that all works.

PR #809, opened 2021-07-06
accelerometer log update

knmcguire:
tiniest PR ever? :D


src/modules/src/stabilizer.c
```diff
@@ -493,7 +493,7 @@ LOG_ADD(LOG_UINT32, intToOut, &inToOutLatency)
 LOG_GROUP_STOP(stabilizer)
 
 /**
- * Log group for accelerometer data
+ * Log group for accelerometer sensor measurement
```
jonasdn, line 496: nice! should we also include that it is aligned to gravity?

knmcguire: good point! done

jonasdn (approved):

PR #810, opened 2021-07-06
Add epsilon to divisions in kalman core

pablo-ng:
When all gyro measurements are equal to zero several division by zero are made which causes unexpected behaviour and an undefined state estimaton. A small number EPS=1e-6 is added to each quotient.

### DIFF
--- a/src/modules/src/kalman_core/kalman_core.c
+++ b/src/modules/src/kalman_core/kalman_core.c
@@ -121,6 +121,9 @@ static void assertStateNotNaN(const kalmanCoreData_t* this)
 #define MAX_COVARIANCE (100)
 #define MIN_COVARIANCE (1e-6f)
 
+// Small number epsilon, to prevent dividing by zero
+#define EPS (1e-6f)
+
 // Initial variances, uncertain of position, but know we're stationary and roughly flat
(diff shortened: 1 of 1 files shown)


knmcguire:
Hi! Thanks for the pull request. It is odd for it to happen that all gyros measurements are zero simultaneously, but by murphy's law... it could happen:) 

Did you experience it this problem yourself? How did you find out it was due to the gyro measurements

pablo-ng:
I was looking at this implementation for my own project. When I was testing the code without any inputs I got some unexpected states. But sure, in a real scenario it is very rare to happen :)

knmcguire:
Yeah good point. I checked it out and it all seem to work fine! 

The only problem that I might foresee, is that if the sensor is indeed sending anything out and it is constantly sending out zero (or is stuck at one value), there is no hard fault anymore and the state will keep on diverging, while the user does not know what is going on. Perhaps not immediately linked to the PR, but perhaps we should implement some kind of mechanism that checks if these essential sensors are failing in flight.

knmcguire:
I've made an issue to remind ourselves to put in this check. #815

So for now I'll merge your changes! Thanks again :)

PR #811, opened 2021-07-08
stabilizer: Update log documentation on IMU sensors

jonasdn:
This was provoced by a [comment](https://github.com/bitcraze/crazyflie-firmware/commit/4c2b1a57e8a40697efe9c28da43ebf85c9dcb38c#commitcomment-53198019) on a commit about documentation on the (acc) log group.

### DIFF
--- a/src/modules/src/stabilizer.c
+++ b/src/modules/src/stabilizer.c
@@ -493,7 +493,13 @@ LOG_ADD(LOG_UINT32, intToOut, &inToOutLatency)
 LOG_GROUP_STOP(stabilizer)
 
 /**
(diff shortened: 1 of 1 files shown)


knmcguire:
Perhaps we need to double check this, as we do some preprocessing ourselves before we send the sensor data out to this log variable... Maybe we need to have  a seperate logging group called raw sensor data, where we show this information? 

By the way, I noticed that the parser does not take second paragraphs for the log group definitions (like stateExtimateZ) I'll try to look at this today

knmcguire (commented):
Good! More doc is more better  :D

PR #812, opened 2021-07-08
Fix log and param doc parser

knmcguire:
everything with a NOTE: was ignore before by the parser, which has now been fixed..

and also there was one table that wasn' t rendering in markdown


jonasdn:
Did something go wrong here @knmcguire? The title does not match the commits

knmcguire:
oh woops!!!

PR #813, opened 2021-07-08
Fix auto doc for loggs and params 

knmcguire:
Second try!

everything with a NOTE: was ignore before by the parser, which has now been fixed..

and also there was one table that wasn' t rendering in markdown


jonasdn (approved):
Nice! Perhaps we could deploy the web when this has been merged

PR #816, opened 2021-07-13
allow modifying the name of the compiled binary by setting PROG in the environment

ntamas:
This PR adds support for overriding the name of the compiled firmware binary by setting `PROG` from the command line or from another Makefile; e.g., one can now do fancy stuff this:

```sh
PROG=crazyflie-firmware-$(git describe) make
```

This will build the firmware in a way that the name of the Git revision is included in the filename. This is useful for archival purposes. One other possible use-case is that when compiling a Crazyflie app, you can override the name of the firmware in the app makefile so you can easily identify which app was built in a particular binary.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -359,7 +359,7 @@ ifeq ($(LTO), 1)
 endif
 
 #Program name
-PROG = $(PLATFORM)
+PROG ?= $(PLATFORM)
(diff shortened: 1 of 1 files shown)


knmcguire:
Sounds good! I'll merge it. THanks!

Issue #540, opened 2020-01-28
Position stabilization with flow deck v2


EnricaSo:
I have a Crazyflie 2.1 flashed with  febc78015a9b (CLEAN) and equipped with a flow deck v2 and I am using the crazyflie-lib-python to send high-level commands (goto). I am facing the following issue.

I would like to **stabilize the drone height** with the only help of IMU, barometer, and optic flow, and **no z-ranger**. However, I would like to access the z-ranger measurements to know if there is any obstacle on the ground. Right now, I observe that whenever the drone detects an obstacle beneath itself it overshoots in position believing that its height from the ground has changed. However, it is not the case: the ground level is unchanged, but there is an obstacle. How could I change this behavior? If the stabilization in z without the z-ranger is unfeasible, how would you solve this problem?

Regards,
Enrica

knmcguire:
Hi!

So just to rephrase a bit: Currently you are looking at the state estimate of the position, and you expected the position to jump up whenever there is a obstacle underneath there, right? But that is not happening?

About your question about doing IMU only height/stabilization, it is a interesting feature but since it is not generic to our firmware (yet), I would advice that you post those questions on the forum.bitcraze.io. We would like to keep the issues in github to be about problems with the current firmware or wanted generic functionality if that is okay :)

EnricaSo:
Hello!

Thank you for your reply. What you just wrote is not what I meant, I am sorry for the miscommunication. 

The situation is the following: the ground is flat and there is an obstacle on it. The position of the drone does jump up when it flies over the obstacle, while I would like it to be constant. This leads me to think that the z-position estimate that I read is more the estimate of the height from the ground level, rather than the relative position to the starting point. I would like to measure the height of the obstacle with the z-ranger without overshooting in position. Is it possible? How would you do it? I can provide plots of the ground-truth position and the estimated one if needed. And yes, I can repost this question on the forum.

Thanks a lot for your help.

knmcguire:
Ahhh oke. I understand now. Actually maintaining a constant height while flying over obstacles with the zranger is indeed a wanted feature. I will change the title to make the issue more clear.

So this can be done in two ways:

- A step detection with the z-ranger (sudden changes will not be included for the position estimation)
- Integrating the barometer properly into the kalman filter 

It is not an quick solution for this, but I think it will be beneficial for other people as well.

krichardsson:
A quick comment on the barometer and the kalman filter: 
[...]

Issue #711, opened 2021-03-02
Unified estimator task loop


knmcguire:
Currently the extended Kalman filter is always initialized and taking up RAM, even if only the complementary filter is used. It is not really adding problems now, but if in the future more types of estimators are added, this might be a bit difficult to handle.

How about just starting one generic estimator task, where within we can switch between the Comp. filter, EKF or estimator number# ?

jonasdn:
Is this something we want to pursue? If so, do we have a plan of attack?

knmcguire:
To kind of give a background information: In the current state of the firmware, the kalman estimator is already initialized [here](https://github.com/bitcraze/crazyflie-firmware/blob/d7d9c3dfbe1dd13fe0376be5632281a74a0c968a/src/modules/src/system.c#L190) in system.c and the rest of the estimators [here in stabilizer.c ](https://github.com/bitcraze/crazyflie-firmware/blob/d7d9c3dfbe1dd13fe0376be5632281a74a0c968a/src/modules/src/stabilizer.c#L171). The kalman filter is currently now setup as a seperate task, but I think we should also do that with the complementary and the OOT one.

Technically, the kalman filter is not taking up RAM if the complementary filter is selected, as the loop is physically stopped [here](https://github.com/bitcraze/crazyflie-firmware/blob/d7d9c3dfbe1dd13fe0376be5632281a74a0c968a/src/modules/src/estimator_kalman.c#L207) if the complementary filter is selected, so that was a wrong assumption from me, but... the general structure could be improved, starting with putting the complementary filter also in a task for starters

knmcguire:
This would also be valid if we want to make python bindings of these api's like mentioned in #602

krichardsson:
There is some documentation in the oot build section https://www.bitcraze.io/documentation/repository/crazyflie-firmware/master/development/oot/

PR #823, opened 2021-08-10
Kalman filter modification for constant altitude flight 

KristofferSkare:
## Overview
This is a proposed solution for flying with constant altitude over obstacles as brought up by [this github issue](https://github.com/bitcraze/crazyflie-firmware/issues/540). The solution is based on "step detection with the zranger" as suggested in the issue, but this solution uses a bit more for example by using the upward range measurement form the multiranger for better performance. Below is a video to show how well it works when flying off a table.

ezgif com-gif-maker(6)

I worked on this as part of my summer internship at [SINTEF](https://www.sintef.no/en/digital/departments-new/applied-mathematics/robotics-and-control/) as part of the [Ingenious project](https://www.sintef.no/en/projects/2019/ingenious/). 

## How it works
The state in the Kalman filter was extended with two new states. These are f ("floor height") and r ("roof height"). f is defined as the height of the object under the Crazyflie compared to the height where the z state is defined as 0. 
r is defined as the height of the object over the Crazyflie compared to the same reference height. These states are used together with downward and upward range measurements in the Kalman filter for better estimation of z. The states are illustrated below.
z_estimation_state_drawing

Previously the downward range measurement (l_D in illustration) was used as a direct measurement of z, but with the new definitions it now is a measurement of (z - f). Similarly the upward range measurement (l_U in illustration)is a measurement of (r - z). Both of the measurements also have an effect from the angle to the floor (theta) which is used in the same way as previously, that is by dividing by R[2][2] (rotation matrix). 

However, using this directly in the Kalman filter is not enough since one is now using two measurments to estimate three states. This is handled by having the new states f and r generally not changing, meaning I assume the floor and roof is flat most of the time. However, there is a detector for both the upward and downward range measurements for detecting fast abrupt changes in the range measurements. When a fast change is detected it is assumed it is the corresponding state (f or r) that is supposed to change. Then either f or r is set to a "best first guess" and given a high variance in the covariance matrix for the Kalman filter to fine tune the estimate of the state that was changed. 

In addition to these main changes the optical flow sensor also needed a small change where the distance to the floor is no longer z, but (z - f). Here there is also a difference from the "correct" way to use a Kalman filter where "h_z" and "h_f" are set to zero. This is because setting them to their "correct" values caused the states to drift and converge to completely wrong values. 

This new feature can be turned on an off using the parameter "kalman.useFAndR". Setting it to true (or nonzero) causes the EKF to use this feature. When setting it to false the firmware SHOULD work as it did without the change, but I can't be completely sure that it does not cause any difference. 

## Problems

The optical flow measurement implementation in the Kalman filter assumes the Crazyflie is flying over a flat surface. Now that it can fly over obstacles which are not a flat surface, but can have very large differences in distance to the sensor it sometimes messes up the estimates in x and y. This manifests as the x and y estimates stopping or "going the wrong way" while the Crazyflie is moving over an edge. If one has a set point at the other side of the edge it will cause the Crazyflie to speed up and go too far past the set point. I found that this effect happens less or completely disappears when the distance to the obstacles are "large enough". This is because when having a larger distance to the obstacle the relative difference between the distances to the sensor between the different surfaces gets smaller, which makes the errors smaller. I found that this effect tends to disappear when flying at an altitude that is double of the height of the obstacle, e.g if flying over a box that is 0.5m tall one should fly at an altitude of at least 1.0m to the floor.

### DIFF
--- a/src/deck/drivers/src/multiranger.c
+++ b/src/deck/drivers/src/multiranger.c
@@ -43,6 +43,8 @@
 
 #include <stdlib.h>
 
+#include "cf_math.h"
+
 static bool isInit = false;
 static bool isTested = false;
 static bool isPassed = false;
@@ -53,11 +55,20 @@ static bool isPassed = false;
 #define MR_PIN_LEFT   PCA95X4_P6
 #define MR_PIN_RIGHT  PCA95X4_P2
 
+#define RANGE_UP_OUTLIER_LIMIT 5000 // the measured range is in [mm]
+
 NO_DMA_CCM_SAFE_ZERO_INIT static VL53L1_Dev_t devFront;
 NO_DMA_CCM_SAFE_ZERO_INIT static VL53L1_Dev_t devBack;
 NO_DMA_CCM_SAFE_ZERO_INIT static VL53L1_Dev_t devUp;
 NO_DMA_CCM_SAFE_ZERO_INIT static VL53L1_Dev_t devLeft;
 NO_DMA_CCM_SAFE_ZERO_INIT static VL53L1_Dev_t devRight;
+
+// Measurement noise model
+static const float expPointA = 2.5f;
+static const float expStdA = 0.0025f; // STD at elevation expPointA [m]
+static const float expPointB = 4.0f;
+static const float expStdB = 0.2f;    // STD at elevation expPointB [m]
+static float expCoeff;
 
 static bool mrInitSensor(VL53L1_Dev_t *pdev, uint32_t pca95pin, char *name)
 {
@@ -78,6 +89,10 @@ static bool mrInitSensor(VL53L1_Dev_t *pdev, uint32_t pca95pin, char *name)
       DEBUG_PRINT("Init %s sensor [FAIL]\n", name);
       status = false;
   }
+
+  // pre-compute constant in the measurement noise model for kalman
+  expCoeff = logf(expStdB / expStdA) / (expPointB - expPointA);
+
 
   return status;
 }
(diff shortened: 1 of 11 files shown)


krichardsson:
Nice!

We'll check it out

knmcguire:
It's a really cool project you have done here! 

We do have a couple of concerns with the contribution though, as there are a lot of changes to the kalman filter, including the state. Have you been able to measure how much this will increase the computational load of the kalman filter?

Moreover, we are also in the middle of trying to reduce the complexity of the code and as this is for a quite specific application, this would perhaps not really fit. I'm wondering if it might be rather suitable for perhaps an out of tree estimator build, which is initialized through the app layer? 

(PS: But it's still quite a cool achievement, so if you are interested in writing a blogpost about this, contact us!)

dominiknatter:
Hi guys,
I work in the group Kristoffer has joined during summer. By now he has left us in order to continue his studies. He might chime in in the discussion from time to time, but for now I'll try to push this forward.

Unfortunately, no tests concerning the computational load were conducted. Even if tests were conducted, I'd say very likely there is room for improvement as the development did not focus on being time efficient (e.g., if I'm correct `useFAndR` should only suppress the usage of the new states, but not their calculation).

About the concrete implementation:
First of all I can understand your desire to reduce code complexity and indeed the proposed solution is certainly not useful for everyone.
I am not fully sure how your questioning suggestion would be implemented in detail, as I have not worked with [your out of tree (oot) development](https://www.bitcraze.io/documentation/repository/crazyflie-firmware/master/development/oot/) so far. The current pull request changes directly the estimator module. However, in [one of your oot examples](https://github.com/ataffanel/crazyflie-push-demo) an app layer application in an src directory interfaces the original firmware as a submodule. Using that approach, we'd need to interface the estimator in the original firmware from within the new src files somehow, right?
A further question to deepen my understanding: that approach would mean that the oot build will be developed with a specific version of the firmware. So if the firmware changes too much over time, and the oot build shall use the newest firmware, the oot build itself might demand changes?

As you guys are working more closely with the firmware, it is likely you who can choose the most suitable approach here. Therefore we are happy for your input :)

knmcguire:
It's a good question of how to better implement this. Currently the code has compatibility to do an [out of three estimator ](https://www.bitcraze.io/documentation/repository/crazyflie-firmware/master/)as I told before, but we might be wanting look at unifying the estimator structure in general which is still up for discussion (see #711 ). So it's still a complicated situation actually. 

You can already try to get it to work as out of tree but maybe it's smarter to wait a little until we have concluded what the final structure of the estimator framework should be

dominiknatter:
Hi @knmcguire,
thanks for the update. I totally forgot about closing this. The blog post is still in my mind though and will be tackled soon. Sorry for both delays.
Good to see the out of tree estimator structure being used! It will certainly be helpful if we convert @KristofferSkare's solution into that format ourselves.

knmcguire:
no problem at all! just give us an update whenever you are ready to write the blogpost or want help with converting your kalman filter

KenjiNM:
Hi. I want to test this code for constant altitude flight. 
Is this pull request currently merged into master?

Where can I get the code?

Chakad-lab (approved):

PR #827, opened 2021-08-17
Do not mark conflict when bus periphs are shared

jonasdn:
- flowdeck: Fix typo in defining peripherals
- deck_info: Do not conflict when buses are shared


evoggy (approved):
Looks good!

PR #829, opened 2021-08-17
Allow non-standard IMU orientation (mostly for Bolt)

matejkarasek:
This PR enables to define non-standard orientation of the IMU on the airframe, e.g. when using the Crazyflie Bolt on unconventional aircrafts.
The user can specify the Euler angles of the IMU using the defines in the config file.


tobbeanton (approved):
Code looks good. Testing is needed.

matejkarasek:
As for the testing, I did test the code with:

1) standard CF2.1 and default IMU orientation (IMU_PHI=0.0f, IMU_THETA=0.0f, IMU_PSI=0.0f) 

2) a CF2.1 with an upside-down PCB, manual flight as well as position control in both LPS and Lighthouse :
https://user-images.githubusercontent.com/10943061/134651320-94adfd33-1af2-443c-8617-1f8f7c5118b6.mp4
https://user-images.githubusercontent.com/10943061/134651302-33ec14b0-ba20-4c30-b951-d5c2783f036d.mp4
(not spot on hover, but in our setup it oscillates like this also with standard firmware and orientation)
I didn't keep the code, but I think I used IMU_PHI=180.0f and I had to swap motor numbers & signs in the power distribution.

3) with our Flapper Drone, where the Bolt is mounted vertically, we use IMU_THETA=90.0f and IMU_PSI=180.0f

While I beleive this is enough evidence, always good if someone could reproduce the tests.
And, any volunteer to test "non-orthogonal" orientations? :)

tobbeanton:
@matejkarasek Could you try to resolve the merge conflict? The history is better preserved if you do it.

matejkarasek:
Done!

tobbeanton (approved):

tobbeanton:
Thanks!

I think this functionality is a really good start which also highlights that we should make this easier to configure in the future. The functionality is also overlapping the pitch and roll trimming which I think should be replaced with a easy to configure Euler angle setup in the future. Well that is then, let's start with this step first :-)

matejkarasek:
Great! I agree it would be nice to have this easily configurable, e.g. through some sort of persistent setting, and then it could indeed replace the roll/pitch trims...

PR #830, opened 2021-08-17
Add systemRequestShutdown() to allow apps to request a full system shutdown

ntamas:
This PR adds a new function called `systemRequestShutdown()` to the public API of the `system` module. The purpose of the function is to allow apps to request a full, immediate system shutdown.

This would enable an app developer to add extra safety features, e.g., some kind of a "safety fence" functionality that stops the Crazyflie immediately if the position estimate diverges outside a predefined bounding box.

### DIFF
--- a/app_api/src/app_main.c
+++ b/app_api/src/app_main.c
@@ -46,6 +46,7 @@
 #include "param.h"
 #include "pm.h"
 #include "app_channel.h"
+#include "system.h"
 
 
 #define DEBUG_MODULE "APPAPI"
@@ -131,5 +132,10 @@ void appMain() {
     appchannelSendPacket("hello", 5);
     appchannelReceivePacket(buffer, APPCHANNEL_MTU, APPCHANNEL_WAIT_FOREVER);
     appchannelHasOverflowOccured();
+  }
+
+  // System
+  {
+    systemRequestShutdown();
   }
 }
--- a/src/hal/src/pm_stm32f4.c
+++ b/src/hal/src/pm_stm32f4.c
@@ -149,11 +149,7 @@ static void pmSetBatteryVoltage(float voltage)
 static void pmSystemShutdown(void)
 {
 #ifdef ACTIVATE_AUTO_SHUTDOWN
-  SyslinkPacket slp;
-
-  slp.type = SYSLINK_PM_ONOFF_SWITCHOFF;
-  slp.length = 0;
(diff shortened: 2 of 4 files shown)


jonasdn:
Thank you!

So, right now we also have `powerStop()`, from `power_distribution.h`, which is called when you call `systemEmergencyStop()`, or set the parameter [stabilizer.stop](https://www.bitcraze.io/documentation/repository/crazyflie-firmware/master/api/params/#stabilizerstop) to true.

The `powerStop()` function will attempt to turn of power to all engines, it is called fro various fault handlers today, and also when the supervisor (`supervisor.c`) module detects that the quad copter is tumbled.

Could you spell out in the commit message when this `systemRequestShutdown()` should be used and `powerStop()` is not sufficient?

Thanks again!

ntamas:
I guess you meant `stabilizerSetEmergencyStop()` instead of `systemEmergencyStop()`, right? (I could not find `systemEmergencyStop()` and `stabilizerSetEmergencyStop()` seems to be the only public function that I could call in order to set `emergencyStop` to `true`, apart from messing around with the parameter subsystem). The problem with `stabilizerSetEmergencyStop()` is that the GCS can easily undo its effect by calling `stabilizerSetEmergencyStopTimeout()`; imagine a Python script running on a computer that repeatedly calls `stabilizerSetEmergencyStopTimeout()` in a loop.

Now that I was browsing the code a little bit, I noticed that there is also `systemSetArmed()` that could potentially be used to stop the motors, but this can also be overridden with the `system.forceArm` parameter. In particular, if you use brushless motors and `-DSTART_DISARMED -DDEFAULT_IDLE_THRUST=5000`, you need to set `system.forceArm` first before starting the flight, so this is probably not sufficient either. All in all, I could not find a way that stops the motor with certainty (irrespectively of prior settings and irrespectively of what the GCS is doing) apart from shutting down the system completely.

My primary goal is to have a safety feature that the app layer can invoke if it feels that the motor should be stopped immediately. An alternative to this PR would be to ensure that:

1. Emergency stops invoked from "within" the board cannot be overruled from the GCS, or
2. Expose `systemSetArmed()` via CRTP commands so one could "arm" the drone from the GCS if the firmware is built with `-DSTART_DISARMED`; `system.forceArm` could then be removed and the app layer could use `systemSetArmed(false)` in emergencies.

Issue #577, opened 2020-05-05
Consider reworking the setpoint struct


krichardsson:
The setpoint struct (https://github.com/bitcraze/crazyflie-firmware/blob/2020.04/src/modules/interface/stabilizer_types.h#L170-L191) is very flexible, but also makes controller implementations hard to understand. Maybe we can find a better structure?

Comments from #567:

@whoenig 
> I don't like the setpoint struct that much for the reason that not all combinations make sense. It would be nicer to have an enum and union for the desired "flight mode". Then each controller could also decide which flight modes it can actually support properly. In the situation awareness framework, we would only enable the collision avoidance in some of the flight modes. The firmware controllers already go that route and only support setpoint mode combinations that are "common", but it is hard to decipher which combinations are even valid.

@jpreiss 
> I think the concept of "flight mode" could also potentially subsume the state machine logic in planner.c, the switching between low-level and high-level modes in commander.c, and the low-level setpoint packet types in crtp_commander_generic.c.

jpreiss:
Thinking about this some more.

## Flaws of current design
First, there are some aspects of the current design that could be improved without changing the overall philosophy:

- I can't think of a case when x and y position modes need to be different. We could consolidate them in `mode.xy`. Same for roll/pitch modes.
- If z control is disabled, then a thrust input is required, but the mode enum does not make this obvious.
- The setpoint contains an acceleration vector, but the enum doesn't indicate how/if the controller should use it.
- It is possible that we want to control both {xy abs, xy vel} with vel as a feedforward term. The setpoint struct can express this, but the enum values cannot. Same for {quat, attitude vel} and {attitude abs, attitude vel}.

## Rules of factored design

Next, I tried to figure out the rules to decide whether a combination of mode flags is valid. I am imagining that we add a `mode.thrust` field to explicitly state if the thrust input is to be used.

Note that it is OK for more then one of {xy, attitude abs, quat, attitude rate} to be controlled, because the higher order derivatives can be interpreted as feedforward terms by the following ordering:

    xy abs <- xy vel <- {attitude abs, quat} <- attitude rate

Of course, the user could send nonsensical commands like constant position over time with nonzero velocity, but that is "undefined behavior". On the other hand, breaking any of these rules makes a setpoint nonsensical on its own:

- quat should never be vel - if we want to control angular velocity, use body Euler angles.
- thrust can never be vel.
- If quat is abs, then none of roll/pitch/yaw can also be abs.
[...]

PR #833, opened 2021-08-19
Enabling PID position control in body-aligned coordinates

matejkarasek:
This PR allows to control the position and velocity in body-aligned coordinates, instead of global coordinates (currently the only option). The change only affects the PID controller.
This functionality is useful when flying with non-symmetric vehicles whose flight dynamics along x and y directions are not the same, and thus require different gain tuning.
It can be turned on and off through a define POSITION_CONTROL_IN_BODY.

Further updates in this PR:
- low pass filters in the position and control loops are now separated: position and velocity as well as xy and z directions can be set separately
- the above settings can be changed through configuration parameters
- feed-forward terms and gains were added for the velocity loop (useful for vehicles with non-negligible drag)


knmcguire:
Hi!

It is an interesting point though, but perhaps there is some discussion necessary for this first. The firmware was initially built to work for a quadcopter mainly but with the bolt it's been expanding a lot now. Theoratically perhaps, I would vote that the the lower level controllers are the ones that are different for different platforms but reality proves that will be difficult.

The current problem with PR is that the velocity_body boolean now makes it a bit complicated, and more over, 2 different controllers complicates the code even further... 

I was thinking, since the PID gains for the crazyflie is the same for the x and y position and velocity controller due to the symmetry, doesn't it make sense to replace the global coordinate PID based controller entirely with a body fixed one? Technically  as long as the state and setpoint are rotated correctly, the control result should be completely similar, don't you think @matejkarasek ? 

I'll also discuss this internally

knmcguire:
I've discussed it with some here, and there are a couple of factors. 

First of all, we need to have a good look at the way we handle the setpoint modes (relative vs global, position vs velocity) as that currently not super logical. But this is part of a bigger discussion of the whole framework and is probably not part of this pull request (maybe I'll make a separate issue to discuss this)

Second, I do think that it will not make a difference for the crazyflie if we would make the controller in body standard. However, since more rotations are done for the in body controller, there might be more CPU load because of it, albeit it probably is just small. 

The one who original implemented the PID controller is @ataffanel, so once he is back from his holiday, I'll involve him in the discussion.

matejkarasek:
Hi Kimberly,
Thanks for looking into this!

Yeah, I am aware that this affects the core functionality, that is why I make this optional... But I agree there is now maybe too many options. The velocity_body boolean indeed increases the complexity even further...

I personally think that control in the body coordinates makes more sense for any platform, and should indeed make no big difference if implemented correctly for symmetric platforms (most quad/multi copters).

This does not necessarily introduce more body rotations - even with the current control in global coordinates, you need to rotate the output eventually since attitude control is done around body-fixed axes.
With control in body coordinates you just do this rotation earlier.

matejkarasek:
As for the setpoints and their structure, I agree this for for a larger discussion and outside of the scope of this PR.
But already removing the velocity_body boolean and having a function instead that will rotate the setpoint from the user to the setpoint coordinates (be it global or body-fixed) could make things a bit simpler?

ataffanel:
I do not have much input about the functionality in itself. Velocity in the body coordinate system makes some sense to have onboard and supporting asymmetrical platforms is most likely going to be nice to have for the CF Bolt. I cannot wrap my head around body-aligned position control though, but I can accept that it is useful :).

As for the set-point structure: my point of view is that it is better to put these king of info in the set-point structure mostly if the alternative is to add a new state to the controllers: with the current approach there is never uncertainty on what the controller should do since we pass control parameters for each setpoints, if the setpoint meaning is an external state we start having problems like synchronization issues: for which setpoint in the stream does the parameter change. So, making the structure narrower in scope and more adapted to real-life limitations like discussed in #577 sounds better to me.

matejkarasek:
@ataffanel The body aligned position control seems a bit strange indeed, but think of it as if you were sitting on board: How much forward/backward and how much left/right do I need to go. 
If the xy gains are identical, this will make no difference.

I see your point with the setpoints... I think the current situation is acceptable, just the code around the velocity_body boolean requires some attention when reading it, as it is not so straightforward.

So if there is no objection against including this functionality, would you prefer to:
1) Let the user decide what to use using the defines as it is implemented now (advantage: safer for current users, disadvantage: more code to maintain)
2) Switch to "in body" by default (simpler code, but some extra transformations needed as the states are saved in global coordinates)

Thanks!

matejkarasek:
I have resolved the conflicts with the current master. 
Any thoughts on this PR in the meantime? :)

matejkarasek:
Closing, replaced by #885

Issue #840, opened 2021-09-02
Spaces in current path causes build errors


jonasdn:
To reproduce:

```
$ mkdir "dir with spaces"
$ cd "dir with spaces"
$ git clone git@github.com:bitcraze/crazyflie-firmware.git
$ cd crazyflie-firmware/
$ git submodule init
$ git submodule update
$ make
  CLEAN_VERSION
  CC    list.o
  CC    tasks.o
[...]
make -C .//tools/make/cmsis_dsp/ CRAZYFLIE_BASE=/home/jonasdn/sandbox/dir with spaces/crazyflie-firmware PROJ_ROOT=/home/jonasdn/sandbox/dir with spaces/crazyflie-firmware V= CROSS_COMPILE=arm-none-eabi-
make[2]: *** No rule to make target 'with'.  Stop.
make[1]: *** [Makefile:408: libarm_math.a] Error 2
rm version.c
make: *** [Makefile:393: build] Error 2
```

jonasdn:
From current GNU Make maintainer:

```
%% "Goszczynski, Piotr" <address@hidden> writes:

  gp> Please advise how to handle Windows' whitespace unfortunate
  gp> innovation in the GNU environment.

The short answer is, "don't".  Don't create files and directories with
whitespace in them.

I can't speak for the entire GNU environment, but there is really no
chance you can use files with whitespace in them throughout most of
make.  In certain specific areas you can sometimes escape the space, but
not consistently, and not everywhere.

If there is absolutely no way to avoid it, here are the only two tricks
I know:

 1) Use 8.3 filenames.  Most MS filesystems will accept either 8.3
    (which won't have whitespace) or long filenames.

    The problem here is that the exact 8.3 name of a file can _change_
    if you add, remove, or rename other files in the directory :-/.

 2) Use Window's (DOS'?) alias feature to alias a drive letter to the
    top of your directory structure, then be sure that no spaces appear
    _within_ the directory structure.

    So, in your example, you could alias H: (or whatever) to
    "C:/SomeDir/Dir With Space", then use H:/My_MakeLocation instead.
    Don't ask me exactly how to do this, I know very little about
    Windows/DOS.  But, it can be done and the info shouldn't be too hard
    to find :).

```

jonasdn:
I think this is to much effort to low of a gain. We should just state that we do not support spaces in the path leading up to the repo.

PR #841, opened 2021-09-02
Makefile: Avoid referencing users absolute path

jonasdn:
We currently get build errors when there is spaces in the absolute path leading up to the crazyflie-firmware directory:

```
$ mkdir "dir with spaces"
$ cd "dir with spaces"
$ git clone git@github.com:bitcraze/crazyflie-firmware.git
$ cd crazyflie-firmware/
$ git submodule init
$ git submodule update
$ make
  CLEAN_VERSION
  CC    list.o
  CC    tasks.o
[...]
make -C .//tools/make/cmsis_dsp/ CRAZYFLIE_BASE=/home/jonasdn/sandbox/dir with spaces/crazyflie-firmware PROJ_ROOT=/home/jonasdn/sandbox/dir with spaces/crazyflie-firmware V= CROSS_COMPILE=arm-none-eabi-
make[2]: *** No rule to make target 'with'.  Stop.
make[1]: *** [Makefile:408: libarm_math.a] Error 2
rm version.c
make: *** [Makefile:393: build] Error 2
```

The general advice to handling whitespace in GNU Make is "don't", it gets very messy, see: https://www.cmcrossroads.com/article/gnu-make-meets-file-names-spaces-them.

In our case I think we want to try hard not to reference any paths out side of our own control. So this commit tries to use relative paths instead of absolute ones.

Fixes: #840

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -405,7 +405,7 @@ bin/vendor:
 	mkdir -p bin/vendor
 
 libarm_math.a:
-	+$(MAKE) -C $(CRAZYFLIE_BASE)/tools/make/cmsis_dsp/ CRAZYFLIE_BASE=$(abspath $(CRAZYFLIE_BASE)) PROJ_ROOT=$(CURDIR) V=$(V) CROSS_COMPILE=$(CROSS_COMPILE)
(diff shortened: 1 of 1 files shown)


jonasdn:
This is tricky to get right with regards to out-of-tree build and the app framework.

PR #843, opened 2021-09-03
Fix incorrect PID filter reset introduced in #803

matejkarasek:
This PR fixes a bug introduced in #803 

The function `positionControllerResetAllfilters()` reinitialized the z velocity filter to ZVELOCITY_LPF_CUTOFF_FREQ even if IMPROVED_BARO_Z_HOLD was not defined. This is now fixed and the filter gets reset to POSITION_LPF_CUTOFF_FREQ by default.


jonasdn (approved):
Thanks!

PR #844, opened 2021-09-03
Add graceful shutdown support

### LINKED ISSUE #623: Add graceful shutdown support
Currently to shut-down the Crazyflie, the nRF51 simply cut the power to the system. This was working fine until we started implementing SD-card functionality: when shutting down the system that way we have huge chances of corrupting the SD-card filesystem.  This has lead to implementation of very inefficient sd-car logging procedure to decrease the probability of corruption.

A better solution for the shutdown problem would be to implement a graceful shutdown in the system:
 - When the nRF51 wants to shutdown the system (ie. after a button press), it sends a message to the stm32 requesting shutdown.
 - In the stm32, modules that wants to be notified of shutdown are called in sequence and return when they are ready for shutdown
 - When all the module have returned, the stm32 sends to the nRF51 a "ready for shutdown" message
 - The nRF51 cuts the power. If no "ready to shutdown" was received after a timeout, the power is cut anyway.

This will for now only be used for the SD-card, but it could be useful for other part of the system as well in the future.

jonasdn:
Currently to shut-down the Crazyflie, the nRF51 simply cut the power to the system. This was working fine until we started implementing SD-card functionality: when shutting down the system that way we have huge chances of corrupting the SD-card filesystem. This has lead to implementation of very inefficient sd-car logging procedure to decrease the probability of corruption.

This PR suggests a better solution for the shutdown problem would be to implement a graceful shutdown in the system ...

  ... when the nRF51 wants to shutdown the system (ie. after a button press), it sends a message to the stm32 requesting shutdown

  ... in the stm32, modules that wants to be notified of shutdown are called in sequence and return when they are ready for shutdown

  ... when all the module have returned, the stm32 sends to the nRF51 a "ready for shutdown" message

  ... the nRF51 cuts the power. If no "ready to shutdown" was received after a timeout, the power is cut anyway

Closes #623


jonasdn:
@whoenig @ataffanel I would like to add a commit to provide graceful shutdown for the usddeck, could you help me with that? Or point me towards a source than could?

thanks!

src/hal/interface/syslink.h
```diff
@@ -64,8 +64,10 @@
 #define SYSLINK_OW_READ     0x22
 #define SYSLINK_OW_WRITE    0x23
 
-#define SYSLINK_SYS_GROUP        0x30
-#define SYSLINK_SYS_NRF_VERSION  0x30
+#define SYSLINK_SYS_GROUP            0x30
+#define SYSLINK_SYS_NRF_VERSION      0x30
+#define SYSLINK_SYS_SHUTDOWN_REQUEST 0x31
+#define SYSLINK_SYS_SHUTDOWN_ACK     0x32
```
ataffanel, line 70: Same comment at on the nRF side, I think this should be in the PM group.

ataffanel (commented):
Just a nitpick otherwise I like it!

whoenig:
I agree that this should only be merged after we have at least one callback registered and tested. I would sign up for the new callback after it is confirmed that the usd card can be mounted and the config be read, namely at https://github.com/bitcraze/crazyflie-firmware/blob/949d3786c01e6bc9a4e858afeddf8b08485faacb/src/deck/drivers/src/usddeck.c#L684. In the writer-task, you could check for a new flag, e.g., `not_shutdown` that is initially false at https://github.com/bitcraze/crazyflie-firmware/blob/949d3786c01e6bc9a4e858afeddf8b08485faacb/src/deck/drivers/src/usddeck.c#L791. In your callback you would switch the flag to true and wait until the task is finished. This sort of wait is the tricky part of the whole design.

jonasdn:
@ataffanel @whoenig would something like the latest commit work?

src/deck/drivers/src/usddeck.c
```diff
@@ -92,6 +93,10 @@
 #define FIXED_FREQUENCY_EVENT_ID          (0xFFFF)
 #define FIXED_FREQUENCY_EVENT_NAME        "fixedFrequency"
 
+
+/* set to true when graceful shutdown is triggered */
+static volatile bool in_shutdown;
```
jonasdn, line 98: this is implicit false ... should I make it explicit?

```
static volatile bool in_shutdown = false;
```

src/deck/drivers/src/usddeck.c
```diff
@@ -521,6 +527,14 @@ static void usddeckEventtriggerCallback(const eventtrigger *event)
   }
 }
 
+static void usdGracefulShutdownCallback()
+{
+  uint32_t timeout = 15; /* ms */
+  in_shutdown = true;
+  vTaskResume(xHandleWriteTask);
+  ulTaskNotifyTake(pdTRUE, timeout / portTICK_PERIOD_MS);
```
jonasdn, line 535: Should this be ...

```c
ulTaskNotifyTake(pdTRUE, pdMS_TO_TICKS(timeout));
```

...?

src/deck/drivers/src/usddeck.c
whoenig: Sure!

src/deck/drivers/src/usddeck.c
whoenig: Yes, that would be cleaner.

whoenig (commented):
The changes look good to me - I especially appreciate your timeout rather than an indefinite wait. However, I think the change is major enough that it should actually be tested:-) Does it work for you properly?

jonasdn:
> The changes look good to me - I especially appreciate your timeout rather than an indefinite wait. However, I think the change is major enough that it should actually be tested:-) Does it work for you properly?

I have now, and fixed some bugs together with @ataffanel

PR #845, opened 2021-09-03
motors: Remove unnecessary function calls

### LINKED ISSUE #705: Unnecessary function calls
In motorsInit there are two unnecessary function calls with the currently used motor timers (2, 3, 4, 5 and 14 according to [motors_def_cf2.c](https://github.com/bitcraze/crazyflie-firmware/blob/master/src/drivers/src/motors_def_cf2.c)).

- [MOTORS_TIM_DBG_CFG](https://github.com/bitcraze/crazyflie-firmware/blob/master/src/drivers/src/motors.c#L158) is defined in [motors.h](https://github.com/bitcraze/crazyflie-firmware/blob/master/src/drivers/interface/motors.h#L57) as [DBGMCU_APB2PeriphConfig](https://github.com/bitcraze/crazyflie-firmware/blob/master/src/lib/STM32F4xx_StdPeriph_Driver/src/stm32f4xx_dbgmcu.c#L152). Since the used STM32F405 timers are connected via bus APB1, according to the reference manual, the define should be changed (or removed).

- [TIM_CtrlPWMOutputs](https://github.com/bitcraze/crazyflie-firmware/blob/master/src/drivers/src/motors.c#L160) is only used to [enable](https://github.com/bitcraze/crazyflie-firmware/blob/master/src/lib/STM32F4xx_StdPeriph_Driver/src/stm32f4xx_tim.c#L2265) the advanced timers 1 & 8. It therefore has no effect on the current configuration.

jonasdn:
From user X-N-C in issue #705:

  In motorsInit there are two unnecessary function calls with the
  currently used motor timers (2, 3, 4, 5 and 14 according to
  motors_def_cf2.c).

  MOTORS_TIM_DBG_CFG is defined in motors.h as DBGMCU_APB2PeriphConfig.
  Since the used STM32F405 timers are connected via bus APB1, according
  to the reference manual, the define should be changed (or removed).

Closes #705

### DIFF
--- a/src/drivers/src/motors.c
+++ b/src/drivers/src/motors.c
@@ -228,10 +228,6 @@ void motorsInit(const MotorPerifDef** motorMapSelect)
     // Configure Output Compare for PWM
     motorMap[i]->ocInit(motorMap[i]->tim, &TIM_OCInitStructure);
     motorMap[i]->preloadConfig(motorMap[i]->tim, TIM_OCPreload_Enable);
-
-    MOTORS_TIM_DBG_CFG(motorMap[i]->timDbgStop, ENABLE);
-    //Enable the timer PWM outputs
-    TIM_CtrlPWMOutputs(motorMap[i]->tim, ENABLE);
   }
 
   // Start the timers


jonasdn:
@tobbeanton do you see any issues with removing these?

PR #850, opened 2021-09-15
Fix setting the parameter by name for CORE parameters.

whoenig:
The introduction of core parameters missed a bitmask which caused EINVAL
to be returned when trying to set a parameter by name.

### DIFF
--- a/src/modules/src/param.c
+++ b/src/modules/src/param.c
@@ -422,7 +422,7 @@ static char paramWriteByNameProcess(char* group, char* name, int type, void *val
     return ENOENT;
   }
 
-  if (type != params[ptr].type) {
+  if (type != (params[ptr].type & (~(PARAM_CORE | PARAM_RONLY)))) {
     return EINVAL;
   }
 


whoenig:
Too bad that our tests didn't catch it. The crazyswarm uses this firmware functionality, but the relevant tests are disabled because there was no way to force reading back a parameter. If the python lib supports setting a parameter by name, it might be worthwhile adding a test there (a fix on the crazyswarm side will only happen after we switched to the crazyflie-link-cpp).

jonasdn:
> Too bad that our tests didn't catch it. The crazyswarm uses this firmware functionality, but the relevant tests are disabled because there was no way to force reading back a parameter. If the python lib supports setting a parameter by name, it might be worthwhile adding a test there (a fix on the crazyswarm side will only happen after we switched to the crazyflie-link-cpp).

Was thinking the same thing. We should make sure to add a test for this.

Thanks for catching this brown paper bag bug!

jonasdn:
So I added using this code path in crazyflie-lib-python: https://github.com/bitcraze/crazyflie-lib-python/commit/fde5d10c234ea0c9140445eed4e81a893bb23a2a

And I added a test to crazyflie-testing: https://github.com/bitcraze/crazyflie-testing/commit/0fe9eb43c1d9d582f76f16f4715fb16a2b536c03

Then I ran the crazylab tests and observed them failing:
https://github.com/bitcraze/crazyflie-testing/runs/3649693743?check_suite_focus=true

jonasdn:
Now I will merge this and restart the crazylab tests.

whoenig:
Very nice - thanks!

jonasdn:
And now it passes: https://github.com/bitcraze/crazyflie-testing/actions/runs/1253237262

PR #856, opened 2021-09-28
aideck: Initial ninalink implementation

jonasdn:
Syslink inspiried protocol with the aideck NINA module.

### DIFF
--- a/src/deck/drivers/src/aideck.c
+++ b/src/deck/drivers/src/aideck.c
@@ -46,14 +46,115 @@
 #include "system.h"
 #include "uart1.h"
 #include "uart2.h"
+#include "static_mem.h"
+
+#define NINALINK_MAGIC "\xbc\xa1"
+#define NINALINK_MTU 128
 
 static bool isInit = false;
 static uint8_t byte;
 
-//Uncomment when NINA printout read is desired from console
-//#define DEBUG_NINA_PRINT
+typedef enum {
+  NOP,
+  PRINT,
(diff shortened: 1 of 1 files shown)


tobbeanton:
As this is inspired from the syslink I know that we at some point would like it do the communication more efficient. This probably means using DMA in some way. I have not looked into it but it would be good if the protocol works with a DMA scheme, which it might already do.

src/deck/drivers/src/aideck.c
```diff
@@ -70,15 +171,13 @@ static void NinaTask(void *param)
...
-    while (1)
-    {
-        if (uart2GetDataWithDefaultTimeout(&byte) == true)
-        {
-            consolePutchar(byte);
+    while (1) {
+        if (uart2GetDataWithDefaultTimeout(&byte) == true) {
+            ninalinkReceive(byte);
         }
+        vTaskDelay(M2T(10));
```
ataffanel, line 178: This delay should not be needed and might make it loose packet. Since "uart2GetData" blocks, it will take care of limiting the cpu usage of the task.

ataffanel (changes requested):
Looks good in principle except for the sleep. I am about to review/rewrite Syslink to add flow-control. Lets keep this implementation in sync when we have the new safelinkk (we might even be able to share code between the two procotols?).

knmcguire:
in light of the latest aideck constructions going on, is this PR still valid?

PR #859, opened 2021-10-01
Initial port to the Linux kernel build system

jonasdn:
Screenshot from 2021-10-22 09-54-02

---

Screenshot from 2021-10-22 09-41-46


knmcguire:
Hi! I just tried it out. I'm getting the following error if I do 'make menuconfig'

```
  HOSTCC  scripts/kconfig/mconf.o
<command-line>: fatal error: curses.h: No such file or directory
compilation terminated.
make[1]: *** [scripts/Makefile.host:108: scripts/kconfig/mconf.o] Error 1
make: *** [Makefile.kbuild:393: menuconfig] Error 2
```

jonasdn:
> 

Thanks for testing! Could you try:
```
sudo apt-get install libncurses5-dev libncursesw5-dev
```

knmcguire:
yes that works, thanks!

knmcguire:
App layer configuration menu item: 
- no response on Build API test application
- Enable app entry port

Also, was it the idea hear to include the app examples?

Kconfig
```diff
@@ -0,0 +1,126 @@
...
+    default 0
+    help
+      Set the task priority between 0 and 5. Default is 0 (same as IDLE).
+
+endmenu
+
+menu "Expansion deck configuration"
+
+config DECK_FORCE
+    string "Force loading of a deck driver"
```
knmcguire, line 119: The string now looks like '() Force loading of a deck driver' but it it unclear that that line does any thing specific or goes to a different windo. Perhaps better to do:
'Force loading of a manual deck driver --->'

(also indicating that it is not a generic one is important)

docs/development/oot.md
```diff
@@ -5,21 +5,49 @@ page_id: oot
...
+#
+OOT_CONFIG := $(PWD)/config
+
+include $(CRAZYFLIE_BASE)/tools/make/oot.mk
 ```
-CFLAGS += -DOOT_ESTIMATOR
 
-VPATH += src/
-PROJ_OBJ += estimator_out_of_tree.o
+This will make the crazyflie-firmware build system look for a `Kbuild` file in your folder.
```
knmcguire, line 23: Currently the kbuild is unresponsive if you don't make it inside of an example folder or oot estimator folder... Is it possible to have some kind of warning or feedback that this is what is happening?

knmcguire (changes requested):

jonasdn:
> App layer configuration menu item, no response on when selecting:
> 
> * Build API test application
> * Enable app entry port
> 
> Also, was it the idea hear to include the app examples?
What do you mean with no response? In my setup I can select / deselect them.

app-layer

No the App examples should not be included in menuconfig. Are they there for you?

docs/development/oot.md
jonasdn: > Currently the kbuild is unresponsive if you don't make it inside of an example folder or oot estimator folder... Is it possible to have some kind of warning or feedback that this is what is happening?

Could you give me an example? I do not quite understand.

Makefile
```diff
@@ -1,455 +1,126 @@
...
-CLOAD_CMDS        ?=
-CLOAD_ARGS        ?=
-PLATFORM          ?= cf2
-LPS_TDMA_ENABLE   ?= 0
-LPS_TDOA_ENABLE   ?= 0
-LPS_TDOA3_ENABLE  ?= 0
 
 # Cload is handled in a special way on windows in WSL to use the Windows python interpreter
 ifdef WSL_DISTRO_NAME
-CLOAD_SCRIPT      ?= python.exe -m cfloader
```
knmcguire, line 32: So... now 'make cload' does not work on my WSL anymore because of this line change which has been reverted.

knmcguire:
> > App layer configuration menu item, no response on when selecting:
> > 
> > * Build API test application
> > * Enable app entry port
> > 
> > Also, was it the idea hear to include the app examples?
> 
> What do you mean with no response? In my setup I can select / deselect them.
> 
> app-layer
> 
> No the App examples should not be included in menuconfig. Are they there for you?

It was more of a suggestion to make what is happening a bit clearer. But no I'm not seeing it.

About the unresponsiveness, if I do 'make menuconfig' in the main crazyflie-firmware folder, not on an example folder, the menu item does not respond.

Kconfig
jonasdn: > The string now looks like '() Force loading of a deck driver' but it it unclear that that line does any thing specific or goes to a different windo. Perhaps better to do: 'Force loading of a manual deck driver --->'
> 
> (also indicating that it is not a generic one is important)

Hmm, ok, is this better?

force

Makefile
jonasdn: Great  catch! Will re-add!

jonasdn: or rather, it is still there, right? Just further up

```
# Cload is handled in a special way on windows in WSL to use the Windows python interpreter
ifdef WSL_DISTRO_NAME
PYTHON      ?= python.exe
else
PYTHON      ?= python3
endif

DFU_UTIL          ?= dfu-util

CLOAD_SCRIPT      ?= $(PYTHON) -m cfloader
```
Is this not the same?

docs/development/oot.md
jonasdn: Strange, I can select it fine.

jonasdn: From where are you doing `make menuconfig` ?

Makefile
knmcguire: hmmm.. looking further in the makefile, it should have worked... but it still picking the python program from within WSL

knmcguire:
I've pushed an commit with a documentation suggestion. I've separated the kbuild instructions to a separate file in development and refered to there from the install instructions. The problem is that they kbuild part of the install instructions were quite big and pushed the flashing and such all the way down which are relevant for developers that don't want to customize. Perhaps here is also where the extra library installation instructions should be too (extra barrier for people that like to try everything)

docs/development/oot.md
knmcguire: the main crazyflie-firmware repo

knmcguire:
However, the make cload from WSL is now broken... and I don't know why. It doesn't seem logical when I look at the makefile. It works in master

Kconfig
knmcguire: Perhaps better to add the arrow to make it consistent with the main menu item ? That indicates that there is a next dialog upcoming

jonasdn: > Perhaps better to add the arrow to make it consistent with the main menu item ? That indicates that there is a next dialog upcoming

Yeah, I am not sure about this, since we are not doing anything special in term of Kconfig, this is how it looks when you specify the type string for a configuration option. Perhaps it should conform to how other Kconfig configs look. We do not draw the arrows on menu items, the Kbuild system does.

Busybox has strings they look like:

busybox

udhcp

Perhaps its more the fact that we have no default? Should we have one? So it would look more natural?
Like so:
default


With help:
help-none

jonasdn:
> I've pushed an commit with a documentation suggestion. I've separated the kbuild instructions to a separate file in development and refered to there from the install instructions. The problem is that they kbuild part of the install instructions were quite big and pushed the flashing and such all the way down which are relevant for developers that don't want to customize. Perhaps here is also where the extra library installation instructions should be too (extra barrier for people that like to try everything)

Very nice!

Makefile
knmcguire: yeah it should be the same if I look at it visually.... but if I do make cload in wsl.... it pick just regular python3 and not python.exe. So there is something wrong here still

Kconfig
knmcguire: ah oke, yeah then it's a bit more clear of what it supposed to be

Makefile
knmcguire: perhaps @ataffanel can see what is going on here?

jonasdn: Ah, it is because PYTHON is being set down in `Makefile.kbuild`. And since we go:

```
PYTHON ?= python.exe
```

It will have no effect if `$PYTHON` is set. I will fix so that we force it set when WSL detected.

jonasdn: Fixed in latest commit

krichardsson (changes requested):
I think this is super nice!

I have some comments though:
- As kbuild is a tool I think it would make sense to move it to the tools directory. Maybe put as much as possible in tools/kbuild?
- It seems as a lot of files (.o, .o.cmd, .o.d) end up in the /src tree when building. Is this possible to avoid? In the current setup they end up in the /bin dir? I find it very noisy.
- When building for the bolt the script reports a build for cf2, `Build for the cf2!` and the generated binares are named as cf2
- `make clean` does not remove all binares, for instance tag binaries.

I have to run to a meeting. More testing later...

PR #860, opened 2021-10-03
Add comment about nullSetpoint semantics

jpreiss:
Readers of this code may wonder what `nullSetpoint` means semantically. This comment helps explain it so they do not need to read other files.

Please correct if my interpretation is wrong!


whoenig:
LGTM!

ataffanel:
This is a good addition, makes it less magic. Thanks! Merging.

PR #862, opened 2021-10-04
Improved CPPM - alternative GPIOs, arming and altitude hold, init fix

matejkarasek:
This PR further improves the support of external CPPM receivers:
- alternative GPIO's can be used: PA7(MOSI), PA2(TX2), PA3(RX2), PB4(IO_3), PB5(IO_2) or PB8(IO_1). This allows to use a CPPM receiver alongside many of the decks without pin interference
- switches can be used for arming and triggering altitude hold
- initialization of all channels is improved to prevent undefined behaviour when no physical receiver is connected, or when transmitter is switched on after the crazyflie
- logs are split into two groups: ext_raw for raw channel data (8 channels now) and ext for processed data
- some other minor fixes

### DIFF
--- a/src/deck/interface/deck_core.h
+++ b/src/deck/interface/deck_core.h
@@ -66,6 +66,7 @@ bool deckTest(void);
 #define DECK_USING_I2C     (DECK_USING_PB6  | DECK_USING_PB7)
 #define DECK_USING_TIMER3  (1 << 13)
 #define DECK_USING_TIMER5  (1 << 14)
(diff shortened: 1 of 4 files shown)


matejkarasek:
Flight tested on a CF2.1 with FrSky R-XSR receiver

tobbeanton:
I've tested this with the default PA2 pin and it works well. I have not tested with the others but anyway think this pull request is ok to merge. Well done and many thanks!

matejkarasek:
Great, and thanks for testing it and the merge!

Issue #602, opened 2020-07-02
Python Bindings


whoenig:
In the past, Python bindings for parts of the firmware have been very helpful:

- Simulation of the planning portion as [part of the Crazyswarm](https://github.com/USC-ACTLab/crazyswarm/tree/master/ros_ws/src/crazyswarm/scripts/pycrazyswarm/cfsim),
- Offline parameter tuning of the EKF,
- Software-in-the-loop simulations / controller testing and development.

Other potential applications include:

- Improvement of code quality by better separation of generic quadrotor logic and FreeRTOS-specific functionality,
- Easier automated (unit) testing.

This issue is opened to discuss where to include the python bindings moving forward. There are two options:

1. Move them as part of this repository and invoke the build with `make python-bindings`. Pros: cleaner commit-history for changes related to the bindings, easier test integration, easy integration as part of CI. Cons: "bloat" of this repo (we can put all the logic in a sub-folder, though).

2. Have a separate repo that uses this repo as submodule (similar to the current Crazyswarm solution). Pros: Cleaner code separation. Cons: harder to maintain, difficult to prevent firmware breaking changes because of missing CI integration.

I personally feel option 1 is better in the long run, but I wanted to start this discussion on how everyone else (including @jpreiss) feel about it. The goal is to replace the current solution in the Crazyswarm with these newly proposed bindings and to extend the bindings to other parts of the firmware for improved simulation and ease of tuning.

jpreiss:
I agree with @whoenig's list of benefits.  We could use the version from Crazyswarm mostly unchanged (I think) as a starting point.

The files in [that link](https://github.com/USC-ACTLab/crazyswarm/tree/master/ros_ws/src/crazyswarm/scripts/pycrazyswarm/cfsim) are everything you need to generate the bindings - not that much bloat. Since one of the goals is unit testing, and unit tests usually live in the same repo as the main code, I think option 1 makes sense.

## Impact on code design

I have written several firmware modules with the goal of "bindability" from the beginning. I can share some thoughts on the design changes needed to make current modules more bindable.

In general, the core logic should be implemented in a "library file" that has no transitive dependencies on ARM or FreeRTOS.  The code that plugs this library into FreeRTOS, the param/log system, etc., is in a different "integration file".

### Global variables
[...]

PR #863, opened 2021-10-05
Python bindings draft

jpreiss:
This is the minimal-effort starting point for Python bindings (#602). Every design decision in this PR is subject to change! The bindings compile, unit tests pass, and there is CI.

I thought it would be useful to have a place for design discussions where we can use the code review tools, etc.

Topics for discussion:

- Directory structure and naming
- Integration with main Makefile
- Continue using Conda for CI?
- How much of the Crazyswarm abstraction layer to strip out for the first version (visualizer?)
- Is it ok to ship our own copy of the NumPy SWIG wrapper `numpy.i`?
- Others?


jpreiss:
@whoenig I gave you push access to my fork so you can modify this PR.

whoenig:
Great to start discussing more concretely!

- I think this should be directly integrated in the build, so that `make bindings-python` does the trick
- Naming: In general it might be good to use "bindings-python" to allow Arnaud to have "bindings-rust" later.
- I think there should be no crazyswarm code (or even excessive Python wrappers). If we want to ever have a Python-based Software-in-the-loop simulator with visualization etc, this should be in a separate repo. The purpose for the bindings to me is to be minimal here, adding at most some glue code to allow reasonable tests.
- To focus the discussion, it might be better to have an even more minimalistic PR, that only includes bindings for the smallest module (might be pptraj.c or even just math3d.h)? We can add the other modules step-by-step and have specific discussions during that (e.g., to avoid having hacks like "-D__fp16=uint16_t"). This way we can focus on good integration to the build system, CI, and tests first.
- I don't have strong feelings about conda. It's nice that this allows us to test on mac as well. The downside is that this is probably not the set up most people use.
- Isn't numpy.i part of Swig? I think it's OK to have a copy, as long as we check the license (and include it if necessary).

I am happy to update the draft with the points above, but lets have a discussion first so we don't have code ping/pong:-) Once we are all happy we can squash and merge.

krichardsson:
Hi, great start! 
I agree on most of what you have said so far and my input is:

I like @whoenig approach of a minimalistic PR and could imagine taking it even further. I would be happy to use this first PR mainly to drive the setup of infrastructure (builds, environments, configs and so on) and possibly only implement a tiny part of the bindings (one function?) to make sure everything is in place. Start with a tiny muddy trail from point A to B and later widen it step by step into a fancy highway!

From a Bitcraze perspective we would like to avoid adding new tools, mechanisms or frameworks if possible. Conda is not used today in our systems and if possible it might be a good idea to find other solutions. Today all our builds are done in a docker container based on the bitcraze/builder image and my hope is that we can squeeze this new functionality into the same image. That would also make it possible to run the new tests locally on all platforms (I think).

I have not used python bindings a lot and I do not know very much about it, but I assume there is a dependency between numpy.i and the version of numpy that is installed on the system? As far as I can see there is no mechanism that connects the versions of the two together. Is it possible to download numpy.i from somewhere rather than storing it in git? Add it to the builder docker image together with the numpy installation? Enforce compatibility?

Happy hacking!

krichardsson:
@jpreiss no worries, I will help out with the building/CI parts.

The advantage of putting most of the build process into a docker container is that it can run on your local machine for debugging, also a MAC (possibly except the M1).

> Unfortunately it seems that Docker cannot run on Mac in Github Actions

I don't think there is a need to run in different environments in Github Actions if we use a docker container. The container is providing the runtime environment and it should be identical regardless of the underlying OS. 

Either you can let me know what you would need in the runtime environment for the minimal PR to build and I'll set up a docker image that you can use while working on the PR. The other option is that you simply ignore CI for now and let us (Bitcraze) fix that when we all are happy with the PR.

jpreiss:
I was not thinking clearly. For the Crazyswarm bindings, until now we have compiled a Mac binary and ran the tests natively on Mac. So unless the toolbelt image has a cross compiler, it would not help us even if we could run it on Mac.

I am assuming you want to add the bindings build/test in the toolbelt regardless of how we deal with Mac. So one option is simply forcing all Mac developers to use Docker.

For now, I replaced Conda with pip. Do you think this amount of non-toolbelt CI would be reasonable to keep around for Mac?

whoenig:
I updated the PR to be significantly smaller (just math3d.h bindings for now). The build is now integrated in the main Makefile with a new target (`bindings_python`). This target is not build by default and therefore not tested in the docker container. However, we have a new github actions workflow that only builds the bindings on Mac and Linux.

From what I understand, the question is now if we want to keep it like that, or trust that the docker build on Linux is sufficient for Mac and add CI that uses this particular build target as well.

whoenig:
Thanks Kristoffer! I addressed your two concerns with a new commit. Let me know if you find anything else.

krichardsson:
Thanks Wolfgang! I have update the bitcraze/builder docker image with swig and it should now be possible to build the bindings with the toolbelt by running 

`tb make bindings_python`

@jpreiss @whoenig  Do you want to add one simple test as well in this PR or do you think this is enough? 
I think we could add an initial test in this repo just to make sure it can run end-to-end. If we want we can move it to https://github.com/bitcraze/crazyflie-testing later on, what do you think @jonasdn ? 

The question is how do we run the test? Is it run as a part of the `bindings_python` make target or is it a separate target? It could also be a script in `tools/build`

whoenig:
I think it would be nice to have a simple test in this repo. In particular, trying to import the bindings in Python can already catch many (linker) issues. Functional tests could be in crazyflie-testing or in here (depending if it is considered a unit test or a system test, I think). For consistency with the current unit tests, we could use a new target `bindings_python_unit`?

krichardsson:
Sounds good to me!

jpreiss:
I was imagining we will eventually write many Python unit tests for the bound firmware modules that run in CI for this repo, such that you cannot get a green check mark on your firmware PR without passing the tests. Assuming others share this goal, if we put everything in `crazyflie-testing` how would we accomplish that? Would we clone the `crazyflie-testing` repo in the CI script for this repo?

jonasdn:
> I was imagining we will eventually write many Python unit tests for the bound firmware modules that run in CI for this repo, such that you cannot get a green check mark on your firmware PR without passing the tests. Assuming others share this goal, if we put everything in `crazyflie-testing` how would we accomplish that? Would we clone the `crazyflie-testing` repo in the CI script for this repo?

My opinion is this.

For tests that ...

- ... check that the bindings can still be generated ... or ...
- ... check the correctness of the generated bindings  ... or ...
- ... check coding style or other stylistic aspects of the PR ...

... we add the tests here in the firmware repo.

For tests that uses the generated bindings to ...

- ... check correctness of parts of our estimators / controllers / other aspect of the control loop ... 

... we add tests in the crazyflie-testing repo.

Does this make sense?

jpreiss:
Are you saying that the binding-based correctness tests for the control loop should not be run in CI for this repo? Or would you run them in CI by cloning crazyswarm-testing in the CI script?

whoenig:
I thought about this more and I think anything that can be easily tested within a repo, should be tested as part of the CI there (and of course it should be easy to run it on a local developer computer, too). Anything that requires multiple repositories for testing (e.g., firmware + Python lib), should go in crazyflie-testing.

This also means that some of the correctness tests that James is considering should actually be part of the firmware. When considering this PR, this could be unittests for math3d, verifying that the output matches the equivalent numpy computation.

I am not entirely sure about quantitative analysis tests. For example, if Python bindings exists for a controller, it would be very helpful to run this controller on a virtual quadrotor model and quantify the convergence behavior for a step input etc. I feel that this is not really a unit-test anymore and would therefore belong to crazyflie-testing, as it relies on other large software infrastructure such as a quadrotor simulator.

whoenig:
Any thoughts @jonasdn , @krichardsson ?

jonasdn:
> Any thoughts @jonasdn , @krichardsson ?

I think this sound like a very good guidline!

- If it can be done with what exists in the repo; without cloning any helper library/program then keep it here and run in CI
- If it needs interaction with helper lib or program then move it to crazyflie-testing

That translates to pretty much:

"Run the things that tests the firmware infrastructure from the firmware and run the things that test the Crazyflie as a whole to crazyflie-testing"

whoenig:
Do you guys have any guidelines on where the testing code should be? Perhaps `test/python`? Do you have any preference about pytest vs. unittest?

krichardsson:
The directory structure in test/ maps the structure for the production code and it feels like test/python would break the symmetry. Maybe a new directory on the root level, something like test-system/ or test-module/?

In the python lib we use unittest but in crazyflie-testing we use pytest. Not sure which one is the best, @jonasdn do you have any opinions? (At some point we should maybe consider harmonizing over repositories?)

whoenig:
Thanks Kristoffer! Crazyswarm uses pytest and pytest can execute any regular unittests. So if it would be OK to add this additional dependency to the docker image, pytest would be the more powerful option moving forward.

krichardsson:
Wolfgang, I have added pytest to the builder and it should now work from the toolbelt

whoenig:
I updated the PR to include a test example. @krichardsson How do I enable this test to be run in the CI? Locally, it just requires `make test_python`, which will build the bindings and then run the tests using pytest.

krichardsson:
@whoenig I think the best solution is to add a script in tools/build that runs the python tests. Call that script from tools/build/build which will include it in all builds. Use the tools/build/test as a guide

This will be a good start and if we want to rearrange it later on this will require very little work

whoenig:
Thanks! I added those scripts and CI passes. Marking this PR ready to be reviewed.

knmcguire:
Ah great! I might look into it in light of the simulation work I'm doing now. 

Would it perhaps be a good idea to add short page of this functionality under development instructions in the [documentation](https://www.bitcraze.io/documentation/repository/crazyflie-firmware/master/)?

whoenig:
> Would it perhaps be a good idea to add short page of this functionality under development instructions in the documentation?

Ultimately yes, but this PR has basically zero functionality, because we just tried to agree on the "infrastructure" (integration to CI, where to place the files etc.) Once this is merged, we can contribute a feature PR that contains bindings to what we currently use in the Crazyswarm simulator. Afterwards, I'd like to add bindings for controllers and EKF. All together would be super useful in SIL simulation.

knmcguire:
Ah oke my bad, I thought that this already contained the state estimators and controllers. Yes I agree, it should then be part of the second PR

krichardsson:
Thanks!

I have a general comment on the test. 

From experience I have found that it is good to clearly separate each test into three parts and use comments to indicate where they are:
1. Fixture - set up your test data
2. Test - the actual test
3. Assert - verification of the result

In my opinion keeping the same structure in all tests makes it easier for a reader to figure out what is going on.

whoenig:
Good point - thanks Kristoffer!

Issue #854, opened 2021-09-23
A possibility to override and control the Crazyflie 2.X onboard LEDs from the ground


tobbeanton:
Sometimes it is wanted to control the onboard LEDs from e.g. a script rather then letting them show the build in functionality.
The Crazyflie 2.X has 6 controllable LEDs. 5 of them controlled by the STM32F405 and 1 controlled by the nRF51. As not all LEDs are controlled by the same MCU this complicates things a little.

The first step could be a stealth mode so they could all be off, but if they could be controlled individually this would give even bigger value.

tobbeanton:
This is a first shot at controlling the STM32F405 LEDs through parameters. It needs more work but basically works and could already now be tried out using the patch below.

[cf2_ext_led_control_base308ceb32.zip](https://github.com/bitcraze/crazyflie-firmware/files/7216181/cf2_ext_led_control_base308ceb32.zip)

To override LED control set the led.extCtrl parameter to 1. Setting it to 0 will restore normal functionality.
To control the LEDs set the led.bitpattern parameter. Each bit controls the LED to be on or off and the bit order is according to the led_t enum.
`typedef enum {LED_BLUE_L = 0, LED_GREEN_L, LED_RED_L, LED_GREEN_R, LED_RED_R} led_t;`
Thus setting led.pitpattern to 0x01 would turn on left blue (M3) LED and all others off.

jonasdn:
Coud you create a pull request from the patch in the zip?

jonasdn:
This is possible now on master

PR #865, opened 2021-10-07
Implementation for #854. Control onboard LEDs via parameters.

tobbeanton:
This adds functionality to control the LEDs via parameters. If it is enabled it will override the normal LED control.

The control is done through a bitmask parameter and if bit 7 is set it will be enabled:
| 7:ENABLE | 6:N/A | 5:BLUE_R | 4:RED_R | 3:GREEN_R | 2:RED_L | 1:GREEN_L | 0:BLUE_L |

The blue NRF LED was added to the led driver and will be controlled via the syslink. Thus the nRF FW with this support must run a version including [this](https://github.com/bitcraze/crazyflie2-nrf-firmware/issues/65) issue.

To simplify the update of a change of the parameter a callback option when creating a parameter was added, e.g:
PARAM_ADD_WITH_CALLBACK(PARAM_UINT8, setIO1pin, &pinValue, &myCallbackFunction)

One of the test I did using the a clib script can be found below.
```
import logging
import time

import cflib.crtp
from cflib.crazyflie.syncCrazyflie import SyncCrazyflie
from cflib.utils import uri_helper

# URI to the Crazyflie to connect to
URI = uri_helper.uri_from_env(default='radio://0/80/2M/E7E7E7E7E7')

# Only output errors from the logging framework
logging.basicConfig(level=logging.ERROR)

bitSeq = [0b10000001, 0b10000010, 0b10001000, 0b10100000,
          0b10000001, 0b10000100, 0b10010000, 0b10100000] 

if __name__ == '__main__':
    # Initialize the low-level drivers
    cflib.crtp.init_drivers()

    with SyncCrazyflie(URI) as scf:
        cf = scf.cf

# Create a spinning effect using the onboard LEDs
        for i in range(10):         
            for j in range(8):
                # Set black color effect
                cf.param.set_value('led.bitmask', bitSeq[j])
                time.sleep(0.15)
```

### DIFF
--- a/docs/userguides/logparam.md
+++ b/docs/userguides/logparam.md
@@ -109,6 +109,9 @@ variables in run-time, but note the following:
     logging framework.
 -   The reading or writing of a parameter can be done at any time once
     you are connected to the Crazyflie.
+-   If the callback parameter version is used to get notified that it has
+    changed, it will run from the param task, it should not block and should 
+    run quickly.
 
 ## Logging
 
@@ -210,3 +213,21 @@ Note: The rate computation function is called from the logging framework with th
 specifed in the logging configuration. The rate, on the other hand, is calculated if the time since
 last computation exceeds the configured update time of the rate logger, and if the logging intervall
 is longer than the update intervall, updates will be done for each logging call.
+
+### Parameter callback function to get notifed when a parameter has been updated.
+
+Using the macro `PARAM_ADD_WITH_CALLBACK` it is possible to register a callback function that will be called
+when the parameter gets updated. This callback will run from the parameter task so it is important to not 
+block in this callback or make it run for too long.
+
+Example:
+         void myCallbackFunction(void)
+         {
+            // The parameter has been updated before the callback and the new parameter value can be used
+            digitalWrite(DECK_GPIO_IO1, pinValue);
+         }
+         
+         ...
+         
+         PARAM_ADD_WITH_CALLBACK(PARAM_UINT8, setIO1pin, &pinValue, &myCallbackFunction)
+
--- a/src/drivers/src/led.c
+++ b/src/drivers/src/led.c
@@ -33,6 +33,7 @@
 
 #include "led.h"
 #include "syslink.h"
+#include "param.h"
 
 static GPIO_TypeDef* led_port[] =
 {
@@ -62,7 +63,19 @@ static int led_polarity[] =
   [LED_BLUE_NRF] = LED_POL_POS,
 };
 
-static bool isInit = false;
+static bool isInit = 0;
+static bool enableBitmaskControl;
+static uint8_t ledBitmask;
+
+static void ledBitmaskCallback(void)
+{
+  enableBitmaskControl = (ledBitmask & (1<<7)) != 0;
+
+  for (int i = 0; i < LED_NUM; i++)
+  {
+    ledSetOverride(i, ledBitmask & (1<<i));
+  }
+}
 
 //Initialize the green led pin as output
 void ledInit()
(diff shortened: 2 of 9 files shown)


src/hal/src/syslink.c
```diff
@@ -139,6 +139,12 @@ bool syslinkTest()
   return isInit;
 }
 
+bool isSyslinkUp()
+{
+  return isInit;
+}
+
+
```
jonasdn, line 147: nit: extra newline

src/drivers/src/led.c
```diff
@@ -138,10 +145,29 @@ void ledSet(led_t led, bool value)
...
+  {
+    SyslinkPacket slp;
+    slp.type = value ? SYSLINK_PM_LED_ON : SYSLINK_PM_LED_OFF;
+    slp.length = 0;
+    syslinkSendPacket(&slp);
+  }
   else
-    GPIO_ResetBits(led_port[led], led_pin[led]); 
+  {
+    if(value)
```
jonasdn: nit: space after `if`, and I prefer måsvingar here, but you decide.

src/drivers/src/led.c
```diff
@@ -64,20 +66,77 @@ static int led_polarity[] =
...
+  if (led == LED_BLUE_NRF && isSyslinkUp())
+  {
+    SyslinkPacket slp;
+    slp.type = value ? SYSLINK_PM_LED_ON : SYSLINK_PM_LED_OFF;
+    slp.length = 0;
+    syslinkSendPacket(&slp);
+  }
+  else
+  {
+    if(value)
```
jonasdn, line 98: nit: space after `if` and I prefer _måsvingar_ even on one-liners, but you decide.

docs/userguides/logparam.md
```diff
@@ -109,6 +109,9 @@ variables in run-time, but note the following:
     logging framework.
 -   The reading or writing of a parameter can be done at any time once
     you are connected to the Crazyflie.
+-   If the callback parameter version is used to get notified that it has
+    changed, it will run from the param task, it should not block and should 
+    run quickly.
```
jonasdn, line 115: I am having trouble parsing this, what does _version_ mean it this context?

ah, I think I got it, perhaps:

"If the `PARAM_ADD_WITH_CALLBACK` macro is used [...]" or "if the `PARAM_CALLBACK`type is set [...]"

src/modules/interface/param.h
```diff
@@ -172,16 +172,20 @@ struct param_s {
 #ifndef UNIT_TEST_MODE
 
 #define PARAM_ADD(TYPE, NAME, ADDRESS) \
-   { .type = TYPE, .name = #NAME, .address = (void*)(ADDRESS), },
+  { .type = TYPE, .name = #NAME, .address = (void*)(ADDRESS), },
+
+// The callback notification function will run from the param task, it should not block and should run quickly.
+#define PARAM_ADD_WITH_CALLBACK(TYPE, NAME, ADDRESS, CALLBACK) \
+  { .type = TYPE | PARAM_CALLBACK, .name = #NAME, .address = (void*)(ADDRESS), .callback = (void *)CALLBACK, },
```
jonasdn, line 179: This will cause problems everywhere we check type to see what type this is... right? And in the elf-sanity script that also checks types.

jonasdn: For instance in param.c `paramWriteByNameProces`:

```c
  if (type != (params[ptr].type & (~(PARAM_CORE | PARAM_RONLY)))) {
    return EINVAL;
  }
  ```

This will break by this change?

tobbeanton: Possibly yes but I thought it wouldn't as the PARAM_CALLBACK is using an unused bit. I have to investigate.

PR #866, opened 2021-10-10
cf_math: Avoid function calls with side effects within ASSERT()

whoenig:
Since ASSERT() is potentially optimized out in release builds, avoid
important function calls with side effects.

### DIFF
--- a/src/utils/interface/cf_math.h
+++ b/src/utils/interface/cf_math.h
@@ -7,7 +7,7 @@
  *
  * Crazyflie control firmware
  *
- * Copyright (C) 2018 Bitcraze AB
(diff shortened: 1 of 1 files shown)


whoenig:
Good point. I was working on the EKF Python bindings, and just used `#define ASSERT(e)` for the x86 build, which had the undesired effect. I agree that in regular firmware builds this should not be a problem at all. I think renaming ASSERT would also work, although it seems in most other places in the firmware I've seen ASSERT() is already used like the regular C-style assert, i.e., avoiding any function calls inside the macro. In that sense, this change just improves consistency and readability, at the cost of a slightly longer source file.

PR #869, opened 2021-10-10
Kalman: Move parameters in a separate parameter struct

whoenig:
* Allows to move all Kalman parameters to estimator_kalman
* Better separation between OS-specific (estimator_kalman) and
  OS-independent (kalman_core) code
* Preparation for Python bindings (so that parameters can be tuned
  by adjusting the values in the params struct)

### DIFF
--- a/src/modules/src/estimator_kalman.c
+++ b/src/modules/src/estimator_kalman.c
@@ -149,6 +149,11 @@ static bool quadIsFlying = false;
 
 static OutlierFilterLhState_t sweepOutlierFilterState;
 
+// Indicates that the internal state is corrupt and should be reset
+bool resetEstimation = false;
+
+static kalmanCoreParams_t coreParams;
+
 // Data used to enable the task and stabilizer loop to run with minimal locking
 static state_t taskEstimatorState; // The estimator state produced by the task, copied to the stabilzer when needed.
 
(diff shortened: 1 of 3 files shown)


krichardsson:
Looking good, I like it!

src/modules/src/estimator_kalman.c
```diff
@@ -593,4 +599,52 @@ PARAM_GROUP_START(kalman)
...
+ */
+  PARAM_ADD_CORE(PARAM_FLOAT, mNBaro, &coreParams.measNoiseBaro)
+  /**
+ * @brief Measurement Noise for roll/pitch gyros
+ */
+  PARAM_ADD_CORE(PARAM_FLOAT, mNGyro_rollpitch, &coreParams.measNoiseGyro_rollpitch)
+  /**
+ * @brief Measurement Noise for yaw gyro
+ */
+  PARAM_ADD_CORE(PARAM_FLOAT, mNGyro_yaw, &coreParams.measNoiseGyro_yaw)
```
jonasdn, line 633: I realize this is copy-and-paste, but. Since you are here :) could we fix the spurious capital N in Noise here?

"Measurement Noise for roll/pitch gyros" => "Measurement noise for roll/pitch gyros"

etc.

src/modules/src/estimator_kalman.c
```diff
@@ -593,4 +599,52 @@ PARAM_GROUP_START(kalman)
...
+  /**
+ * @brief Initial Y after reset [m]
+ */
+  PARAM_ADD_CORE(PARAM_FLOAT, initialY, &coreParams.initialY)
+  /**
+ * @brief Initial Z after reset [m]
+ */
+  PARAM_ADD_CORE(PARAM_FLOAT, initialZ, &coreParams.initialZ)
+  /**
+ * @brief Initial Yaw after reset [rad]
```
jonasdn, line 647: "Initial yaw"

Issue #861, opened 2021-10-04
Overall commander architecture is complicated


jpreiss:
Many parts of the Crazyflie firmware can affect how the `setpoint_t` struct is filled in each stabilizer loop, and/or how radio commands are interpreted. I am wondering if it would be worth the effort to refactor.

I am aware of:

1. The `supervisor` (formerly `sitAw`) framework for detecting idle/flying/tumbled state.
2. Low-level setpoints preempting the high-level commander...
2. ... but reverting to "stabilize mode" and then either high-level or "declaring an emergency" if a setpoint is not recieived for too long.
3. The priority system for low-level setpoints.
   - Allows traditional R/C remotes to override CRTP commands. Not sure if used elsewhere.
4. The `notifySetpointsStop` meta-command to override/modify mechanism 2.
5. The high-level commander's internal notion of idle, flying, landing, and emergency stop (via wrapping `planner.c`).
6. The high-level commander's `groupMask` for broadcasted high-level commands.
7. The parameter system param `commander.enHighLevel`.
8. Possible interaction between the app layer and the radio link.
   - I do not know anything about the app layer, but from `crtp_commander_high_level.c` it appears that the app layer interacts with the rest of the firmware by "sending" CRTP packets.
9. The collision avoidance system.

I can see the appeal of implementing the supervisor outside the commander and only using sensor data and motor state. It is harder to introduce bugs. It reminds me of a [subsumption architecture](https://en.wikipedia.org/wiki/Subsumption_architecture).

Collision avoidance is also a "setpoint modifier" instead of a "setpoint creator", so it seems reasonable to keep it separate.

For everything else, I do not personally see much benefit to implementing them as independent subsystems. I understand that we have reached this design by many decentralized developers adding features over time. However, I feel that it is now hard to build a mental model of the firmware. For example, the prospect of making `notifySetpointsStop` a 100% reliable feature in all states (see [bug report in Crazyswarm](https://github.com/USC-ACTLab/crazyswarm/issues/464)) feels more difficult than it should.

My intuition would be to unify many of these factors into a state machine. Maintaining radio protocol compatibility (#849) might make this harder, but I think it is still possible. It would also be a good first step towards making the whole commander framework compilable and bindable for x86 (#602).

I am curious - how others feel about this?

whoenig:
I agree that this is a problem and needs a redesign. One big pain point is that it is really hard (or even impossible?) to switch high-level -> low-level safely.

[...]

PR #872, opened 2021-10-12
move relative trajectory shift computation into planner.c

jpreiss:
When executing a previously uploaded trajectory in relative position mode, we need some slightly bulky and non-obvious branching code to deal with the forward/reverse and regular/compressed polynomial options.

This code currently lives in `crtp_commander_high_level.c`. In my opinion, it would be more appropriate to move it into `planner.c` because that is the high-level wrapper in between polynomials and the user. This would leave `crtp_commander_high_level.c` only responsible for listening to the radio and managing trajectory memory.

Note that this changes the planner API but not the crtpHighLevelCommander API. To my understanding, only the latter is considered "public" w.r.t. the app layer.

This change will be useful for the bind-able top level commander state machine proposed in #861, but IMO is worth doing even if that idea is ultimately rejected.

I flight tested with regular polynomials but not compressed polynomials.


ntamas:
As the author of the compressed trajectory format it also looks good to me!

PR #881, opened 2021-11-04
deck_memory: Support secondary memory definition

jonasdn:
The primary driver here is support two firmwares for one deck, like the
AI deck with the GAP8 and the Nina module.

This makes it possible to set an id to a memory definition so you could
have:

```c
static const DeckMemDef_t memoryDefNina = {
  .write = write_nina,
  .read = read_nina,
  .properties = properties_nina,
  .supportsUpgrade = true,

  .requiredSize = ...,
  .requiredHash = ...,
};

static const DeckMemDef_t memoryDefGap8 = {
  .write = write_gap8,
  .read = read_gap8,
  .properties = properties_gap8,
  .supportsUpgrade = true,

  .requiredSize = ...,
  .requiredHash = ...,

  .id = "gap8"
};

static const DeckDriver aideck_deck = {
    .vid = 0xBC,
    .pid = 0x12,
    .name = "bcAI",

    .usedGpio = DECK_USING_IO_4,
    .usedPeriph = DECK_USING_UART1,

    .init = aideckInit,
    .test = aideckTest,

    .memoryDef = memoryDefNina,
    .memoryDefSecondary = memoryDefGap8,
};
```

And the names visible to the lib would be:
    "bcAI" and
    "bcAI:gap8"

### DIFF
--- a/src/deck/core/deck_memory.c
+++ b/src/deck/core/deck_memory.c
@@ -98,19 +98,53 @@ static uint8_t populateBitfield(const DeckMemDef_t* memDef) {
     return result;
 }
 
-static void populateDeckMemoryInfo(uint8_t buffer[], const uint8_t deckNr) {
+static void populateDeckMemoryInfoBuffer(const DeckMemDef_t *deckMemDef,
+                                         const char *name,
+                                         uint32_t baseAddress,
+                                         uint8_t buffer[])
+{
+    buffer[OFFS_BITFIELD] = populateBitfield(deckMemDef);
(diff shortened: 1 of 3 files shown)


knmcguire:
Seems like some checks are failing for this PR

jonasdn:
> Seems like some checks are failing for this PR

Yes,  this PR raises the length of the DECK_MEMORY_INFO_SIZE from 32 (0x20) to 64 (0x40).
Max number of decks is 4, this gives => 64 * 4 to read all deck-info. 64 * 4 => 256. And the readLen parameter of the read function-pointer in the memory system is a uint8_t, so a read of 256 is a read of 0.

PR #886, opened 2021-11-19
Python bindings for pptraj and planner modules

whoenig:
These bindings are used in the Crazyswarm for SIL simulation of the
high-level mode.

### DIFF
--- a/bindings/setup.py
+++ b/bindings/setup.py
@@ -10,7 +10,9 @@ include = [
 ]
 
 modules = [
-    # list firmware c-files here
+    "pptraj.c",
+    "pptraj_compressed.c",
+    "planner.c"
 ]
 fw_sources = [os.path.join(fw_dir, "src/modules/src", mod) for mod in modules]
 
@@ -20,6 +22,9 @@ cffirmware = Extension(
     sources=fw_sources + ["bin/cffirmware_wrap.c"],
     extra_compile_args=[
         "-O3",
+        # The following flags are also used for compiling the actual firmware
+        "-fno-strict-aliasing",
+        "-Wno-address-of-packed-member",
     ],
 )
 
--- a/test_python/test_planner.py
+++ b/test_python/test_planner.py
@@ -1,0 +1,22 @@
+#!/usr/bin/env python
+
+import numpy as np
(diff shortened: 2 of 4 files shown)


bindings/setup.py
```diff
@@ -20,6 +22,9 @@
     sources=fw_sources + ["bin/cffirmware_wrap.c"],
     extra_compile_args=[
         "-O3",
+        # The following flags are also used for compiling the actual firmware
+        "-fno-strict-aliasing",
+        "-Wno-address-of-packed-member",
```
jonasdn, line 27: I am not sure how all of this newfangled things you kids use work. But. Should we use the same strategy as as in our Makefile here?

In that case we might also need to add:

```
CFLAGS += -Wno-array-bounds
CFLAGS += -Wno-stringop-overread
CFLAGS += -Wno-stringop-overflow
```

Or do we cross that bridge when needed?

whoenig: My design goal was to use as few extra settings as possible and at most those that we use for the regular compilation (i.e., the Python should use a subset of the flags of the real firmware). Those are the two that I needed so far, in order to avoid compiler warnings. I'd suggest that we add the other three flags only when needed.

whoenig:
Any update here? There are more bindings to come.

test_python/test_planner.py
```diff
@@ -0,0 +1,22 @@
...
+    targetHeight = 1.0
+    targetYaw = 0
+    duration = 2
+
+    # Test
+    cffirmware.plan_takeoff(planner, pos, yaw, targetHeight, targetYaw, duration, 0)
+
+    # Assert
+    state = cffirmware.plan_current_goal(planner, duration)
+    assert np.allclose(np.array([0, 0, 1.0]), state.pos)
```
jonasdn, line 21: nit: could that `1.0` be `targetHeight` ?

jonasdn (approved):
Looks good to me!

Just small nit, fix or ignore at will!

test_python/test_planner.py
whoenig: Thanks - fixed!

PR #887, opened 2021-11-23
Low-level ESP32 bootloader SLIP communication protocol

gemenerik:
Low-level implementation of ESP32 ROM bootloader communication protocol. First stepping stone to wireless flashing of ESP32 from zip.

The ESP32 bootloader sends no unsolicited messages. The exchange function can be called from a higher level with a variation of instructions and data (to write). The exchange executes the following tasks:
1. Clear the UART buffer. Despite the ESP32 ROM bootloader otherwise not sending any unsolicited packets, it does send several responses upon successful synchronization, overflowing the UART buffer and potentially putting the SLIP packet decoding state machine in a wrong state. Clearing the UART buffer before sending a new SLIP packet avoids this.
2. Construct a SLIP packet out of 1) a struct containing header information (esp_slip_send_packet) and 2) a buffer containing to be sent data
3. Send this buffer in parts over UART with DMA
4. Receive the response from the ESP32 ROM bootloader, read out the status byte.
5. Return true for status byte indicating success, return false otherwise

### DIFF
--- a/src/drivers/esp32/interface/esp_slip.h
+++ b/src/drivers/esp32/interface/esp_slip.h
@@ -86,4 +86,4 @@ typedef struct
 *
 * @return true if ESP responds with a status byte indicating success.
 **/
-bool espSlipExchange(uint8_t *sendBuffer, esp_slip_receive_packet *receiverPacket, esp_slip_send_packet *senderPacket, coms_sendbuffer_t sendBufferFunction, coms_getDataWithTimeout_t getDataWithTimeout, uint32_t timeoutTicks);
+bool espSlipExchange(uint8_t *sendBuffer, esp_slip_receive_packet *receiverPacket, esp_slip_send_packet *senderPacket, coms_sendbuffer_t sendBufferFunction, coms_getDataWithTimeout_t getDataWithTimeout, uint32_t timeoutTicks, uint32_t txBufferSize);
--- a/src/drivers/esp32/interface/esp_slip.h
+++ b/src/drivers/esp32/interface/esp_slip.h
@@ -45,6 +45,9 @@
 #define CHANGE_BAUDRATE 0x0f
 #define SPI_ATTACH 0x0d
 
+typedef void (*coms_sendbuffer_t)(uint32_t size, uint8_t *data);
+typedef bool (*coms_getDataWithTimeout_t)(uint8_t *c, const uint32_t timeoutTicks);
+
 
 typedef struct
 {
--- a/src/drivers/esp32/interface/esp_slip.h
+++ b/src/drivers/esp32/interface/esp_slip.h
@@ -30,3 +30,6 @@
 
 #include <stdbool.h>
 #include <stdint.h>
+
+#define ESP_MTU 4000
+
--- a/src/drivers/esp32/interface/esp_slip.h
+++ b/src/drivers/esp32/interface/esp_slip.h
@@ -33,3 +33,15 @@
 
 #define ESP_MTU 4000
 
+/* Commands */
+#define DIR_CMD 0x00
+#define FLASH_BEGIN 0x02
+#define FLASH_DATA 0x03
+#define FLASH_END 0x04
+#define SYNC 0x08
+#define SPI_FLASH_MD5 0x13
+#define ERASE_FLASH 0xd0
+#define READ_FLASH 0xd2
+#define CHANGE_BAUDRATE 0x0f
+#define SPI_ATTACH 0x0d
+
--- a/src/drivers/esp32/interface/esp_slip.h
+++ b/src/drivers/esp32/interface/esp_slip.h
@@ -45,6 +45,8 @@
 #define CHANGE_BAUDRATE 0x0f
 #define SPI_ATTACH 0x0d
 
+#define SLIP_START_STOP_BYTE 0xc0
+
 typedef void (*coms_sendbuffer_t)(uint32_t size, uint8_t *data);
 typedef bool (*coms_getDataWithTimeout_t)(uint8_t *c, const uint32_t timeoutTicks);
 
--- a/src/drivers/esp32/interface/esp_slip.h
+++ b/src/drivers/esp32/interface/esp_slip.h
@@ -89,4 +89,5 @@ typedef struct
 *
 * @return true if ESP responds with a status byte indicating success.
 **/
-bool espSlipExchange(uint8_t *sendBuffer, espSlipReceivePacket_t *receiverPacket, espSlipSendPacket_t *senderPacket, espSlipSendBuffer_t sendBufferFunction, espSlipGetDataWithTimeout_t getDataWithTimeout, uint32_t timeoutTicks);
+bool espSlipExchange(uint8_t *sendBuffer, espSlipReceivePacket_t *receiverPacket, espSlipSendPacket_t *senderPacket, espSlipSendBuffer_t sendBufferFunction,
+                     espSlipGetDataWithTimeout_t getDataWithTimeout, uint32_t timeoutTicks);
--- a/src/drivers/esp32/interface/esp_slip.h
+++ b/src/drivers/esp32/interface/esp_slip.h
@@ -32,7 +32,7 @@
 #include <stdint.h>
 
 #define ESP_MTU 4000
-#define TX_BUFFER_SIZE 128
+#define ESP_SLIP_TX_BUFFER_SIZE 128
 
 /* Commands */
 #define DIR_CMD 0x00
--- a/src/drivers/esp32/interface/esp_slip.h
+++ b/src/drivers/esp32/interface/esp_slip.h
@@ -1,0 +1,32 @@
+/**
+ * ,---------,       ____  _ __
+ * |  ,-^-,  |      / __ )(_) /_______________ _____  ___
+ * | (  O  ) |     / __  / / __/ ___/ ___/ __ `/_  / / _ \
+ * | / ,--´  |    / /_/ / / /_/ /__/ /  / /_/ / / /_/  __/
+ *    +------`   /_____/_/\__/\___/_/   \__,_/ /___/\___/
+ *
+ * Crazyflie control firmware
+ *
+ * Copyright (C) 2021 Bitcraze AB
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, in version 3.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ * @file esp_slip.h
+ * Protocol for assembling, sending, receiving and decoding SLIP packets to/from the ESP32 ROM bootloader
+ *  
+ */
+
+#pragma once
+
+#include <stdbool.h>
+#include <stdint.h>
--- a/src/drivers/esp32/interface/esp_slip.h
+++ b/src/drivers/esp32/interface/esp_slip.h
@@ -47,8 +47,8 @@
 
 #define SLIP_START_STOP_BYTE 0xc0
 
-typedef void (*coms_sendbuffer_t)(uint32_t size, uint8_t *data);
-typedef bool (*coms_getDataWithTimeout_t)(uint8_t *c, const uint32_t timeoutTicks);
+typedef void (*espSlipSendBuffer_t)(uint32_t size, uint8_t *data);
+typedef bool (*espSlipGetDataWithTimeout_t)(uint8_t *c, const uint32_t timeoutTicks);
 
 typedef enum
 {
@@ -63,7 +63,7 @@ typedef struct
   uint8_t command;
   uint16_t dataSize;
   uint32_t checksum;
-} __attribute__((packed)) esp_slip_send_packet;
+} __attribute__((packed)) espSlipSendPacket_t;
 
 typedef struct
 {
@@ -74,7 +74,7 @@ typedef struct
   uint8_t data[256];
   uint8_t status;
   uint8_t error;
-} __attribute__((packed)) esp_slip_receive_packet;
+} __attribute__((packed)) espSlipReceivePacket_t;
 
 /**
 * @brief Called to send a SLIP packet to the ESP, and receive the response.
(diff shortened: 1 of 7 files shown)


src/drivers/esp32/interface/esp_slip.h
```diff
@@ -0,0 +1,89 @@
...
+  SLIP_ERROR
+} slipDecoderStatus_t;
+
+typedef struct
+{
+  uint8_t direction;
+  uint8_t command;
+  uint16_t dataSize;
+  uint32_t checksum;
+} __attribute__((packed)) esp_slip_send_packet;
```
krichardsson, line 64: Snake case should be changed to Camel case, maybe espSlipSendPacket_t

src/drivers/esp32/interface/esp_slip.h
```diff
@@ -0,0 +1,89 @@
...
+typedef struct
+{
+  uint8_t direction;
+  uint8_t command;
+  uint16_t dataSize;
+  uint32_t value; // only for READ_REG command
+  uint8_t data[256];
+  uint8_t status;
+  uint8_t error;
+} __attribute__((packed)) esp_slip_receive_packet;
```
krichardsson, line 75: Snake case should be changed to camel case

src/drivers/esp32/interface/esp_slip.h
```diff
@@ -0,0 +1,89 @@
...
+#define FLASH_DATA 0x03
+#define FLASH_END 0x04
+#define SYNC 0x08
+#define SPI_FLASH_MD5 0x13
+#define ERASE_FLASH 0xd0
+#define READ_FLASH 0xd2
+#define CHANGE_BAUDRATE 0x0f
+#define SPI_ATTACH 0x0d
+
+typedef void (*coms_sendbuffer_t)(uint32_t size, uint8_t *data);
```
krichardsson, line 48: coms_sendbuffer_t and coms_getDataWithTimeout_t should be prefixed with espSlip as it is public. Maybe remove the coms part?

src/drivers/esp32/src/esp_slip.c
```diff
@@ -0,0 +1,315 @@
...
+
+static void sendSlipPacket(uint32_t size, uint8_t *data, coms_sendbuffer_t sendBufferFn)
+{
+  uint32_t i;
+  static uint8_t dmaSendBuffer[UART2_DMA_BUFFER_SIZE];
+  uint32_t dmaSendSize = 0;
+
+  for (i = 0; i < size; i++)
+  {
+    if ((data[i] == 0xC0 && i != 0 && i != size - 1) || (data[i] == 0xDB && i != 0 && i != size - 1))
```
krichardsson, line 76: Consider creating constants for magic numbers, especially if used multiple times, for instance 0xC0

src/drivers/esp32/src/esp_slip.c
```diff
@@ -0,0 +1,315 @@
...
+      dmaSendSize = 0;
+    }
+  }
+  if (dmaSendSize)
+  {
+    sendBufferFn(dmaSendSize, &dmaSendBuffer[0]);
+  }
+}
+
+static slipDecoderStatus_t decodeSlipPacket(uint8_t c, esp_slip_receive_packet *receiverPacket, esp_slip_send_packet *senderPacket)
```
krichardsson, line 102: This is a pretty long function, can it maybe be split up somehow?

src/drivers/esp32/src/esp_slip.c
```diff
@@ -0,0 +1,315 @@
...
+    sendBuffer[5] = 0x00;
+    sendBuffer[6] = 0x00;
+    sendBuffer[7] = 0x00;
+    sendBuffer[8] = 0x00;
+  }
+
+  sendBuffer[9 + senderPacket->dataSize] = 0xC0;
+}
+
+static void clearUart2Buffer(coms_getDataWithTimeout_t getDataWithTimeout)
```
krichardsson, line 296: Suggest to rename to avoid the use of "uart2" in the name, maybe flushTxBuffer() or similar. The design with fcn pointers for communication allows for pretty much any mode of communication, not only UART.

src/drivers/esp32/src/esp_slip.c
```diff
@@ -0,0 +1,315 @@
...
+  {
+    checksum ^= sendBuffer[9 + 16 + i];
+  }
+  return checksum;
+}
+
+static void sendSlipPacket(uint32_t size, uint8_t *data, coms_sendbuffer_t sendBufferFn)
+{
+  uint32_t i;
+  static uint8_t dmaSendBuffer[UART2_DMA_BUFFER_SIZE];
```
krichardsson, line 71: It would be nice to get rid of references to DMA and UART2. The design of using fcn pointers for communication should remove as much as possible of the knowledge of the underlying communication layer. Ideally this file should not have a dependency to DMA or UART2 (the #include earlier in the file). The buffer size must obviously be set in one way or the other but it maybe it would be possible to do it without the dependency?

gemenerik: I will pass the buffer size from level up, considering that will have a dependency to UART2 to get the read/write functions. I'll remove references to UART and DMA from variable names

src/drivers/esp32/src/esp_slip.c
gemenerik: Perhaps using a ternary operators for some of the deepest nested if statements can actually improve readability?

src/drivers/esp32/src/esp_slip.c
gemenerik: The above fix doesn't allow for a static dmaSendBuffer, so after discussion we decided to define a buffer size here and compare it to the UART2 DMA buffer size in the higher level.

src/drivers/esp32/src/esp_slip.c
```diff
@@ -33,3 +33,12 @@
 
 #define ESP_OVERHEAD_LEN 8
 
+static uint8_t generateChecksum(uint8_t *sendBuffer, esp_slip_send_packet *senderPacket)
+{
+  uint8_t checksum = 0xEF; // seed
+  for (int i = 0; i < senderPacket->dataSize - 16; i++)
+  {
+    checksum ^= sendBuffer[9 + 16 + i];
```
jonasdn, line 62: Is this magic explained somewhere? Or should be done here?

src/drivers/esp32/src/esp_slip.c
```diff
@@ -42,3 +44,32 @@ static uint8_t generateChecksum(uint8_t *sendBuffer, esp_slip_send_packet *sende
   }
   return checksum;
 }
+
+static void assembleBuffer(uint8_t *sendBuffer, esp_slip_send_packet *senderPacket)
+{
+  sendSize = senderPacket->dataSize + ESP_OVERHEAD_LEN + 2;
```
jonasdn, line 273: Why + 2?

src/drivers/esp32/src/esp_slip.c
```diff
@@ -42,3 +44,32 @@ static uint8_t generateChecksum(uint8_t *sendBuffer, esp_slip_send_packet *sende
...
+
+  sendBuffer[0] = 0xC0;
+  sendBuffer[1] = DIR_CMD;
+  sendBuffer[2] = senderPacket->command;
+  sendBuffer[3] = (uint8_t)((senderPacket->dataSize >> 0) & 0x000000FF);
+  sendBuffer[4] = (uint8_t)((senderPacket->dataSize >> 8) & 0x000000FF);
+
+  if (senderPacket->command == FLASH_DATA) // or MEM_DATA
```
jonasdn, line 281: This comment confuses me more than it enlightens me, what does it mean?

src/drivers/esp32/src/esp_slip.c
```diff
@@ -106,3 +106,14 @@ static void assembleBuffer(uint8_t *sendBuffer, esp_slip_send_packet *senderPack
...
+
+static void clearUart2Buffer(coms_getDataWithTimeout_t getDataWithTimeout)
+{
+  uint8_t c;
+  bool success = true;
+  while (success)
+  {
+    success = getDataWithTimeout(&c, 1);
+  }
+  return;
```
jonasdn, line 308: For me this is a do-while loop, but I've learnt that I am quite alone in liking them, so take this with a grain of salt :)

```
do {
  success = getDataWithTimeout(&c, 1);
} while(success);
```

src/drivers/esp32/src/esp_slip.c
```diff
@@ -33,6 +33,26 @@
...
+  error,
+} ESPblReceiveState;
+static ESPblReceiveState espblReceiveState = receiveStart;
+
+static uint8_t slipSize[2];
+static uint8_t slipValue[4];
+static uint8_t slipDataIndex = 0;
+static uint8_t slipSizeIndex = 0;
+static uint8_t slipValueIndex = 0;
+static bool inEscapeSequence = false;
```
jonasdn, line 53: nit: this is not needed, all static variables are 0 by default.

src/drivers/esp32/src/esp_slip.c
```diff
@@ -78,6 +98,144 @@ static void sendSlipPacket(uint32_t size, uint8_t *data, coms_sendbuffer_t sendB
...
+        decoderStatus = SLIP_ERROR;
+      }
+    }
+    slipSizeIndex++;
+    break;
+  case receiveValue: // only used for READ_REG
+    slipValue[slipValueIndex] = c;
+    if (slipValueIndex == 3)
+    {
+      receiverPacket->value = ((uint32_t)slipValue[0] + ((uint32_t)slipValue[1] << 8) + ((uint32_t)slipValue[2] << 16) + ((uint32_t)slipValue[3] << 24));
```
jonasdn, line 161: This is begging for a comment, it might be obvious to Rik today, but not to Rik tomorrow, or Jonas today :)

```
//
// Explaination of this subscripted bitshifted genious code
//
receiverPacket->value =  ((uint32_t) slipValue[0] + ((uint32_t) slipValue[1] << 8);
receiverPacket->value += ((uint32_t) slipValue[2] << 16) + ((uint32_t) slipValue[3] << 24));
```

src/drivers/esp32/src/esp_slip.c
```diff
@@ -236,6 +236,34 @@ static slipDecoderStatus_t decodeSlipPacket(uint8_t c, esp_slip_receive_packet *
   return decoderStatus;
 }
 
+static bool receivePacket(esp_slip_receive_packet *receiverPacket, esp_slip_send_packet *senderPacket, coms_getDataWithTimeout_t getDataWithTimeout, uint32_t timeoutTicks)
```
jonasdn: perhaps split this definition over some lines?

```c
static bool receivePacket(esp_slip_receive_packet *receiverPacket, esp_slip_send_packet *senderPacket, 
                          coms_getDataWithTimeout_t getDataWithTimeout, uint32_t timeoutTicks)
{
```

src/drivers/esp32/interface/esp_slip.h
```diff
@@ -74,3 +74,16 @@ typedef struct
...
+*
+* @param *sendBuffer Pointer to the to be sent buffer
+* @param *receiverPacket Pointer to receiver packet struct, which will be filled with the response
+* @param *senderPacket Pointer to sender packet struct, which will be used to fill the send buffer header
+* @param sendBufferFunction Function pointer to the function that sends the buffer (split into pages) to the ESP
+* @param getDataWithTimeoutFunction Function pointer to the function that receives data (byte by byte) from the ESP
+* @param timeoutTicks Number of ticks to wait for a response from the ESP
+*
+* @return true if ESP responds with a status byte indicating success.
+**/
```
jonasdn, line 91: Nice!

src/drivers/esp32/interface/esp_slip.h
```diff
@@ -74,3 +74,16 @@ typedef struct
...
+bool espSlipExchange(uint8_t *sendBuffer, esp_slip_receive_packet *receiverPacket, esp_slip_send_packet *senderPacket, coms_sendbuffer_t sendBufferFunction, coms_getDataWithTimeout_t getDataWithTimeout, uint32_t timeoutTicks);
```
jonasdn: Missing newline at end-of-file! \Ö/

src/drivers/esp32/src/esp_slip.c
```diff
@@ -72,11 +72,11 @@ static void sendSlipPacket(uint32_t size, uint8_t *data, coms_sendbuffer_t sendB
 
   for (i = 0; i < size; i++)
   {
-    if ((data[i] == 0xC0 && i != 0 && i != size - 1) || (data[i] == 0xDB && i != 0 && i != size - 1))
+    if ((data[i] == SLIP_START_STOP_BYTE && i != 0 && i != size - 1) || (data[i] == 0xDB && i != 0 && i != size - 1))
     {
       for (int j = 0; j < 2; j++)
       {
-        j == 0 ? (dmaSendBuffer[dmaSendSize] = 0xDB) : data[i] == 0xC0 ? (dmaSendBuffer[dmaSendSize] = 0xDC)
+        j == 0 ? (dmaSendBuffer[dmaSendSize] = 0xDB) : data[i] == SLIP_START_STOP_BYTE ? (dmaSendBuffer[dmaSendSize] = 0xDC)
```
jonasdn: I really would prefer if-statements here, ternary are a hell to parse while reading code. And code is read 1000x more often than it is written, at least. Always focus on readability when in doubt.

jonasdn (changes requested):
In general really really nice and clean code! Well done @gemenerik !

src/drivers/esp32/src/esp_slip.c
gemenerik: It's for the start and end bytes. I'll add a comment to explain it.

src/drivers/esp32/src/esp_slip.c
gemenerik: It refers to another "command" that requires a checksum to be sent to the ESP, but considering we will not use it, I will remove the comment

src/drivers/esp32/src/esp_slip.c
gemenerik: Fair enough, I guess this is a classic example of where it can be implemented.

src/drivers/esp32/src/esp_slip.c
gemenerik: I like nits, makes me a better programmer!

src/drivers/esp32/src/esp_slip.c
gemenerik: I added a comment to explain it's for the start and end byte of the SLIP packet

src/drivers/esp32/src/esp_slip.c
gemenerik: I added a comment to explain the indeed quite magic function

PR #888, opened 2021-11-24
Enable stack overflow check

krichardsson:
Currently the stack overflow check provided by FreeRTOS is enabled in DEBUG compiles only. This pull request also enables it in "normal" compiles. 

The overhead for this check is very low, one if statement per context switch, and I believe it is better to catch this type of problem as early as possible. 

Note: This overflow check is not fool-proof, it only samples the stack size at context switch


ataffanel:
Looks good!

PR #892, opened 2021-11-30
Persistent param

jonasdn:


src/modules/interface/param.h
```diff
@@ -168,52 +91,69 @@ struct param_s {
...
-  { .type = TYPE, .name = #NAME, .address = (void*)(ADDRESS), .callback = (void *)CALLBACK, },
+    PARAM_ADD_FULL(TYPE, NAME, ADDRESS, CALLBACK, 0)
 
 #define PARAM_ADD_CORE(TYPE, NAME, ADDRESS) \
   PARAM_ADD(TYPE | PARAM_CORE, NAME, ADDRESS)
 
 #define PARAM_ADD_CORE_WITH_CALLBACK(TYPE, NAME, ADDRESS, CALLBACK) \
   PARAM_ADD_WITH_CALLBACK(TYPE | PARAM_CORE, NAME, ADDRESS, CALLBACK)
 
+#define PARAM_ADD_PERSISTENT(TYPE, NAME, ADDRESS, DEFAULT_GETTER) \
```
tobbeanton, line 126: As previously discussed should we remove this macro and only use PARAM_PERSISTENT flag. A macro might be simpler but there would have to be multiple flavors. If however the persistent param should always be core then a macro would be preferred to enforce this.

src/modules/src/param_task.c
```diff
@@ -0,0 +1,138 @@
...
+  //Start the param task
+  STATIC_MEM_TASK_CREATE(paramTask, paramTask, PARAM_TASK_NAME, NULL, PARAM_TASK_PRI);
+
+  isInit = true;
+}
+
+bool paramTest(void)
+{
+
+//  p.data[0] = 0x00;
```
tobbeanton: Maybe we should remove the test code that is not used?

tobbeanton (commented):
Should we add documentation in this PR or make a new one for this because better documentation is needed. Also since we removed support for param V1 this should be updated in the documentation.

PR #899, opened 2021-12-09
docs: Add pwm to thrust investigation

jonasdn:


whoenig:
This is a really old measurement. Would be nice to get this updated using our new system-ID board. IMO the old transfer function is not a good model, because it doesn't take the battery state as input. Perhaps in a future change? Ultimately, I'd like to see the data that we provided in the Crazyswarm2 paper (and reproduced earlier this year with Tobias).

knmcguire:
The link to this doc is missing in the _data/menu.yml

knmcguire:
Yes absolutely we need to update it, but I think this PR is currently to move around documents that were still lingering from the wiki to a more suitable place. The 'master' PR is here: https://github.com/bitcraze/bitcraze-website/pull/106

Since moving docs from website to repos is a difficult process involving multiple repositories, I would be in favor of accepting this and making a separate issue to update to docs to the latest measurements if that is okay, just to make sure that we won't lose this doc in PR limbo.

knmcguire (changes requested):
need a link to this page in _data/menu.yml

whoenig:
Agreed with moving forward with this. Just wanted to point out that an update is needed later on.

knmcguire (approved):

knmcguire:
I've just added the menu.yml to this branch, so I'll merge it once the check has been completed.

PR #903, opened 2021-12-14
Push-based high-level commander

jpreiss:
Implements push-based high-level commander as discussed in #861.

Summary of changes (HLcmd is short for high-level commander):

1. Adds a new priority level for HLcmd, lower than all but DISABLE.
2. Stabilizer loop steps HLcmd and pushes setpoints on its behalf.
   (Initially I tried having a separate task, but it was messier due to needing a lock on the state estimate struct.)
3. Introduces a new planner state to distinguish between the conditions:
   - HLcmd has been preempted
   - HLcmd has finished landing and wants motors off
   
   This is necessary (as far as I can tell) to preserve the correct leveling-off and emergency stop behaviors on timeout, while also allowing the high-level commander to immediately cut the motors after landing.
4. Switching from low-level to high-level is accomplished by modifying the priority of the current setpoint sitting in the queue. This means we no longer need the `remainValidMillisecs` argument on the `notifySetpointsStop` command.
5. The `commander.enHighLevel` param is no longer needed.

The priority switching mechanism still needs to "know about" the high-level commander (i.e. `commander.c` still includes `crtp_commander_high_level.h`) to switch it into the disabled state when it is preempted. I couldn't figure out a way around this.

Although the net lines of code is increased, notice the simplification of `commander.c`.

Flight tested in 2 scenarios:

```
takeoff
cmdFullState
goTo
cmdFullState
goTo
land
```
and

```
takeoff
cmdFullState
<wait for leveling and emergency stop>
```

### DIFF
--- a/src/modules/src/commander.c
+++ b/src/modules/src/commander.c
@@ -41,7 +41,7 @@ static bool isInit;
 // Static structs are zero-initialized, so nullSetpoint corresponds to
 // modeDisable for all stab_mode_t members and zero for all physical values.
 // In other words, the controller should cut power upon recieving it.
-const setpoint_t nullSetpoint;
+const static setpoint_t nullSetpoint;
 static state_t lastState;
 const static int priorityDisable = COMMANDER_PRIORITY_DISABLE;
 
--- a/src/modules/src/commander.c
+++ b/src/modules/src/commander.c
@@ -46,6 +46,7 @@ static state_t lastState;
 const static int priorityDisable = COMMANDER_PRIORITY_DISABLE;
 
 static uint32_t lastUpdate;
+static bool enableHighLevel = false;
 
 static QueueHandle_t setpointQueue;
 STATIC_MEM_QUEUE_ALLOC(setpointQueue, 1, sizeof(setpoint_t));
@@ -143,3 +144,30 @@ int commanderGetActivePriority(void)
 
   return priority;
 }
+
+/**
+ *
+ * The high level commander handles the setpoints from within the firmware
+ * based on a predefined trajectory. This was merged as part of the
+ * [Crazyswarm](%https://crazyswarm.readthedocs.io/en/latest/) project of the
+ * [USC ACT lab](%https://act.usc.edu/) (see this
+ * [blogpost](%https://www.bitcraze.io/2018/02/merging-crazyswarm-functionality-into-the-official-crazyflie-firmware/)).
+ * The high-level commander uses a planner to generate smooth trajectories
+ * based on actions like ‘take off’, ‘go to’ or ‘land’ with 7th order
+ * polynomials. The planner generates a group of setpoints, which will be
+ * handled by the High level commander and send one by one to the commander
+ * framework.
+ *
+ * It is also possible to upload your own custom trajectory to the memory of
+ * the Crazyflie, which you can try out with the script
+ * `examples/autonomous_sequence_high_level of.py` in the Crazyflie python
+ * library repository.
+ */
+PARAM_GROUP_START(commander)
+
+/**
+ *  @brief Enable high level commander
+ */
+PARAM_ADD_CORE(PARAM_UINT8, enHighLevel, &enableHighLevel)
+
+PARAM_GROUP_STOP(commander)
--- a/src/modules/src/commander.c
+++ b/src/modules/src/commander.c
@@ -41,13 +41,11 @@ static bool isInit;
 // Static structs are zero-initialized, so nullSetpoint corresponds to
 // modeDisable for all stab_mode_t members and zero for all physical values.
 // In other words, the controller should cut power upon recieving it.
-const static setpoint_t nullSetpoint;
-static setpoint_t tempSetpoint;
+const setpoint_t nullSetpoint;
 static state_t lastState;
 const static int priorityDisable = COMMANDER_PRIORITY_DISABLE;
 
 static uint32_t lastUpdate;
-static bool enableHighLevel = false;
 
 static QueueHandle_t setpointQueue;
 STATIC_MEM_QUEUE_ALLOC(setpointQueue, 1, sizeof(setpoint_t));
@@ -84,23 +82,19 @@ void commanderSetSetpoint(setpoint_t *setpoint, int priority)
     // This is a potential race but without effect on functionality
     xQueueOverwrite(setpointQueue, setpoint);
     xQueueOverwrite(priorityQueue, &priority);
-    // Send the high-level planner to idle so it will forget its current state
-    // and start over if we switch from low-level to high-level in the future.
-    crtpCommanderHighLevelStop();
+    if (priority > COMMANDER_PRIORITY_HIGHLEVEL) {
+      // Disable the high-level planner so it will forget its current state and
+      // start over if we switch from low-level to high-level in the future.
+      crtpCommanderHighLevelDisable();
+    }
   }
 }
 
-void commanderNotifySetpointsStop(int remainValidMillisecs)
+void commanderRelaxPriority()
 {
-  uint32_t currentTime = xTaskGetTickCount();
-  int timeSetback = MIN(
-    COMMANDER_WDT_TIMEOUT_SHUTDOWN - M2T(remainValidMillisecs),
-    currentTime
-  );
-  xQueuePeek(setpointQueue, &tempSetpoint, 0);
-  tempSetpoint.timestamp = currentTime - timeSetback;
-  xQueueOverwrite(setpointQueue, &tempSetpoint);
   crtpCommanderHighLevelTellState(&lastState);
+  int priority = COMMANDER_PRIORITY_LOWEST;
+  xQueueOverwrite(priorityQueue, &priority);
 }
 
 void commanderGetSetpoint(setpoint_t *setpoint, const state_t *state)
@@ -110,12 +104,7 @@ void commanderGetSetpoint(setpoint_t *setpoint, const state_t *state)
   uint32_t currentTime = xTaskGetTickCount();
 
   if ((currentTime - setpoint->timestamp) > COMMANDER_WDT_TIMEOUT_SHUTDOWN) {
-    if (enableHighLevel) {
-      crtpCommanderHighLevelGetSetpoint(setpoint, state);
-    }
-    if (!enableHighLevel || crtpCommanderHighLevelIsStopped()) {
-      memcpy(setpoint, &nullSetpoint, sizeof(nullSetpoint));
-    }
+    memcpy(setpoint, &nullSetpoint, sizeof(nullSetpoint));
   } else if ((currentTime - setpoint->timestamp) > COMMANDER_WDT_TIMEOUT_STABILIZE) {
     xQueueOverwrite(priorityQueue, &priorityDisable);
     // Leveling ...
@@ -154,30 +143,3 @@ int commanderGetActivePriority(void)
 
   return priority;
 }
-
-/**
- *
- * The high level commander handles the setpoints from within the firmware
- * based on a predefined trajectory. This was merged as part of the
- * [Crazyswarm](%https://crazyswarm.readthedocs.io/en/latest/) project of the
- * [USC ACT lab](%https://act.usc.edu/) (see this
- * [blogpost](%https://www.bitcraze.io/2018/02/merging-crazyswarm-functionality-into-the-official-crazyflie-firmware/)).
- * The high-level commander uses a planner to generate smooth trajectories
- * based on actions like ‘take off’, ‘go to’ or ‘land’ with 7th order
- * polynomials. The planner generates a group of setpoints, which will be
- * handled by the High level commander and send one by one to the commander
- * framework.
- *
- * It is also possible to upload your own custom trajectory to the memory of
- * the Crazyflie, which you can try out with the script
- * `examples/autonomous_sequence_high_level of.py` in the Crazyflie python
- * library repository.
- */
-PARAM_GROUP_START(commander)
-
-/**
- *  @brief Enable high level commander
- */
-PARAM_ADD_CORE(PARAM_UINT8, enHighLevel, &enableHighLevel)
-
-PARAM_GROUP_STOP(commander)
--- a/src/modules/src/crtp_commander_high_level.c
+++ b/src/modules/src/crtp_commander_high_level.c
@@ -298,8 +298,12 @@ int crtpCommanderHighLevelDisable()
   return 0;
 }
 
-bool crtpCommanderHighLevelGetSetpoint(setpoint_t* setpoint, const state_t *state)
+bool crtpCommanderHighLevelGetSetpoint(setpoint_t* setpoint, const state_t *state, uint32_t tick)
 {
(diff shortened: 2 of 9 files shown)


jpreiss:
also tested the scenario
```
takeoff
goTo with long duration
cmdFullState preempt (before goTo is done)
```
and verified that, once the `cmdFullState` packets cease, we get leveling followed by emergency stop (vs. jumping to the end of the `goTo` trajectory).

src/modules/src/commander.c
```diff
@@ -154,30 +143,3 @@ int commanderGetActivePriority(void)
...
- * library repository.
- */
-PARAM_GROUP_START(commander)
-
-/**
- *  @brief Enable high level commander
- */
-PARAM_ADD_CORE(PARAM_UINT8, enHighLevel, &enableHighLevel)
-
-PARAM_GROUP_STOP(commander)
```
jonasdn, line 183: Removing this would break a lot of code out there, right? Including some examples

knmcguire: yes indeed, in the cflib, cfclient and also some demo applications

src/modules/src/planner.c
```diff
@@ -88,6 +88,16 @@ bool plan_is_stopped(struct planner *p)
 	return p->state == TRAJECTORY_STATE_IDLE;
 }
 
+void plan_disable(struct planner *p)
```
knmcguire, line 91: So now that we have a plan_disable, is the plan_stop necessary in the grander scheme of things? We do have a s[topping function in the crtp_commander_generic](https://github.com/bitcraze/crazyflie-firmware/blob/56e357e0f4a3f01e04b0849214ef3628292f76d2/src/modules/src/crtp_commander_generic.c#L84). 

Isn't it better to put plan_disable() in the stop function of the generic commander, and use the more general stop command instead. Just to prevent the same functionality in multiple places.

src/modules/src/stabilizer.c
```diff
@@ -256,6 +259,12 @@ static void stabilizerTask(void* param)
       stateEstimator(&state, tick);
       compressState();
 
+      if (RATE_DO_EXECUTE(RATE_HL_COMMANDER, tick)) {
```
knmcguire, line 262: Currently I'm wondering if this is the right place to handle the planner and the execution rate of the high level commander. The position and attitude rate are handled within the controllers so perhaps this part should be in commander.c ? Just thinking out loud how to improve the consistency of the code

jpreiss: OK, but wouldn't it go in crtp_commander_high_level.c?

In the long term I think we might want to move the rate-skipping code in stabilizer.c, because as we make more modules bind-able it would not make sense to include the rate limiting within the bound "library" part. But that is orthogonal to this PR.

src/modules/src/planner.c
jpreiss: I believe we still need both the stopped and disabled states in the `planner.c` state machine due to reason 3 in my original PR comment. However, with this PR `crtpCommanderHighLevelStop` is no longer called by other parts of the firmware. The only place it might still be used is due to radio commands. I am not sure but it seems like we could deprecate and re-route the high-level stop radio command; I can't see why it needs to be different from the low-level one.

src/modules/src/commander.c
jpreiss: Sorry about that! In Crazyswarm missing parameters only cause a warning. I didn't realize it could actually break code.

I will restore the old param and comment. All `crazyflie-firmware` maintainers should be able to push to this branch, so please edit the docstring as you see fit.

src/modules/src/stabilizer.c
jpreiss: moved in 65f3a41

src/modules/src/commander.c
jpreiss: restored in 4109e90

whoenig (approved):
Very well done. I also love the code documentation surrounding it. Originally, I thought we would need another FreeRTOS task for pushing the HL commands, but I think the current solution is actually nicer.

src/modules/src/stabilizer.c
knmcguire: yeah indeed, let's keep it in mind for future discussions!

src/modules/src/planner.c
knmcguire: Just mind that crtpCommanderHighLevelStop are being used in the the [app api examples](https://github.com/bitcraze/crazyflie-firmware/blob/master/app_api/src/app_main.c) and [the app implementation of the swarm_demo](https://github.com/bitcraze/crazyflie-firmware/blob/master/examples/demos/swarm_demo/app.c)

knmcguire: so we just have to make sure to replace those. And the CI would catch it as well as we check all the app layer examples and api

src/modules/src/crtp_commander_high_level.c
```diff
@@ -791,6 +821,10 @@ bool crtpCommanderHighLevelIsTrajectoryFinished() {
   return plan_is_finished(&planner, t);
 }
 
+LOG_GROUP_START(hlCommander)
+LOG_ADD(LOG_UINT8, pstate, &planner.state)
+LOG_GROUP_STOP(hlCommander)
```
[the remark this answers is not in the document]
jpreiss: I used it for debugging but I doubt it will be a hot spot in the future. I'm fine with removing it.

src/modules/src/planner.c
jpreiss: @knmcguire in the swarm demo example, I believe the `crtpCommanderHighLevelStop` is not needed...

It is used here after landing: https://github.com/bitcraze/crazyflie-firmware/blob/250a9e552d793895351898f75d1529d9c12ed148/examples/demos/swarm_demo/app.c#L395-L401 but this same check and state transition will happen automatically within `planner.c`:
https://github.com/bitcraze/crazyflie-firmware/blob/250a9e552d793895351898f75d1529d9c12ed148/src/modules/src/planner.c#L93-L97

(IIRC, this automatic state transition is the only reason we have a `land` command at all - otherwise we could just use `goTo` followed by a manual stop.)

It is used here when a crash is detected: https://github.com/bitcraze/crazyflie-firmware/blob/250a9e552d793895351898f75d1529d9c12ed148/examples/demos/swarm_demo/app.c#L425-L427 but the supervisor framework should be overriding the high-level commander anyway: https://github.com/bitcraze/crazyflie-firmware/blob/250a9e552d793895351898f75d1529d9c12ed148/src/modules/src/supervisor.c#L115-L122

I am not sure what's going on in the first link - it looks like it's just calling every API function in arbitrary order.

src/modules/src/crtp_commander_high_level.c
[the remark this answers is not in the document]
knmcguire: It's not marked as an LOG_ADD_CORE, which means that it's not used in our examples or cfclient, so it's probably safe to remove.

src/modules/src/planner.c
knmcguire: perhaps @krichardsson could take a look at the demo example to make sure it's not broken when we remove it. And then if in the same PR we also remove the app_api one as well, then it should be all good for the checks

src/modules/src/crtp_commander_high_level.c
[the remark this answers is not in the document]
jpreiss: removed in 74d3149.

jpreiss:
Marked as ready for review. The only outstanding comments are related to removing `crtpCommanderHighLevelStop`. If I understand the comments correctly, we are OK with deferring that to a separate PR.

Since this is a core change, it would be nice to get some other flight tests to corroborate mine. @jonasdn do you have the ability to do a flight test with low-to-high switching using the Bitcraze Python API?

src/modules/src/planner.c
krichardsson: I think we can remove it from the swarm demo.

Since crtpCommanderHighLevelStop() is part of the public API we should not remove it completely. I assume the correct way is to NOP it and make it deprecated. I think it should be kept in app_api.c for a while, whith a comment that it has been deprecated.

jonasdn: Could we mark ut with __attribute__(deprecated)__?

knmcguire:
Hi! @jonasdn is not able to come to the office for a bit so I did some flight testing with the following content of run_sequence:

  ```
    commander = cf.high_level_commander
    commander.takeoff(1.0, 2.0)
    time.sleep(3.0)
    cf.commander.send_position_setpoint(0,0.5,1,0)
    time.sleep(1)
    cf.commander.send_position_setpoint(0,0.0,1,0)
    time.sleep(1)
    commander.land(0.0, 2.0)
    time.sleep(2)
    commander.stop()
```

I guess the last one line doesn't do anything right now.

Do I need to also try out some rpyt setpoints in combination too @jpreiss ?

https://user-images.githubusercontent.com/11061260/146773549-53578f0a-6c2d-441e-91b0-69fa16ffcb97.mp4

knmcguire:
rypt from script is a bit difficult without an continuous controller input :sweat_smile: but at least the switch between position setpoint and high level command seems to do the trick. But will that be enough to know if this is all okay?

jpreiss:
Sorry I've never used the Bitcraze API so I have some questions...

In this code:

    cf.commander.send_position_setpoint(0,0.5,1,0)
    time.sleep(1)

will the position setpoint be sent continuously over the 1 second sleep, or just once?

In this code:

    cf.commander.send_position_setpoint(0,0.0,1,0)
    time.sleep(1)
    commander.land(0.0, 2.0)

there is no `notifySetpointsStop` packet sent, so I would expect to see the CF to behave dictated by:
https://github.com/bitcraze/crazyflie-firmware/blob/c2570a935ffa338f97ff756cdfa71ba470d81f9c/src/modules/src/commander.c#L112-L131

meaning it would hold the last position setpoint for 0.5 seconds and then start drifting for 1.5 seconds (based on the  current values of `COMMANDER_WDT_TIMEOUT_SHUTDOWN` and `COMMANDER_WDT_TIMEOUT_STABILIZE`), then jump down to the landing position, because the landing trajectory is already over by the time it starts following the high-level commander setpoints.

What do you observe?

jpreiss:
No need to test rpyt setpoints - any switching between streaming setpoints and the high-level commander will test the mechanism.

knmcguire:
here is a video of what I made yesterday. I won't come to the office again until early January unfortunately for more flight tests

It does not seem to hold the position for a long time... after the second position setpoint it reacts quite quickly to the HL land command..
https://user-images.githubusercontent.com/11061260/146923490-e1fe2b3d-b5f2-4613-a322-83b685fda5df.mp4

jpreiss:
FYI: the video above does not play correctly for me in macOS 10.14 on Firefox, but works in Safari.

jpreiss:
Based on the code https://github.com/bitcraze/crazyflie-lib-python/blob/fbc491c811127785d3318e0ec45e264c055c1ca3/cflib/crazyflie/commander.py#L140-L144
I think the `send_position_setpoint` commands only send a single packet.

If so, then we should expect the CF to switch to leveling behavior 0.5 seconds into the sleep after the `send_position_setpoint` command.

If the white square on your floor corresponds to the coordinate (0, 0, 0), then drift from the leveling mechanism could explain why the landing is somewhat far to the right of the white square.

Then, the timeout-based switch to high-level for the landing would happen 1 second into the landing trajectory instead of at the end, so the CF has a chance to get close to z=0 before the motors are cut.

It's hard for me to tell what's happening in the video because the commands are of similar durations to the leveling/timeout mechanism. @knmcguire when you get back in the lab next year, could you try rerunning the same experiment but with much longer durations, like 4x everything?

knmcguire:
Yes ofcourse! I'll try it once I'm back. I do have flowdeck with me but not a lot of space to crash, so the flight arena is necessary 😁

And perhaps we can also try to do a longer HL trajectory and in the middle of the duration send a position setpoint? That also seems like a good tryout for the priority switching.

jpreiss:
Yes, that sounds like a good idea.

I was thinking it would be useful to collect these flight test scripts somewhere. Each script could have a comment telling the operator how the CF should behave in flight. When possible, we can have equivalent `crazyflie-lib-python` and `crazyswarm` versions of tests.

@jonasdn would this be in the scope of `crazyflie-testing`? Currently I only see tests that can run without flying, like the logging/param system.

jonasdn:
> Yes, that sounds like a good idea.
> 
> I was thinking it would be useful to collect these flight test scripts somewhere. Each script could have a comment telling the operator how the CF should behave in flight. When possible, we can have equivalent `crazyflie-lib-python` and `crazyswarm` versions of tests.
> 
> @jonasdn would this be in the scope of `crazyflie-testing`? Currently I only see tests that can run without flying, like the logging/param system.

It might be! I wamt to add  som flying tests eventually, both supervised and un-supervised.

knmcguire:
Alright here are some more flights:

I did first two scenarios:

- take off in HL
- position low level command 1.0 meter to left
- wait for 4 seconds

and then
- take off in HL
- High level 1.0 forward
- immediately a low level cmd to the left
- wait for 4 seconds

In both scenarios it moved to the left and then the motors shut off after 2 seconds like this:

https://user-images.githubusercontent.com/11061260/148932016-523d874e-1431-4f56-800b-dd4f4f3c4eba.mp4

Then I tried it out during a planned trajectory
- take off in HL command
- start 8 trajectory
- halfway, send a low level position command  to middle
- wait for 4 seconds

Then you see this:

https://user-images.githubusercontent.com/11061260/148932408-5caf37ee-f5fe-48e5-aa7f-a07d19226e02.mp4

What I do find strange, is that for the low level position setpoints (with   #cf.commander.send_position_setpoint()
), it does not seem to really stop at those locations I've indicated but continues to glide in the same direction, but that is probably due to the fact that the motors go to 'hover' after 1 second and the crazyflie is no longer controlled. 

But in general, it seems that this PR indeeds does the trick for switching from HL to low level

knmcguire:
The experiment before the last one also did show, that after the low-level setpoint is given, it can switch to HL without any problem (if no low level setpoint is given at the same time)

jonasdn:
Does this look mergable to you @jpreiss ?

jpreiss:
I agree, in the most recent experiments the glide/drift past the position setpoint seems consistent with the leveling mechanism activating while the x/y velocity is nonzero.

In the earlier experiments, I agree it demonstrates low-to-high switching, but only in the form of a fallback after drifting the 0.5-second timeout period. To test instantaneous switching, you would need to implement the `notifySetpointsStop` packet in the Bitcraze Python API.

Since it works with Crazyswarm I doubt there's any problem on the firmware side. However, it seems less than ideal that a firmware feature would require Crazyswarm to fully test its functionality.

knmcguire:
Yes that is true... we should implement that in the cflib first, but I've made a issue for it for now. Not sure if we manage to do this before the next immediate release within 1 or two weeks but at least we are reminded.

For now, I will trust that this works if you have already tried it out in crazyswarm :)

knmcguire (approved):

knmcguire:
woops! forgot to merge this for the release.... we will do a few extra test to just double check this doesn't break anything.

PR #904, opened 2021-12-14
New communication for AI-deck on UART2

evoggy:
Implementation of the new protocol that will be used for the ESP32/GAP8 firmware on the AI-deck (CPX) on UART2. The new protocol is not being used yet. It temporarily removes the relaying of console printouts from the ESP32 though CRTP, but this will be added again in the future. The functionality for the GAP8 is left unchanged.

As a side-effect of this change the double-reset of the AI-deck is fixed.


src/deck/drivers/interface/aideck.h
```diff
@@ -0,0 +1,59 @@
+/**
+ *    ||          ____  _ __
+ * +------+      / __ )(_) /_______________ _____  ___
+ * | 0xBC |     / __  / / __/ ___/ ___/ __ `/_  / / _ \
+ * +------+    / /_/ / / /_/ /__/ /  / /_/ / / /_/  __/
+ *  ||  ||    /_____/_/\__/\___/_/   \__,_/ /___/\___/
+ *
+ * Crazyflie control firmware
+ *
+ * Copyright (C) 2021 Bitcraze AB
```
jonasdn, line 10: Only 4% left of 2021, cutting it close!

src/deck/drivers/interface/aideck.h
```diff
@@ -27,33 +27,74 @@
...
+  TEST = 0x0E,
+  BOOTLOADER = 0x0F,
+} __attribute__((packed)) CPXFunction_t;
+
 typedef struct {
-    uint8_t start; // Should be 0xFF
-    uint8_t length; // Length of data
-    uint8_t data[AIDECK_UART_TRANSPORT_MTU];
-} __attribute__((packed)) uart_transport_packet_t;
+  CPXTarget_t destination : 4;
```
jonasdn, line 52: So max 7 targets?

src/deck/drivers/src/aideck.c
```diff
@@ -46,79 +47,265 @@
...
+    while (1)
+    {
+      // Wait for start!
+      do {
+        uart2GetDataWithTimeout(&espRxp.start, (TickType_t)portMAX_DELAY);
+      } while (espRxp.start != 0xFF);
+
+      uart2GetDataWithTimeout(&espRxp.length, (TickType_t)portMAX_DELAY);
+
+      //DEBUG_PRINT("Length is %u\n", espRxp.length);
```
jonasdn: Remove the left-over prints, like this one

src/deck/drivers/src/aideck.c
```diff
@@ -35,62 +35,141 @@
...
-  vTaskDelay(10);
-  digitalWrite(DECK_GPIO_IO4, HIGH);
-  pinMode(DECK_GPIO_IO4, INPUT_PULLUP);
+  uint8_t ctr[] = {0xFF, 0x00};
+  EventBits_t evBits = 0;
+
+  // We need to hold off here to make sure that the RX task
+  // has started up and is waiting for chars, otherwise we might send
+  // CTR and miss CTS (which means that the ESP32 will stop sending CTS
+  // too early and we cannot sync)
```
jonasdn, line 123: Great comment!

src/deck/drivers/src/aideck.c
```diff
@@ -35,62 +35,141 @@
...
+                                   pdTRUE,  // Clear bits before returning
+                                   pdFALSE, // Wait for any bit
+                                   portMAX_DELAY);
+      if ((evBits & ESP_CTR_EVENT) == ESP_CTR_EVENT)
+      {
+        uart2SendData(sizeof(ctr), (uint8_t *)&ctr);
+      }
+      if ((evBits & ESP_TXQ_EVENT) == ESP_TXQ_EVENT)
+      {
+      }
```
jonasdn, line 151: What should happen here?

jonasdn (commented):
It looks really good, well done!

src/deck/drivers/src/aideck.c
evoggy: Thanks

src/deck/drivers/interface/aideck.h
evoggy: It should be 16

src/deck/drivers/src/aideck.c
evoggy: Nothing, there used to be a debug printout so the flow could be followed. Removed the printouts and forgot this to remove the if. I'll update it.

src/deck/drivers/src/aideck.c
evoggy: This seems to be an older commit on this branch, it's not in the latest version.

src/deck/drivers/interface/aideck.h
evoggy: Not sure what this should say, should I update it to 2022?

src/deck/drivers/src/aideck.c
evoggy: I've pushed a new commit to fix this one.

src/deck/drivers/interface/aideck.h
jonasdn: yes, that is right, 2^4, my mistake. Maybe we should write a comment about this next to the enum defining targets? To make it more clear?

src/deck/drivers/interface/aideck.h
jonasdn: No I do not think so, I think it should say _when_ the work was done, since Copyright lapses after X years. So it should be updated only on _major_ revisions that alter the work, I think. A bit fuzzy.

src/deck/drivers/interface/aideck.h
evoggy: Yeah, this is a bit sneaky. The structs were initially done when there was not step in between the user assembling the data and sending it, i.e the struct was the same as on the wire. But now since we're copying it into the wire structure we could skip the enums being packed (and the full struct being packed) and fix this in the cpx send/receive. With asserts making sure that the target/function abides by the correct sizes. Yeah, we should probably do that.

evoggy:
I've pushed a new version where no packed structs are exposed to the user of the API and also fixed bugs.

jonasdn (approved):

jonasdn:
This needs to be rebased against master:

```
# From cf-link-aideck branch
$ git fetch origin
$ git rebase origin/master
$ git push -f origin  cf-link-aideck
```

jonasdn:
I have rebased and forced push the origin/cf-link-aideck branch

PR #905, opened 2021-12-16
Make active marker deck parameters persistent

krichardsson:
To use the Active marker deck in a swarm, the IDs and mode has to be set when the Crazyflie is booted, either from an app or script. By persisting these parameters this is no longer necessary, as the values can be stored in the Crazyflie and will be read back upon reboot.

This PR adds the `PARAM_PERSISTENT` to the appropriate parameters.


jonasdn (approved):
Looks good

PR #906, opened 2021-12-20
param: Add GET_DEFAULT_VALUE to MISC_CHANNEL

jonasdn:


src/modules/src/param_logic.c
```diff
@@ -689,6 +689,32 @@ void paramPersistentStore(CRTPPacket *p)
   crtpSendPacketBlock(p);
 }
 
+void paramGetDefaultValue(CRTPPacket *p)
+{
+  uint16_t id;
+
+  memcpy(&id, &p->data[1], 2);
```
krichardsson, line 696: Perhaps use sizeof(id) instead of 2?

krichardsson (approved):
Looking good to me.
A tiny minor comment if you are bored and want something to do :-)

Issue #900, opened 2021-12-13
eps_deck_flasher make error gcc compiler 7.3


rtkartista:
` CC    esp_deck_flasher.o
.//src/modules/src/esp_deck_flasher.c: In function 'espDeckFlasherWrite':
.//src/modules/src/esp_deck_flasher.c:136:21: error: 'lastPacketFull' may be used uninitialized in this function [-Werror=maybe-uninitialized]
     if ((lastPacket && lastPacketFull))
         ~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~
.//src/modules/src/esp_deck_flasher.c: At top level:
cc1: error: unrecognized command line option '-Wno-stringop-overflow' [-Werror]
cc1: error: unrecognized command line option '-Wno-stringop-overread' [-Werror]
cc1: error: unrecognized command line option '-Wno-address-of-packed-member' [-Werror]
cc1: all warnings being treated as errors
tools/make/targets.mk:26: recipe for target 'esp_deck_flasher.o' failed
make[1]: *** [esp_deck_flasher.o] Error 1
rm version.c
Makefile:400: recipe for target 'build' failed
make: *** [build] Error 2
`

I trice using the make PLATFORM=cf2 command after cloning different versions, I still get the same error with the "espDeckFlasherWrite" function

rtkartista:
I just commented out 
1.  PROJ_OBJ += esp_deck_flasher.o
2. PROJ_OBJ += aideck.o

these lines in the makefile and the make was successful

knmcguire:
Hi! Happy you found a solution.

I just reopened as this looks like an issue that we need to address for this repository as we can't uncomment these lines by standard. So just for our investigation

knmcguire:
So we haven't seen this ourselves with the latest master of the code. Perhaps you just pulled a version that had this build error just at the right time. Also double check your gcc compiler version and if your submodules are all properly updated.

I'll close this issue now but please let us know if you still have this problem.

Moj-Dev:
> I just commented out
> 
> 1. PROJ_OBJ += esp_deck_flasher.o
> 2. PROJ_OBJ += aideck.o
> 
> these lines in the makefile and the make was successful

I have the same problem. Did you succeed to fix the issue? Does your crazyflie work well? In my case, the kalman filter resets!

knmcguire:
[...]

PR #877, opened 2021-10-17
disabled warnings for incorrectly detected sector size (fixes compilation for GCC11)

zeroos:
fixes #876

PR #911, opened 2022-01-05
fixed esp flasher compile error with older gcc versions

knmcguire:
New compiler flags for gcc 11 have been added in PR #877 but these are causing problems for older compiler versions on Ubuntu 18.04 (see #900)

I've added a check in the Makefile so that it will only use these flags if version 11 is used. 

Update: it ended up to be an uninitalized variable in esp_flasher which removed the necessity to check versions for these flags.

### DIFF
--- a/src/modules/src/esp_deck_flasher.c
+++ b/src/modules/src/esp_deck_flasher.c
@@ -99,7 +99,7 @@ bool espDeckFlasherWrite(const uint32_t memAddr, const uint8_t writeLen, const u
   // send buffer if full
   const bool sendBufferFull = (sendBufferIndex == ESP_MTU);
   const bool lastPacket = (sequenceNumber == numberOfDataPackets - 1);
-  bool lastPacketFull = false;
+  bool lastPacketFull = lastPacket && (sendBufferIndex == ESP_BITSTREAM_SIZE % ESP_MTU);
   if (lastPacket)
   {
     lastPacketFull = (sendBufferIndex == ESP_BITSTREAM_SIZE % ESP_MTU);
--- a/src/modules/src/esp_deck_flasher.c
+++ b/src/modules/src/esp_deck_flasher.c
@@ -99,7 +99,7 @@ bool espDeckFlasherWrite(const uint32_t memAddr, const uint8_t writeLen, const u
   // send buffer if full
   const bool sendBufferFull = (sendBufferIndex == ESP_MTU);
   const bool lastPacket = (sequenceNumber == numberOfDataPackets - 1);
-  bool lastPacketFull;
+  bool lastPacketFull = false;
   if (lastPacket)
   {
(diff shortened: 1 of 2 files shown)


knmcguire:
after these changes I still get: 

```
.//src/modules/src/esp_deck_flasher.c: In function 'espDeckFlasherWrite': 
.//src/modules/src/esp_deck_flasher.c:136:21: error: 'lastPacketFull' may be used uninitialized in this function [-Werror=maybe-uninitialized]
if ((lastPacket && lastPacketFull)) 
~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~ 
cc1: all warnings being treated as errors       
```

So there are two ways to fix this:
- either initializing lastPacketFull = False
- use the flag `CFLAGS +=-Wno-maybe-uninitialized` in the Makefile

knmcguire:
Going for the first options, and this also eliminates the need to check for compiler versions in the Makefile

src/modules/src/esp_deck_flasher.c
```diff
@@ -99,7 +99,7 @@ bool espDeckFlasherWrite(const uint32_t memAddr, const uint8_t writeLen, const u
   // send buffer if full
   const bool sendBufferFull = (sendBufferIndex == ESP_MTU);
   const bool lastPacket = (sequenceNumber == numberOfDataPackets - 1);
-  bool lastPacketFull;
+  bool lastPacketFull = false;
```
jonasdn, line 102: nit: this _could_ be:

```c
bool lastPacketFull = lastPacket && (sendBufferIndex == ESP_BITSTREAM_SIZE % ESP_MTU);
```

And we could remove the if alltogether.... not sure it is better though

knmcguire: thanks! done. I kept the if statement since I don't know if it is necessary.

gemenerik (approved):
Looks good!

src/modules/src/esp_deck_flasher.c
```diff
@@ -99,7 +99,7 @@ bool espDeckFlasherWrite(const uint32_t memAddr, const uint8_t writeLen, const u
   // send buffer if full
   const bool sendBufferFull = (sendBufferIndex == ESP_MTU);
   const bool lastPacket = (sequenceNumber == numberOfDataPackets - 1);
-  bool lastPacketFull = false;
+  bool lastPacketFull = lastPacket && (sendBufferIndex == ESP_BITSTREAM_SIZE % ESP_MTU);
```
jonasdn, line 102: With this the if-statement below is not needed

knmcguire: ah oke, thought as it was still up in the air, that it was better just to keep it there to sbe sure. but I'll remove it

PR #912, opened 2022-01-10
fixed variance calculation cf2.0 and cf2.1 imu

knmcguire:
a reimplementation of #665, without conflicts and also integrated for CF2.0's IMU. 

Credits to @Leonana69


jonasdn (approved):
:100:

PR #914, opened 2022-01-14
Python bindings: add collision_avoidance

whoenig:


jonasdn:
So, @whoenig, I would like to build an RC, soon, so I think I will merge this since it is self-contained, and we can fix issues with it during the release process or after and do a point release, does that sound ok?

PR #919, opened 2022-01-24
sound: Do not allow bad sound effect

jonasdn:
Do not allow writes outside of effect array.


ataffanel (approved):
Looks good, as it should have been ...

PR #925, opened 2022-01-27
Replace Cygwin by WSL in build doc for Windows

ataffanel:
Add support for bootloading from WSL using the Windows python and cfclient.

Change the build instruction to use WSL instead of Cygwin.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,6 @@ PYTHON            ?= python3
 DFU_UTIL          ?= dfu-util
 CLOAD             ?= 1
 DEBUG             ?= 0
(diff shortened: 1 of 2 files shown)


knmcguire:
Will try it now on my system :)

knmcguire:
works like a charm! just beautiful 😄 I've added some small changes to the doc just to clarify things

knmcguire (approved):

matejkarasek:
Works great indeed, and the instructions are also clear!

Funny coincidence, it seems USB devices can now finally be accessed from WSL:
https://docs.microsoft.com/en-us/windows/wsl/connect-usb

PR #929, opened 2022-02-01
Relay messages from GAP8 printf to the console

whoenig:
Not a very advanced version, but IMO better than nothing.


tobbeanton (approved):
There is a big revamp going on but until then this is good.

PR #931, opened 2022-02-02
Bolt + BQdeck persistent parameters

knmcguire:
…tent

### DIFF
--- a/src/modules/src/attitude_pid_controller.c
+++ b/src/modules/src/attitude_pid_controller.c
@@ -298,37 +298,37 @@ PARAM_GROUP_START(pid_rate)
 /**
  * @brief Proportional gain for the PID roll rate controller
  */
-PARAM_ADD(PARAM_FLOAT, roll_kp, &pidRollRate.kp)
+PARAM_ADD(PARAM_FLOAT|PARAM_PERSISTENT, roll_kp, &pidRollRate.kp)
 /**
  * @brief Integral gain for the PID roll rate controller
  */
-PARAM_ADD(PARAM_FLOAT, roll_ki, &pidRollRate.ki)
+PARAM_ADD(PARAM_FLOAT|PARAM_PERSISTENT, roll_ki, &pidRollRate.ki)
 /**
  * @brief Derivative gain for the PID roll rate controller
  */
-PARAM_ADD(PARAM_FLOAT, roll_kd, &pidRollRate.kd)
+PARAM_ADD(PARAM_FLOAT|PARAM_PERSISTENT, roll_kd, &pidRollRate.kd)
 /**
  * @brief Proportional gain for the PID pitch rate controller
  */
-PARAM_ADD(PARAM_FLOAT, pitch_kp, &pidPitchRate.kp)
+PARAM_ADD(PARAM_FLOAT|PARAM_PERSISTENT, pitch_kp, &pidPitchRate.kp)
 /**
  * @brief Integral gain for the PID pitch rate controller
  */
-PARAM_ADD(PARAM_FLOAT, pitch_ki, &pidPitchRate.ki)
+PARAM_ADD(PARAM_FLOAT|PARAM_PERSISTENT, pitch_ki, &pidPitchRate.ki)
 /**
  * @brief Derivative gain for the PID pitch rate controller
  */
-PARAM_ADD(PARAM_FLOAT, pitch_kd, &pidPitchRate.kd)
+PARAM_ADD(PARAM_FLOAT|PARAM_PERSISTENT, pitch_kd, &pidPitchRate.kd)
 /**
  * @brief Proportional gain for the PID yaw rate controller
  */
-PARAM_ADD(PARAM_FLOAT, yaw_kp, &pidYawRate.kp)
+PARAM_ADD(PARAM_FLOAT|PARAM_PERSISTENT, yaw_kp, &pidYawRate.kp)
 /**
  * @brief Integral gain for the PID yaw rate controller
  */
-PARAM_ADD(PARAM_FLOAT, yaw_ki, &pidYawRate.ki)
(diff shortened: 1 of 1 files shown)


knmcguire:
actually we should have mass outside of the Mellinger controller but as a more global assessable parameters.

knmcguire:
So currently tuning values in both PID and Mellinger, as noise values in the EKF are persistent in this PR. Anything else missing?

knmcguire:
Perhaps the nonstandard IMU orientation values? But then these need to be converted to a parameters as well. @matejkarasek  what do you think?

knmcguire:
Just going to start a list here of what I think should be persistent based on the parameter list.
- [ ] [cmdrCPPM](https://www.bitcraze.io/documentation/repository/crazyflie-firmware/master/api/params/#cmdrcppm)
- [ ] [ctrlINDI](https://www.bitcraze.io/documentation/repository/crazyflie-firmware/master/api/params/#ctrlindi)
- [ ] [ctrlMel](https://www.bitcraze.io/documentation/repository/crazyflie-firmware/master/api/params/#ctrlmel)

matejkarasek:
> Perhaps the nonstandard IMU orientation values? But then these need to be converted to a parameters as well. @matejkarasek what do you think?

Yes, good idea to doo that.

If we consider some motors can be servos, two more ideas (that will require code changes):
- for supervisor.c, param that says which motors are "thrusters" and which servos
- for motors.c, param that specifies the neutrel position at which the (servo) motor initializes

These would certainly deserve a new PR, though... can help...

matejkarasek:
And two more ideas (with code changes needed):
- 1S voltage is now hardcoded, but with Bolt you can have 2S-4S
- thrust mapping?
https://github.com/bitcraze/crazyflie-firmware/blob/b0c72f2a4cb8b432211a2fa38d97c5a1dcef07ff/src/drivers/src/motors.c#L109-L173

matejkarasek:
For lighthouse, it could certainly be the position of the markers on the airframe...

knmcguire:
Very good input @matejkarasek, thanks! 

For the controller gains, currently I made the attitude rate parameters persistent and planning now to do attitude paramters. Theoretically, or rather in an ideal world,  the position and velocity gains should not have to be changed... but ofcourse reality is different. 
Do you need to hardcode those pos and vel pid values for your platform ?

knmcguire:
About your comments about motors.c and the initialized thrust, Isn't the param group [powerdist](https://www.bitcraze.io/documentation/repository/crazyflie-firmware/master/api/params/#powerdist) about that?

matejkarasek:
Could be a suitable param group indeed. Now it contains only idle power for brushless motors.  
I suggest adding the startup power, which is the "power"/pwm signal/position of the motor at startup. This should be 0 for motors giving thrust (as hardcoded now) but for servos this should be a non-zero value, configurable for each servo separately...

matejkarasek:
> Do you need to hardcode those pos and vel pid values for your platform ?
Yes, we do... #reality :)

knmcguire:
> For lighthouse, it could certainly be the position of the markers on the airframe...

This will be a bit difficult for the yaw estimation due to a strong assumption that the deck needs to be flat and on top of the platform. But it's a good one thought! I have put it in the list so it least it is written down.

knmcguire:
@matejkarasek with the thrust mapping and 1s voltage you mean the values that are hardcoded in these lines?
https://github.com/bitcraze/crazyflie-firmware/blob/b0c72f2a4cb8b432211a2fa38d97c5a1dcef07ff/src/drivers/src/motors.c#L163
https://github.com/bitcraze/crazyflie-firmware/blob/b0c72f2a4cb8b432211a2fa38d97c5a1dcef07ff/src/drivers/src/motors.c#L168-L169

matejkarasek:
Yes indeed, for the thrust mapping those two lines.

For the battery voltage, there might be a few more places in the code... Like detection of low voltage and indication with the LEDs...

matejkarasek:
> > For lighthouse, it could certainly be the position of the markers on the airframe...
> 
> This will be a bit difficult for the yaw estimation due to a strong assumption that the deck needs to be flat and on top of the platform. But it's a good one thought! I have put it in the list so it least it is written down.

Yeah, true... Well it could be xyz position of the deck on the airframe at least (now in the code you do that per receiver I think). Just an idea anyway for possible future updates :)

knmcguire:
For now there are 90 parameters set as persistent.

knmcguire:
So I've just flashed this on a crazyflie, set every persistent parameter in eeprom that is possible, restarted the crazyflie and...... no crash! So it seems to be doing pretty good :)

tobbeanton:
I think the list is a really good start. For those properties that do not have a parameter yet I think we should add them in separate pull requests such as board alignment, battery levels etc. As we have also decided to shrink the platform concept a bit, due to different default values, and introduce kbuild to make selection easy. Thus the CF2.1, Bolt and possibly Flapper would be its's own platform with own binary (own compile time flags).

One parameter I'm thinking about the the forceArm which is tagged now. The whole idea of that parameter is for something external to have to set it to be able to arm the motors. If this could be set permanently the purpose might be gone. On the other side, it would be easy to set it to different default value for different platform. E.g CF2.1 will not require forceArm but Bolt could. I think I changed my mind, let's keep it as persistent.

We should keep in mind also that the 8Kbyte EEPROM can store somewhere between 200-300 parameters in total.

krichardsson (approved):

matejkarasek:
Nice start indeed! Already helps to minimize the amount of changes needed in our fork compared to master..

I agree it might be even better to split the current CF2 platform into more specific ones (CF2.1, Bolt, Flapper, ...) as it would simplify the very first setup, no need to change to the platform-specific default values after the first flashing...

PR #934, opened 2022-02-07
Add mechanism for default parameters

krichardsson:
A need for multiple default values of variables (parameters) depending on hardware has been identified as part of the work on the Crazyflie Bolt, for instance tuning. This PR adds functionality to make it easy to add defines that will be used depending on which platform that is built.


Makefile
```diff
@@ -38,6 +38,9 @@ endif
 -include current_platform.mk
 include $(CRAZYFLIE_BASE)/tools/make/platform.mk
 
+# Create a define for the platform (upper case), for instance PLATFORM_CF2
+CFLAGS += -DPLATFORM_$(shell echo $(PLATFORM) | tr a-z A-Z)=1
```
jonasdn, line 42: nit: this could be:

```
CFLAGS += -DPLATFORM_$(shell echo $(PLATFORM) | tr '[:lower:]' '[:upper:]')=1
```

Not sure if it is better though, a bit more self-explanatory.

src/platform/interface/platform_defaults_cf2.h
```diff
@@ -0,0 +1,9 @@
+#pragma once
+
+#ifndef __INCLUDED_FROM_PLATFORM_DEFAULTS__
+    #pragma GCC error "Do not include this file directly, include platform_defaults.h instead."
+#endif
+
+// Defines for default values in the cf2 platform
+
+// #define EXAMPLE_DEFAULT_VALUE 0.4
```
jonasdn, line 9: Nice!

jonasdn (approved):

Issue #923, opened 2022-01-27
Add USB message to reboot to DFU bootloader


ataffanel:
A need has been expressed on the forum to be able to easily program the STM32 using USB. Currently DFU programming is implemented as a failsafe functionality and so is pretty hard to access. 

If there was a way to reboot in dfu mode from USB, we could implement a ```make flash-dfu``` make target that restarts the Crazyflie in DFU mode, flash it and restart it in firmware mode. This would make working with the Crazyflie without radio a lot easier and could also help regular with-radio Crazyflie development since flashing via USB should be at least twice as fast as flashing via the radio.

I described in the forum a way to implement this functionality: https://forum.bitcraze.io/viewtopic.php?p=21533#p21533

C-Glick:
Hi ataffanel, I'd be interested in looking into this further (I was the one who recently mentioned this topic on the forum thread). 

I'm fairly new to the Crazyflie firmware so I'm trying to find my way as I go and could use some guidance. The code you linked for rebooting the CF into DFU mode seems complete (https://github.com/bitcraze/lps-node-firmware/pull/19/files) just need to run those functions at the right location.

I am trying to understand the USB setup request now.
https://github.com/bitcraze/crazyflie-firmware/blob/61f3506cce6bfbed21808676caf38d807f823c45/src/hal/src/usb.c#L167

 Is `usbd_cf_Setup()` only called when a USB cable is initially connected to the CF? It looks like right now all it does is set the CRTP link to the USB link when the index is 0x01.

Ideally, the CF would already be connected via USB and we would then send a signal (through CRTP?) over USB to initiate the reboot. Is this possible using a setup request?

C-Glick:
@ataffanel I've done some work on this but have been having issues and would like some guidance. 
I've got the crazyflie 2.0 successfully responding to a USB request but the code to jump to the DFU mode doesn't appear to be working. It doesn't seem the HAL functions used in the example given are available in the stock firmware, so I converted them to what appears to be equivalent functions. When responding to the USB request, the crazyflie reboots the firmware into a failed self-test state (red led flashing 5 times) if I attempt to jump to DFU mode again it completely locks up and I need to remove power to restore functionality. 

[...]

PR #939, opened 2022-02-14
Streamline DFU flashing over USB

C-Glick:
## Description
- This pull request streamlines the process of flashing the Crazyflie over a USB connection utilizing the STM32F405's DFU mode. 
- Currently, DFU flashing is intended as a recovery backup method, however, in some scenarios it is useful to be able to easily flash the Crazyflie over USB without a Crazyradio. This improves that process.
- Through the `make flash_dfu` command, a python script sends a USB control transfer request with `wIndex = 0x02` to the Crazyflie, the Crazyflie then handles that request by setting a flag in SRAM and rebooting. 
- Upon every start-up, this flag is checked and if set correctly, will jump to the STM32405's DFU mode. 
- Next `dfu-util` is used to download the new cf2.dfu file to the Crazyflie, and automatically transition back to firmware mode once completed.

## References
- Issue #923 
- Crazyflie forum post: https://forum.bitcraze.io/viewtopic.php?p=23802 

## Comments/Concerns
- The storage address for the bootloader flag was chosen to be the start of the 16kb SRAM bank `SRAM2_BASE = 0x2001C000` (see [STM32F405 data sheet](https://www.st.com/resource/en/datasheet/dm00037051.pdf) page 71) this shouldn't conflict with anything else, as the only time it is modified is right before a restart, and it is set to 0 during the startup check. Would there be any future issues with using this address?
- Currently the `check_enter_bootloader` function is called before `platformInit`, I did this to minimize the number of unknown operations that could modify the state of the microcontroller before jumping to the bootloader. This system initialization might need to be performed before the bootloader check to check we are running on the right hardware.
- `usbd_cf_Setup` never returns, as the system reboots while handling the request, this causes an IO error in python that is expected. I believe the system would need to return from `usbd_cf_Setup` for this python error to be avoided. Would there be an easy way to call `enter_bootloader` after the USB request is responded to?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -168,7 +168,7 @@ PROJ_OBJ += vl53l1_api_calibration.o vl53l1_api_debug.o vl53l1_api_preset_modes.
 PROJ_OBJ += vl53l1_register_funcs.o vl53l1_wait.o vl53l1_core_support.o
 
 # Modules
-PROJ_OBJ += system.o comm.o console.o pid.o crtpservice.o param_task.o param_logic.o
+PROJ_OBJ += system.o comm.o console.o pid.o crtpservice.o param_task.o param_logic.o bootloader.o
 PROJ_OBJ += log.o worker.o queuemonitor.o msp.o
 PROJ_OBJ += platformservice.o sound_cf2.o extrx.o sysload.o mem.o
 PROJ_OBJ += range.o app_handler.o static_mem.o app_channel.o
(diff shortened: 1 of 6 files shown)


ataffanel (approved):
Looks good and works great. Thanks!

We are currently doing some big work on the Makefile (hence the merge conflict) so I will wait a bit for the dust to settle down on the master branch before attempting to merge the PR.

C-Glick:
Sounds good, happy to contribute to an excellent project!
Let me know if any further action is needed on my part.

jonasdn:
Merged manually as 6c33f7eae10e539c9c1affc015776651aeb37e2e

PR #947, opened 2022-02-18
Do not take address of packed member

jonasdn:
- pptraj.h: Do not pack struct poly4d
- Build: Do not ignore warnings about packed members


jonasdn:
> My initial feeling is that if you can still upload and execute a trajectory with multiple pieces that it should be fine. Also, the members are already all floats, so I guess they should be actually all consecutive anyways. Given that, I am not sure why this would have caused a hard fault...

This is my understanding as well ... that they should be consecutive.

My guess is that the compiler sees that it is a "packed member" and takes that as an excuse for some more aggresive optimization and the struct ends up not aligned when accessed from inside an unpacked struct.

PR #951, opened 2022-02-22
Kbuild: Convert MADWICK_QUATERNION define to config

jonasdn:
Add a new "IMU configuration" menu to Kconfig and add a choice between
the two different IMU algorithms offered.


Kconfig
```diff
@@ -50,6 +50,27 @@ endchoice
...
+    default IMU_MAYHONY_QUATERNION
+
+config IMU_MADGWICK_QUATERNION
+    bool "Madgwick's AHRS and IMU algorithms"
+    help
+        Use Madgwick's IMU and AHRS algorithms.
+        See: http://www.x-io.co.uk/open-source-ahrs-with-x-imu
+
+config IMU_MAYHONY_QUATERNION
+    bool "Madgwick's implementation of Mayhony"
```
knmcguire, line 66: To avoid confusion, just call this Mayhony and refer to the paper they refer too? https://ieeexplore.ieee.org/document/4608934

knmcguire:
wow, didn't think these configs excited! There are so many...

Kconfig
jonasdn: Agreed! I pushed a commit to fix it up!

PR #953, opened 2022-02-22
Kbuild: Convert DEBUG_PRINT_ON_UART1 to kconfig

jonasdn:


src/modules/src/system.c
```diff
@@ -171,8 +171,8 @@ void systemTask(void *arg)
   queueMonitorInit();
 #endif
 
-#ifdef ENABLE_UART1
-  uart1Init(9600);
+#ifdef CONFIG_DEBUG_PRINT_ON_UART1
+  uart1Init(115200);
```
knmcguire, line 175: Ah yeah good catch. Don't know why this was different then uart2 before.

Kconfig
```diff
@@ -24,6 +24,14 @@ config DEBUG_QUEUE_MONITOR
     help
         Enable the queue monitoring functionality.
 
+config DEBUG_PRINT_ON_UART1
+    bool "Switch DEBUG_PRINT to UART1"
+    default n
+    help
+      Switch output of DEBUG_PRINT from radio to UART1 with 115200 baudrate.
+      Note! This will interfere with several expansion decks, see:
+      https://www.bitcraze.io/documentation/system/platform/cf2-expansiondecks/#deck-pin-allocation
```
knmcguire, line 33: it's a bit dangerous to put links in these places. These tend to get forgotten very easily, as we noticed lately 😄 wonder if we should make a short link to the expansiondecks instead. Or just refer them to the website and search deck pin allocation?

knmcguire: With forgotten, I mean if we every change the urls of these things and such

jonasdn: Agreed, pushed a commit to make it a bit safer.

knmcguire (approved):

knmcguire:
perfect. thanks!

tobbeanton (commented):
All looks very nice. I really like that we can use the dependency to prevent decks from compiling at the same time.

PR #954, opened 2022-02-23
Kbuild: Convert CALIBRATED_LED_MORSE to Kconfig

jonasdn:


jonasdn:
Does this belong in the _Build and debug options_ menu? @knmcguire, do you have an opnion?

knmcguire:
Actually... I'm not sure what this define does! Let me investigate a bit

knmcguire (approved):

knmcguire:
Its a pretty old define but I think it is still handy for those that want to have more visual feedback that the crazyflie has been properly calibrated. 

So yes, build and debug options is a good location for this.

Issue #960, opened 2022-02-23
Kbuild: adding C++ support?


ntamas:
I'm testing the new Kbuild-based build system and trying to convert an app of our own to the new build system. One issue that I can't seem to work around is that some of the files in the app are written in C++ because of historical reasons. With the previous build system, I was able to add support for C++ files to the makefile simply by adding extra rules to the app-level Makefile like this:

```
CXX = $(CROSS_COMPILE)g++
LD = $(CROSS_COMPILE)g++

CXX_COMMAND=$(CXX) $(CFLAGS) -c $< -o $(BIN)/$@
CXX_COMMAND_SILENT="  CXX   $@"
.cpp.o:
     @$(if $(QUIET), ,echo $(CXX_COMMAND$(VERBOSE)) )
     @$(CXX_COMMAND)
```

It seems like this is not possible with the new Kbuild system because the app-level makefile eventually ends up invoking another `make` process with the top-level makefile, which does not have support for C++ files. In the previous build system, the top-level makefile was included instead of forking off to another `make` process so it was easy to add the modifications.

Would it be possible to add support for C++ source files directly in `tools/make/targets.mk` even if the current Crazyflie firmware does not use any C++ sources?

jonasdn:
Hi @ntamas !

Could you try the linked pull-request and see if it fixes stuff for you?

Thanks!

ntamas:
It seems like it's going to be a bit more complicated. In order for the linking stage to work properly, one probably needs to use `$(CROSS_COMPILE)g++` as the linker and not `$(CROSS_COMPILE)gcc` if at least one C++ file is involved in the build process. I'm not sure whether there's an easy way around it to handle this in `kbuild` itself; right now I'm just patching `tools/kbuild/Makefile.kbuild` to use the appropriate linker.

jonasdn:
> It seems like it's going to be a bit more complicated. In order for the linking stage to work properly, one probably needs to use `$(CROSS_COMPILE)g++` as the linker and not `$(CROSS_COMPILE)gcc` if at least one C++ file is involved in the build process. I'm not sure whether there's an easy way around it to handle this in `kbuild` itself; right now I'm just patching `tools/kbuild/Makefile.kbuild` to use the appropriate linker.

Ok. What issues are you seeing in linking and what is your current work-around? Let's see if we can manage an ok permanent solution.

jonasdn:
[...]

PR #962, opened 2022-02-23
Support C++ in app-layer

jonasdn:
- Kbuild: Add support for C++ (.cc and .cpp) source
- app-layer: Add C++ example app

Fixes: #960


scripts/Makefile.lib
```diff
@@ -70,8 +70,10 @@ _c_flags       = $(filter-out $(CFLAGS_REMOVE_$(basetarget).o), $(orig_c_flags))
 orig_a_flags   = $(KBUILD_CPPFLAGS) $(KBUILD_AFLAGS) $(KBUILD_SUBDIR_ASFLAGS) \
                  $(asflags-y) $(AFLAGS_$(basetarget).o)
 _a_flags       = $(filter-out $(AFLAGS_REMOVE_$(basetarget).o), $(orig_a_flags))
-_cpp_flags     = $(KBUILD_CPPFLAGS) $(cppflags-y) $(CPPFLAGS_$(@F))
+_cpp_flags     = $(KBUILD_CPPFLAGS)  $(KBUILD_CFLAGS) $(cppflags-y) $(CPPFLAGS_$(@F))
```
ntamas, line 73: Isn't `_cpp_flags` supposed to be the C _preprocessor_ flags only? Usually it is `CXXFLAGS` and similar that contains C++-specific flags.

jonasdn: You are right, I will make it more clear.

scripts/Makefile.build
```diff
@@ -135,6 +135,9 @@ quiet_cmd_cc_o_c = CC $(quiet_modtag)  $@
 
 cmd_cc_o_c = $(CC) $(c_flags) -c -o $@ $<
 
+cmd_cxx_o_cpp = $(CXX)  $(cpp_flags) -c -o $@ $<
```
ntamas, line 138: I needed to use `$(c_flags)` here to make things work, not `$(cpp_flags)`. In particular, include directories added in `Kbuild` files with `ccflags-y += ...` did not work until I changed this to `$(c_flags)`.

scripts/Makefile.lib
```diff
@@ -70,8 +70,10 @@ _c_flags       = $(filter-out $(CFLAGS_REMOVE_$(basetarget).o), $(orig_c_flags))
...
 
+# Remove all non c++ compatible flags
+_cpp_flags     := $(filter-out -Wno-pointer-sign -Werror=implicit-int -std=gnu11,$(_cpp_flags))
```
ntamas, line 76: I needed to filter stuff from `_c_flags` and not `_cpp_flags` due to the other change I mentioned above.

ntamas:
It works! At least with the changes I outlined above. The part that was not working is that I was adding an extra include folder to the build in `Kbuild` with `subdir-ccflags-y += -I$(srctree)/app/vendor/some-library/include` and it was not picked up until I switched to `_c_flags`.

Thanks for the quick fix!

scripts/Makefile.build
jonasdn: Could you try with latest added commit in PR?

scripts/Makefile.lib
jonasdn: Fixed in later commit

jonasdn:
> 

Thank you for testing it out! Could you try with the last commit: "Kbuild: Use cxx_flags for C++" ? To see if it solves your last issue.

scripts/Makefile.build
```diff
@@ -181,6 +190,15 @@ $(obj)/%.o: $(src)/%.c $(objtool_obj) FORCE
 	$(call cmd,force_checksrc)
 	$(call if_changed_rule,cc_o_c)
 
+# Built-in and composite module parts
```
ntamas, line 193: I think this line is not needed here; it was probably copied accidentally.

jonasdn: You are correct. :)

scripts/Makefile.lib
ntamas: Note that you should not add `$(KBUILD_CFLAGS)` then to `_cpp_flags` as `$(KBUILD_CFLAGS)` are flags for the C compiler and `_cpp_flags` are flags for the preprocessor.

ntamas:
Yes, it works now! One tiny nitpick is that the output now shows `CC something.o` for C compilations but `arm-none-eabi-g++ something.o` for C++ files (and for some reason it includes the full path), but I couldn't figure out where this output is coming from. It doesn't bother me, but if you are super pedantic, this is the only issue left :)

jonasdn:
> Yes, it works now! One tiny nitpick is that the output now shows `CC something.o` for C compilations but `arm-none-eabi-g++ something.o` for C++ files (and for some reason it includes the full path), but I couldn't figure out where this output is coming from. It doesn't bother me, but if you are super pedantic, this is the only issue left :)

Thanks!

It was a typo $(CXX) vs CXX, fixed in latest commit.

```
 CXX     /home/jonasdn/sandbox/kbuild-firmware/examples/app_hello_world-cpp/src/hello_world.o
```

tobbeanton (approved):
Looks good!

jonasdn:
> > Yes, it works now! One tiny nitpick is that the output now shows `CC something.o` for C compilations but `arm-none-eabi-g++ something.o` for C++ files (and for some reason it includes the full path), but I couldn't figure out where this output is coming from. It doesn't bother me, but if you are super pedantic, this is the only issue left :)
> 
> Thanks!
> 
> It was a typo $(CXX) vs CXX, fixed in latest commit.
> 
> ```
>  CXX     /home/jonasdn/sandbox/kbuild-firmware/examples/app_hello_world-cpp/src/hello_world.o
> ```

The full-path I think is an artifact of the oot build.

ntamas:
Before you guys merge this, here's a proposed patch to the build system that gets rid of some annoying `clang` warnings when building `Kbuild` itself on macOS:

```patch
diff --git a/tools/kbuild/Makefile.kbuild b/tools/kbuild/Makefile.kbuild
index 017ad491..fd7586c6 100644
--- a/tools/kbuild/Makefile.kbuild
+++ b/tools/kbuild/Makefile.kbuild
@@ -217,12 +217,17 @@ CONFIG_SHELL := $(shell if [ -x "$$BASH" ]; then echo $$BASH; \
 HOSTCC       = gcc
 HOSTCXX      = g++
 HOSTCFLAGS   = -Wall -Wmissing-prototypes -Wstrict-prototypes -O2
-HOSTCFLAGS  += -fomit-frame-pointer -Wno-format-overflow -std=gnu89
+HOSTCFLAGS  += -fomit-frame-pointer -std=gnu89
 HOSTCXXFLAGS = -O2

 ifeq ($(shell $(HOSTCC) -v 2>&1 | grep -c "clang version"), 1)
+# These flags are clang-only
 HOSTCFLAGS  += -Wno-unused-value -Wno-unused-parameter \
-               -Wno-missing-field-initializers -fno-delete-null-pointer-checks
+               -Wno-missing-field-initializers -fno-delete-null-pointer-checks \
+               -Wno-format-security
+else
+# These flags are gcc-only
+HOSTCFLAGS  += -Wno-format-overflow
 endif

 # Decide whether to build built-in, modular, or both.
```

(Sent this earlier on Discord).

jonasdn:
> ccflags

Thanks I will test this a bit and add as a separate PR!

jonasdn:
> Before you guys merge this, here's a proposed patch to the build system that gets rid of some annoying `clang` warnings when building `Kbuild` itself on macOS:
> 
> ```diff
> diff --git a/tools/kbuild/Makefile.kbuild b/tools/kbuild/Makefile.kbuild
> index 017ad491..fd7586c6 100644
> --- a/tools/kbuild/Makefile.kbuild
> +++ b/tools/kbuild/Makefile.kbuild
> @@ -217,12 +217,17 @@ CONFIG_SHELL := $(shell if [ -x "$$BASH" ]; then echo $$BASH; \
>  HOSTCC       = gcc
>  HOSTCXX      = g++
>  HOSTCFLAGS   = -Wall -Wmissing-prototypes -Wstrict-prototypes -O2
> -HOSTCFLAGS  += -fomit-frame-pointer -Wno-format-overflow -std=gnu89
> +HOSTCFLAGS  += -fomit-frame-pointer -std=gnu89
>  HOSTCXXFLAGS = -O2
> 
>  ifeq ($(shell $(HOSTCC) -v 2>&1 | grep -c "clang version"), 1)
> +# These flags are clang-only
>  HOSTCFLAGS  += -Wno-unused-value -Wno-unused-parameter \
> -               -Wno-missing-field-initializers -fno-delete-null-pointer-checks
> +               -Wno-missing-field-initializers -fno-delete-null-pointer-checks \
> +               -Wno-format-security
> +else
> +# These flags are gcc-only
> +HOSTCFLAGS  += -Wno-format-overflow
>  endif
> 
>  # Decide whether to build built-in, modular, or both.
> ```
> 
> (Sent this earlier on Discord).
> 
> _Edit_: this is on an Apple M1 CPU with clang 13, not sure whether it appears on Intel Macs.

Thanks I will test this a bit and add as a separate PR!

PR #967, opened 2022-02-23
Add Kconfig options for some LED, motor and parameter related settings

ntamas:
This PR restore support for the pre-Kconfig `SILENT_PARAM_UPDATES`, `LEDRING_DEFAULT_EFFECT` and `LED_RING_NBR_LEDS` options (now renamed to use the `CONFIG_` prefix).

It also adds support for configuring whether the drone starts in armed or disarmed mode, and the PWM ratios to use during idling and motor tests.


jonasdn (commented):
Looks good!

How about some of these being selected automagicly when you select `PLATFORM_BOLT`?

jonasdn:
> Looks good!
> 
> How about some of these being selected automagicly when you select `PLATFORM_BOLT`?

As in:

```
config PLATFORM_BOLT
    bool "Build for Bolt"
    select SENSORS_BMI088_BMP388
    select SENSORS_BMI088_SPI
    select MOTORS_START_DISARMED
```

I don't know if any more configs make sense, @tobbeanton ?

src/deck/drivers/src/Kconfig
```diff
@@ -93,6 +93,22 @@ config DECK_LEDRING
       control from your computer. Two additional strong front facing
       LEDs that can be switched on and off, act as headlights.
 
+config DECK_LEDRING_DEFAULT_EFFECT
+    int "Default light effect to use on the LED ring"
+    depends on DECK_LEDRING
+    default 6
+    help
+      Index of the default light effect that the LED ring should use
+      after boot. Use zero to turn off the LEDs when booting.
```
jonasdn, line 102: Would it make sense to list what the effects are here? Or is that superfluous?

src/modules/src/Kconfig
```diff
@@ -59,3 +59,49 @@ config ESTIMATOR_OOT
...
+
+config MOTORS_DEFAULT_IDLE_THRUST
+    int "Default idle thrust for motors in armed state"
+    range 0 65535
+    default 0
+    depends on MOTORS_START_DISARMED
+    help
+      Default thrust for motors when idling in armed state, expressed as an
+      integer in the range 0 to 65535. This can be overridden with parameters;
+	  the value specified here applies to the case when the persistent storage
```
jonasdn, line 80: Some indent shenanigans here.

src/modules/src/Kconfig
```diff
@@ -59,3 +59,49 @@ config ESTIMATOR_OOT
...
+	  does not contain an idle thrust value.
+
+config MOTORS_DEFAULT_PROP_TEST_PWM_RATIO
+    int "Override default PWM ratio to use during motor tests"
+    range 0 65535
+    default 0
+    help
+      PWM ratio to use during motor tests, expressed as an integer in the range
+	  0 to 65535. Zero means not to override the PWM ratio; in this case, the
```
jonasdn, line 89: Indent issue here as well.

src/modules/src/Kconfig
```diff
@@ -59,3 +59,49 @@ config ESTIMATOR_OOT
...
+
+endmenu
+
+menu "Parameter subsystem"
+
+config PARAM_SILENT_UPDATES
+    bool "Silent parameter updates"
+    default n
+    help
+      Do not send CRTP packets when a parameter is updated
```
jonasdn, line 104: Do we need a stronger note here? Some examples/behavior in the python lib will break if this is enabled, right? @ataffanel ? @knmcguire ?

src/modules/src/Kconfig
```diff
@@ -59,3 +59,49 @@ config ESTIMATOR_OOT
...
+	  0 to 65535. Zero means not to override the PWM ratio; in this case, the
+	  drone uses full thrust for brushed motors and does not spin up brushless
+	  motors (for safety reasons). It is therefore advised to set this parameter
+	  explicitly for brushless builds to enable the motor test functionality.
+	  
+	  The value specified here can also be overridden with parameters.
+
+endmenu
+
+menu "Parameter subsystem"
```
jonasdn, line 98: Perhaps 
```
menu "Parameter- and logging subsystems"
```
?

jonasdn: Or perhaps we wait to see if there will be any logging configs.

src/deck/drivers/src/Kconfig
ntamas: I wouldn't list them here, maybe point the user to the appropriate header file instead. If the list of effects change, this is the first place that everyone will forget to update :)

src/modules/src/Kconfig
ntamas: Yep, I wanted to ask about indentation policy. Some Kconfig files use mixed 2-space vs 4-space indenting, and I've also seen a few tabs occasionally here and there. I can re-tab the whole file to use spaces in this PR if that's okay, and then I'll also switch to consistent 4-space indentation.

jonasdn: That is ok, thanks.

src/deck/drivers/src/Kconfig
jonasdn: Agreed, perhaps do something similar as in this help text: https://github.com/bitcraze/crazyflie-firmware/blob/master/Kconfig#L34

tobbeanton:
> > Looks good!
> > How about some of these being selected automagicly when you select `PLATFORM_BOLT`?
> 
> As in:
> 
> ```
> config PLATFORM_BOLT
>     bool "Build for Bolt"
>     select SENSORS_BMI088_BMP388
>     select SENSORS_BMI088_SPI
>     select MOTORS_START_DISARMED
> ```
> 
> I don't know if any more configs make sense, @tobbeanton ?

This is a good start. We will have to improve and extend it with more configs as we get the Bolt out of early access. What should be set as well is DEFAULT_IDLE_THRUST=7000 and a possibility to set it. Is that available?

src/modules/src/Kconfig
knmcguire: I think that most python examples use set_value for the parameter update, which does not wait for a callback. But I think will be a few on which this will fail yes, so this should not be on by standard for sure. 

This might be an issue for the automated testing though, but let's see. We can fix that once it emerges.

ntamas:
> It is pretty dependent on the drone build and can probably vary 2000 - 10000. Better then to force the user to set it?

It definitely depends on the drone build, and for one-off custom builds it is okay to leave the default at zero and use the persistent parameter subsystem to set the appropriate value. However, if you have multiple drones of the same build, it is better to set the default in a custom firmware tailored specifically to that drone, because doing a "factory reset" (i.e. wiping the internal persistent parameter storage) would then bring the drone back to a state where the default idle thrust is set to a sensible value for that particular drone.

src/modules/src/Kconfig
ntamas: All Kconfig files were now reformatted to use 4-space indentation consistently.

src/deck/drivers/src/Kconfig
ntamas: Can you help me with finding the page to link to? Or is it okay to just link to bitcraze.io in general?

jonasdn: > Can you help me with finding the page to link to? Or is it okay to just link to bitcraze.io in general?

I think just to https://www.bitcraze.io/ is what we said, right @knmcguire, since pages can move around.

"[...] for more information about the LED ring effects see the parameter documentation on https://www.bitcraze.io/"

Maybe?

ntamas:
I have a few more configuration options in the pipeline that I'd like to see as part of the Kconfig-based system:

* `ENABLE_ONESHOT125`
* `LOW_VOLTAGE` (something like `CONFIG_PM_LOW_VOLTAGE` in a separate "Power management" section maybe?). This is useful for builds with two-cell Li-Ion batteries.
* `CRITICAL_LOW_VOLTAGE`. Same as above.

Shall I also add these in this PR or shall we get this merged first?

I also have two extra configuration options in our fork that I'd be happy to contribute upstream in separate PRs if you are interested.

One is `DISABLE_CHARGER_DETECTION`, which disables the charger detection for Bolt-based builds where the system seems to "think" that it's always connected to a USB port. Not sure if this is needed for the long-term if it's a hardware issue and if it will be fixed in later revisions.

The other one is `LED_RING_REDUCE_BRIGHTNESS`, which is used to shift the RGB values of the LED ring to the right by a given number of bits, for drone builds where the LED ring and the motors together draw too much power and would shut down the drone when both the motors and the LED are operated at full brightness.

src/deck/drivers/src/Kconfig
ntamas: Done.

jonasdn:
> I have a few more configuration options in the pipeline that I'd like to see as part of the Kconfig-based system:
> 
> * `ENABLE_ONESHOT125`
> * `LOW_VOLTAGE` (something like `CONFIG_PM_LOW_VOLTAGE` in a separate "Power management" section maybe?). This is useful for builds with two-cell Li-Ion batteries.
> * `CRITICAL_LOW_VOLTAGE`. Same as above.
> 
> Shall I also add these in this PR or shall we get this merged first?
> 
> I also have two extra configuration options in our fork that I'd be happy to contribute upstream in separate PRs if you are interested.
> 
> One is `DISABLE_CHARGER_DETECTION`, which disables the charger detection for Bolt-based builds where the system seems to "think" that it's always connected to a USB port. Not sure if this is needed for the long-term if it's a hardware issue and if it will be fixed in later revisions.
> 
> The other one is `LED_RING_REDUCE_BRIGHTNESS`, which is used to shift the RGB values of the LED ring to the right by a given number of bits, for drone builds where the LED ring and the motors together draw too much power and would shut down the drone when both the motors and the LED are operated at full brightness.

Great! Add the converting ones here!

And we would be very happy to look at the new ones from your fork!

src/modules/src/health.c
```diff
@@ -44,19 +44,20 @@
 #include "motors.h"
 #include "sensors.h"
 #include "pm.h"
+#include "autoconf.h"
 
 #include "static_mem.h"
 
-#ifndef DEFAULT_PROP_TEST_PWM_RATIO
-#  define DEFAULT_PROP_TEST_PWM_RATIO 0
+#ifndef CONFIG_MOTORS_DEFAULT_PROP_TEST_PWM_RATIO
```
ataffanel, line 51: The `#ifndef` to set default value should not be needed, kbuild will define it with a default value already

src/modules/src/power_distribution_stock.c
```diff
@@ -51,11 +52,11 @@ static struct {
   uint16_t m4;
 } motorPowerSet;
 
-#ifndef DEFAULT_IDLE_THRUST
-#define DEFAULT_IDLE_THRUST 0
+#ifndef MOTORS_DEFAULT_IDLE_THRUST
+#define MOTORS_DEFAULT_IDLE_THRUST 0
```
ataffanel, line 56: This needs to be prepended with `CONFIG_`

src/deck/drivers/src/ledring12.c
```diff
@@ -42,12 +42,13 @@
...
 #include "pulse_processor.h"
 #include "mem.h"
+#include "autoconf.h"
 
 #define DEBUG_MODULE "LED"
 #include "debug.h"
 
-#ifdef LED_RING_NBR_LEDS
-#define NBR_LEDS  LED_RING_NBR_LEDS
+#ifdef CONFIG_DECK_LEDRING_NBR_LEDS
```
ataffanel, line 50: My understanding is that since `DECK_LEDRING_NBR_LEDS` is an int, kconfig should already have declared it with the proper default value. So the `#ifdef`s should not be required anymore.

src/deck/drivers/src/ledring12.c
```diff
@@ -138,8 +139,10 @@ typedef void (*Ledring12Effect)(uint8_t buffer[][3], bool reset);
   dest[1] = ((uint16_t)G6 * 259 + 33) >> 6;                                    \
   dest[2] = ((uint16_t)B5 * 527 + 23) >> 6;
 
-#ifndef LEDRING_DEFAULT_EFFECT
-#define LEDRING_DEFAULT_EFFECT 6
+#ifdef CONFIG_DECK_LEDRING_DEFAULT_EFFECT
```
ataffanel, line 142: Same as above, the `#ifdef` should not be required and default value should in in kconfig.

ataffanel (changes requested):
Looks good! I just found one typo and have some comments about how to handle default values.

src/modules/src/power_distribution_stock.c
jonasdn: And the `#ifndef` dance is not needed, as noted below.

src/deck/drivers/src/ledring12.c
ntamas: Fixed in 0afe821

src/deck/drivers/src/ledring12.c
ntamas: Fixed in 0afe821

src/modules/src/power_distribution_stock.c
ntamas: Fixed in 0afe821

ntamas: It turns out that the `#ifndef` dance is needed because `CONFIG_MOTORS_DEFAULT_IDLE_THRUST` might be undefined if the drone is configured to start in armed mode (because in this case this option does not appear in `Kconfig`).

src/hal/src/Kconfig
```diff
@@ -1,9 +1,25 @@
-menu "IMU sensor configuration"
```
ntamas, line 1: IMU sensor configuration was removed from here because a more complete version was used from the top-level `Kconfig` file and this file was unused until now.

src/modules/src/Kconfig
```diff
@@ -62,6 +62,21 @@ endmenu
 
 menu "Motor configuration"
 
+choice
+    prompt "ESC protocol"
+    default MOTORS_ESC_PROTOCOL_STANDARD_PWM
+
+	config MOTORS_ESC_PROTOCOL_STANDARD_PWM
+	    bool "Standard PWM"
```
ntamas, line 70: I named the default protocol "standard PWM" based on [this article](https://quadmeup.com/pwm-oneshot125-oneshot42-and-multishot-comparison/). Feel free to rename if this is not the name that you would prefer.

ntamas:
> And we would be very happy to look at the new ones from your fork!

I'll wait with those PRs until this one is merged because I want to put the configuration options in the menus that were created in this PR.

jonasdn:
Merged as 7c72ef39ff8aa42d93cac7dc5b7908618b89802b

PR #970, opened 2022-02-25
docs: Add guide for creating your own platform

jonasdn:


krichardsson (commented):
Wow that was fast! 

It would also be possible to see this as a variation of a platform and only replace the default values, that is only this part.

```c
#pragma once
#define __INCLUDED_FROM_PLATFORM_DEFAULTS__
#ifdef CONFIG_PLATFORM_CF2
    #include "platform_defaults_cf2.h"
#endif
#ifdef CONFIG_PLATFORM_BOLT
   // Not exactly sure how to do this but use platform_defaults_bolt.h if RINCEWIND is not selected
    #ifdef CONFIG_PLATFORM_BOLT_RINCEWIND
        #include "platform_rincewind.h"
   #else
        #include "platform_defaults_bolt.h"
    #endif
#endif
#ifdef CONFIG_PLATFORM_TAG
    #include "platform_defaults_tag.h"
#endif

```
It would reduce the complexity and also limit the possibility of messing up settings based on the hardware such as IMU type and motorMap. I'm not sure there is a use case where the hardware conf needs to be changed?

jonasdn:
> Wow that was fast!
> 
> It would also be possible to see this as a variation of a platform and only replace the default values, that is only this part.
> 
> ```c
> #pragma once
> #define __INCLUDED_FROM_PLATFORM_DEFAULTS__
> #ifdef CONFIG_PLATFORM_CF2
>     #include "platform_defaults_cf2.h"
> #endif
> #ifdef CONFIG_PLATFORM_BOLT
>    // Not exactly sure how to do this but use platform_defaults_bolt.h if RINCEWIND is not selected
>     #ifdef CONFIG_PLATFORM_BOLT_RINCEWIND
>         #include "platform_rincewind.h"
>    #else
>         #include "platform_defaults_bolt.h"
>     #endif
> #endif
> #ifdef CONFIG_PLATFORM_TAG
>     #include "platform_defaults_tag.h"
> #endif
> ```
> 
> It would reduce the complexity and also limit the possibility of messing up settings based on the hardware such as IMU type and motorMap. I'm not sure there is a use case where the hardware conf needs to be changed?

It is not possible with how we do it right now, since `PLATFORM_BOLT` and `PLATFORM_RINCEWIND` can never both be true when they are in a `choice` in Kconfig.

jonasdn:
> > Wow that was fast!
> > It would also be possible to see this as a variation of a platform and only replace the default values, that is only this part.
> > ```c
> > #pragma once
> > #define __INCLUDED_FROM_PLATFORM_DEFAULTS__
> > #ifdef CONFIG_PLATFORM_CF2
> >     #include "platform_defaults_cf2.h"
> > #endif
> > #ifdef CONFIG_PLATFORM_BOLT
> >    // Not exactly sure how to do this but use platform_defaults_bolt.h if RINCEWIND is not selected
> >     #ifdef CONFIG_PLATFORM_BOLT_RINCEWIND
> >         #include "platform_rincewind.h"
> >    #else
> >         #include "platform_defaults_bolt.h"
> >     #endif
> > #endif
> > #ifdef CONFIG_PLATFORM_TAG
> >     #include "platform_defaults_tag.h"
> > #endif
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It would reduce the complexity and also limit the possibility of messing up settings based on the hardware such as IMU type and motorMap. I'm not sure there is a use case where the hardware conf needs to be changed?
> 
> It is not possible with how we do it right now, since `PLATFORM_BOLT` and `PLATFORM_RINCEWIND` can never both be true when they are in a `choice` in Kconfig.

Instead maybe:


```c
#pragma once

#define __INCLUDED_FROM_PLATFORM_DEFAULTS__

#ifdef CONFIG_PLATFORM_CF2
    #include "platform_defaults_cf2.h"
#endif
#ifdef CONFIG_PLATFORM_BOLT
    #include "platform_defaults_bolt.h"
#endif
#ifdef CONFIG_PLATFORM_TAG
    #include "platform_defaults_tag.h"
#endif
#ifdef CONFIG_RINCEWIND
    #include "platform_bolt.h"
    #include "platform_rincewind.h"
#endif
```

To first include bolt and then rincewind to override or add to what bolt has.

krichardsson:
Looks like a better solution to me :-)

jonasdn:
So, @ntamas the idea we have that is for changing default values of parameters, you add your own platform. For instance for stuff like the critical voltage you added. That would be better as defines in a platform file.

So if you create for instance a collmot_defconfig with PLATFORM_COLLMOT we would be able to merge that and you can build by going `make collmot_defconfig`.

tobbeanton (approved):
Looks really good!

ntamas:
> the idea we have that is for changing default values of parameters, you add your own platform

Seems OK to me, but does it mean that settings like `CONFIG_PM_BAT_LOW_VOLTAGE`, `CONFIG_PM_BAT_CRITICAL_LOW_VOLTAGE` etc (that I added in a previous PR) do not belong to the Kconfig menus any more?

jonasdn:
> > the idea we have that is for changing default values of parameters, you add your own platform
> 
> Seems OK to me, but does it mean that settings like `CONFIG_PM_BAT_LOW_VOLTAGE`, `CONFIG_PM_BAT_CRITICAL_LOW_VOLTAGE` etc (that I added in a previous PR) do not belong to the Kconfig menus any more?

Correct! @tobbeanton will change them as part of his Bolt work.

PR #971, opened 2022-02-25
Add Kconfig option to force the firmware to assume that the drone has no charger

ntamas:
This PR adds a configuration option to the firmware that forces it to assume that the drone has no charger. This is useful for the Bolt board, which has no charger, but the syslink info incorrectly reports that the drone is constantly being charged.

### DIFF
--- a/src/hal/src/Kconfig
+++ b/src/hal/src/Kconfig
@@ -22,4 +22,14 @@ config PM_BAT_CRITICAL_LOW_VOLTAGE
         time. The length of this period is configured in the code; see
         hal/interface/pm.h
 
+config PM_DISABLE_CHARGER_DETECTION
+    bool "Disable charger detection"
+    default n
+    help
+        Use this setting on boards without a charger circuit that may
+        incorrectly report that a charger is connected. Enabling this setting
+        will force the firmware to assume that no charger is connected and the
+        drone is not charging, even if the syslink info from the nRF51 chip
+        tells otherwise.
+
 endmenu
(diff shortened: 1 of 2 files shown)


tobbeanton (commented):
It all looks good but I'm thinking this might be better to fix in the nRF51 which should report false on those isPgood & isCharging flags.

tobbeanton:
I've checked the nRF51 code and it looks like it is doing the right thing:

```
bool pmUSBPower(void) {
	return pmConfig->hasCharger ? nrf_gpio_pin_read(PM_PGOOD_PIN) == 0 : 0;
}

bool pmIsCharging(void) {
  return pmConfig->hasCharger ? nrf_gpio_pin_read(PM_CHG_PIN) == 0 : 0;
}
```
@ntamas What version of nRF51 FW are you seeing the problem on the Bolt?

ntamas:
To be honest I don't know because I don't have that drone with me. We've definitely seen this last August, and I'm pretty sure that the Bolt was flashed with the latest official firmware first that was available at that time, so most likely 2021.06.

ntamas:
It looks like this was fixed in the nRF51 firmware in [this commit](https://github.com/bitcraze/crazyflie2-nrf-firmware/commit/5e76db39a42e5c32e0c1881343ee6cb54244c4cf) in Sept 2021. We've tested this in Aug 2021 the last time, so most likely the underlying problem is indeed fixed in the nRF51 firmware and all that we need to do is to re-flash the nRF51 with a newer version. Closing the PR; will reopen if it turns out that the fix is not enough.

PR #974, opened 2022-03-01
Make battery limits persistent

tobbeanton:
The battery limits are made persistent instead of kbuild configs so they are easier to change. Also the power managment automatic shutdown was added as a kconfig. Default values for the different platforms was also added in platform_defaults_*

We came to the conclusion that as a rule of thumb, build time configuration should be in kconfig and parameters that can be changed without a re-build should be persistent.

### DIFF
--- a/src/hal/src/pm_stm32f4.c
+++ b/src/hal/src/pm_stm32f4.c
@@ -524,11 +524,11 @@ PARAM_GROUP_START(pm)
 /**
  * @brief At what voltage power management will indicate low battery.
  */
-PARAM_ADD_CORE(PARAM_FLOAT | PARAM_PERSISTENT, LowVoltage, &batteryLowVoltage)
+PARAM_ADD_CORE(PARAM_FLOAT | PARAM_PERSISTENT, lowVoltage, &batteryLowVoltage)
 /**
  * @brief At what voltage power management will indicate critical low battery.
  */
-PARAM_ADD_CORE(PARAM_FLOAT | PARAM_PERSISTENT, CriticalLowVoltage, &batteryCriticalLowVoltage)
+PARAM_ADD_CORE(PARAM_FLOAT | PARAM_PERSISTENT, criticalLowVoltage, &batteryCriticalLowVoltage)
 
 PARAM_GROUP_STOP(pm)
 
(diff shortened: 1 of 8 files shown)


tobbeanton:
@ntamas Could you please look at this PR?

src/hal/src/pm_stm32f4.c
```diff
@@ -504,3 +516,19 @@ LOG_ADD_CORE(LOG_UINT8, batteryLevel, &batteryLevel)
...
 LOG_GROUP_STOP(pm)
+
+/**
+ * Power management parameters.
+ */
+PARAM_GROUP_START(pm)
+/**
+ * @brief At what voltage power management will indicate low battery.
+ */
+PARAM_ADD_CORE(PARAM_FLOAT | PARAM_PERSISTENT, LowVoltage, &batteryLowVoltage)
```
ntamas, line 527: Most of the parameters in the firmware start with a lowercase letter (even if they are camelCased). This one is uppercase, I'm not sure whether it's intentional.

ntamas (approved):
Just one minor nitpick, otherwise it looks good to me!

knmcguire:
seems like the CI is failing on this PR, but it is on loco_tdoa3_all config... which I don't think have anything to do with these changes?

src/hal/src/pm_stm32f4.c
tobbeanton: Good one, I'll change it to camel case.

src/hal/src/pm_stm32f4.c
```diff
@@ -36,12 +36,19 @@
...
 #include "ledseq.h"
 #include "commander.h"
 #include "sound.h"
 #include "deck.h"
 #include "static_mem.h"
 #include "worker.h"
+#include "platform_defaults.h"
+
+// Battery time limit conversions to ticks
+#define PM_BAT_CRITICAL_LOW_TIMEOUT   M2T(1000 * DEFAULT_BAT_LOW_DURATION_TO_TRIGGER_SEC)
```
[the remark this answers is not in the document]
tobbeanton: The defines are used so they do not trigger the low voltage by a voltage dip. The voltage needs to be below the threshold for the amount of time in DEFAULT_BAT_LOW_DURATION_TO_TRIGGER_SEC. The defines are used but that code did not change.

PR #985, opened 2022-03-08
Add support for new(), malloc(), free() and assert() in C++ app code

### LINKED ISSUE #960: Kbuild: adding C++ support?

ntamas:
This PR attempts to fix #960 for more complex C++ apps that uses classes or assertions. See the discussion in #960 for more details.

The PR extends the build system to use the C++ linker and to link to `stdc++` when the OOT app makefile indicates that it needs C++ facilities by setting `OOT_USES_CXX` to a non-empty value. The PR also provides implementations of `malloc()` and `free()` in terms of FreeRTOS functions to ensure that C++ code that was not designed with FreeRTOS in mind keeps on working even in this environment.

### DIFF
--- a/tools/kbuild/Makefile.kbuild
+++ b/tools/kbuild/Makefile.kbuild
@@ -256,6 +256,10 @@ OBJCOPY	= $(CROSS_COMPILE)objcopy
 OBJDUMP	= $(CROSS_COMPILE)objdump
 SIZE    = $(CROSS_COMPILE)size
 
+ifneq ($(OOT_USES_CXX),)
+LD		= $(CROSS_COMPILE)g++
+endif
+
 AWK		= awk
 PERL	= perl
 PYTHON	= python3
@@ -556,6 +560,10 @@ firmware-dirs	:= $(objs-y)
 firmware-objs	:= $(patsubst %,%/built-in.o, $(objs-y))
 firmware-libs	:= $(patsubst %,%/lib.a, $(libs-y))
 firmware-all	:= $(firmware-objs) $(firmware-libs)
+
+ifneq ($(OOT_USES_CXX),)
+	firmware-libs += -lstdc++
+endif
 
 quiet_cmd_firmware = LD      $@
       cmd_firmware = $(LD) $(image_LDFLAGS) $(LDFLAGS) $(firmware-objs) $(firmware-libs) -lm -o $@.elf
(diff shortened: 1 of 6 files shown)


knmcguire:
Awesome, thanks! I will check within Bitcraze who can help with the review.

tools/kbuild/Makefile.kbuild
```diff
@@ -557,6 +561,10 @@ firmware-objs	:= $(patsubst %,%/built-in.o, $(objs-y))
 firmware-libs	:= $(patsubst %,%/lib.a, $(libs-y))
 firmware-all	:= $(firmware-objs) $(firmware-libs)
 
+ifneq ($(OOT_USES_CXX),)
+	firmware-libs += -lstdc++
```
ntamas, line 565: Note to the reviewer: we need this because we also pass `-nostdlib` to the linker for plain C-based apps, so it won't link to `stdc++` by default. However, we apparently need `stdc++` for the `new` operator to work.

I also tried adding this to `LDFLAGS` instead, but then it comes too early in the command line that eventually calls the linker. It had to be placed just before where we have `-lm` there (see a few lines below).

krichardsson:
Looking good to me! 
Nice with C++ support for apps.

Our general approach to memory management so far has been to favor static allocation over dynamic, support for C++ probably leans away from the static approach, but I don't see any harm in supporting it for apps. 

I'll merge it.

Thanks @ntamas !

PR #991, opened 2022-03-09
Improved LH2 data decoding

### LINKED ISSUE #901: Lighthouse data not received in some angles
In some situations, when two sensors on the lighthouse deck are in line with the base station, no data is received by the deck.
The problem sees to be located in the decoding of data from the LH deck in `pulse_processor_v2.c`. When two sensors are aligned with the base station, the deck will not be able to figure out the channel until a third sensor is hit by the light sweep, ending up with something like this:

```
Sensor: 1  TS:c53a8f  Width: 10f  Chan: None  offset:     0  BeamWord:10b2c
Sensor: 3  TS:c53a79  Width: 125  Chan: None  offset:     0  BeamWord:1a42c
Sensor: 2  TS:c53cbb  Width: 11e  Chan: 2(1)  offset: 44731  BeamWord:1409a
Sensor: 0  TS:c53cd4  Width: 11e  Chan: 2(1)  offset:     0  BeamWord:0269a
```

The chan of sensor 1 and 3 in the example above must be back tracked by the decoder and set to 2. The current implementation only supports one step of back tracking. 

This is most likely not a problem when flying, but more common when the CF is on the ground and estimating the geometry for instance. It can be worked around by turning the CF slightly.

krichardsson:
Minor improvements in decoding of data from the FPGA

Closes #901

### DIFF
--- a/src/utils/src/lighthouse/pulse_processor_v2.c
+++ b/src/utils/src/lighthouse/pulse_processor_v2.c
@@ -151,14 +151,19 @@ TESTABLE_STATIC bool processWorkspaceBlock(const pulseProcessorFrame_t slots[], 
 TESTABLE_STATIC void augmentFramesInWorkspace(pulseProcessorV2PulseWorkspace_t* pulseWorkspace) {
     const int slotsUsed = pulseWorkspace->slotsUsed;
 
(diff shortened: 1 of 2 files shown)


evoggy:
Looks good!

PR #992, opened 2022-03-10
Updated platform.md with KBuild vs persistent param.

tobbeanton:
Updated dock to clarify when to use KBuild and when to use persistent param. It is not an easy cut and would have to be thought about for each case.


krichardsson:
Sweet!

PR #993, opened 2022-03-11
Added docs for lighthouse multi base station systems

krichardsson:
Added basic documentation on how to set up a multi base station lighthouse system.

System geometry estimation is implemented in bitcraze/crazyflie-lib-python#324


evoggy (approved):
Looks good, just fixed some small stuff.

knmcguire:
I think that this looks good too. I do feel that it needs to be in the cflib instead closer to the API docs, and perhaps refer it from here, but let's talk about that and fix that another time. For now it is okay to be here

knmcguire (approved):

PR #994, opened 2022-03-15
Added Bolt 1.1 platform config and motor mapping.

tobbeanton:
A finding made during motor output testing was that the expansion tests where not compiled in. Therefor these where added to Kbuild so the outputs could be tested.

### DIFF
--- a/src/drivers/src/motors_def_cf2.c
+++ b/src/drivers/src/motors_def_cf2.c
@@ -301,6 +301,31 @@ static const MotorPerifDef BOLT_M4_BL =
     .preloadConfig = TIM_OC4PreloadConfig,
 };
 
+// Bolt 1.1 M4, PB10, TIM2_CH3, Brushless config
+static const MotorPerifDef BOLT11_M4_BL =
+{
+    .drvType       = BRUSHLESS,
+    .gpioPerif     = RCC_AHB1Periph_GPIOB,
+    .gpioPort      = GPIOB,
+    .gpioPin       = GPIO_Pin_10,
+    .gpioPinSource = GPIO_PinSource10,
+    .gpioOType     = GPIO_OType_PP,
+    .gpioAF        = GPIO_AF_TIM2,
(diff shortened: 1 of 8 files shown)


src/drivers/src/motors_def_cf2.c
```diff
@@ -581,3 +606,14 @@ const MotorPerifDef* motorMapBoltBrushless[NBR_OF_MOTORS] =
...
 
+/**
+ * Brushless motors mapped to the Bolt 1.1 PWM outputs.
+ */
+const MotorPerifDef* motorMapBolt11Brushless[NBR_OF_MOTORS] =
+{
+  &BOLT_M1_BL,
+  &BOLT_M2_BL,
+  &BOLT_M3_BL,
+  &BOLT11_M4_BL
```
knmcguire, line 617: Why is this one called BOLT11_M4_BL while the other ones are called BOLT_M#_BL?

src/drivers/src/motors_def_cf2.c
```diff
@@ -301,6 +301,31 @@ static const MotorPerifDef BOLT_M4_BL =
     .preloadConfig = TIM_OC4PreloadConfig,
 };
 
+// Bolt 1.1 M4, PB10, TIM2_CH3, Brushless config
```
knmcguire, line 304: Hi is there only here a addition for motor M4?

knmcguire (commented):
I still saw something a bit weird in the PR. Seems like only Motor M4 was added. 

And what are the differences in the motor defines between the bolt and bolt 1.1?

src/drivers/src/motors_def_cf2.c
tobbeanton: Yes, this is the only HW change.

src/drivers/src/motors_def_cf2.c
tobbeanton: We could decouple it from the platform and call it e.g. PA1_TIM2_CH2_BL and do the same with all of them.

knmcguire (approved):

PR #995, opened 2022-03-15
Add setting to Kconfig for max number of Lighthouse base stations

evoggy:
This adds a setting to the `Expansion deck configuration` menu for setting the max number of Lighthouse base stations. This will also effect the Lighthouse V1 implementation, but using settings where the number of base stations is greater than 2 has been tested and is working.

### DIFF
--- a/docs/functional-areas/lighthouse/multi_base_stations.md
+++ b/docs/functional-areas/lighthouse/multi_base_stations.md
@@ -17,10 +17,9 @@ guide on our web to get a good understanding of basic use.
 in the python client for this task
 2. Place your base stations. They must overlap, but avoid too many in one spot, there should not be more than 4 visible
 at the same time.
-3. Re-flash the Crazyflie with support for more base stations. Start by editing the
-`src/utils/interface/lighthouse/pulse_processor.h` file in this project and change the PULSE_PROCESSOR_N_BASE_STATIONS
-to the desired number of base stations. Note: more base stations use more RAM. Build the code and flash it to the
-Crazyflie, see the documentation in this repo for instructions on building and flashing.
+3. Re-flash the Crazyflie with support for more base stations. Run `make menuconfig` and go to the `Expansion deck configuration`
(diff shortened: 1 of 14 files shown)


evoggy:
This needs a bit more work... The previous define for max number of base stations was used in code that is compiled in even if Lighthouse deck support is not enabled. This is not working out since the config is now moved to Kbuild instead.

src/modules/src/lighthouse/Kbuild
```diff
@@ -1,5 +1,5 @@
-obj-y += lighthouse_core.o
-obj-y += lighthouse_deck_flasher.o
-obj-y += lighthouse_position_est.o
-obj-y += lighthouse_storage.o
-obj-y += lighthouse_transmit.o
+obj-$(CONFIG_DECK_LIGHTHOUSE)	+= lighthouse_core.o
```
knmcguire, line 1: Some extra spaces that are not consistent with the rest of the list

knmcguire (approved):
Looks good to me! But I do think that @krichardsson should have last say in this to make sure that we have caught all defines.

knmcguire:
Ah, the CI is indicating a lot of build issues.

ntamas:
It would be great to revive the PR as we'd also need this feature in Kconfig. Unfortunately, the build logs are now gone. Any chance to re-run the CI build to see what the failures were?

evoggy:
Seems like the workflow cannot be re-run after the retention period has passed. I've merged master into this branch to update it and to trigger a new build. If I recall correctly the define was used in a test that was failing, but let's see what the CI says.

ntamas:
Thanks! It seems like there's yet another occurrence of `PULSE_PROCESSOR_N_BASE_STATIONS` that needs to be replaced. If you don't have time to work on this now, let me know and I'll send a PR on my own as I have already started working on this.

ntamas:
Awesome, thanks! (Looks like it was more complex than what I anticipated :))

PR #1000, opened 2022-03-22
release workflow for actions

knmcguire:
Started with integrating the firmware release into github actions. This used to be on Jenkins

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,13 +5,19 @@ on:
 
 jobs:
   cf2:
-   runs-on: ubuntu-latest
+    runs-on: ubuntu-latest
   
     steps:
-      - name: Checkout Repo
-        uses: actions/checkout@v2
-        with:
-          submodules: true
+    - name: Checkout Repo
+      uses: actions/checkout@v2
+      with:
+        submodules: true
 
-      - name: build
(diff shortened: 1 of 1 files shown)


knmcguire:
Currently I don't have the DFU file in as it is unsure if actions also support an upload of that file, but perhaps it can work with the same asset type as with  the bin file. 

The idea is, is that the same steps will be done for the other platforms as well (building and uploading assets)

krichardsson:
Looking good! 

Is the `- name: Upload Build Artifact` really required? Looks like the artifacts are uploaded to the release later

Is the plan to add some sort of matrix build with cf2, tag and bolt?

knmcguire:
Ah yeah ofcourse, we dont that extra upload here. That is being used by crazy flie testing CI at ci.yml. I was planning to simply copy paste it like in the build ci, but using a matrix sounds cleaner. I'll try it!

Btw @krichardsson were you able to try out the action out before officially merging? Or does it need to be fully in master for us to try it out

knmcguire:
Got the idea to seperate the release and upload bin files from this issue: https://github.com/actions/create-release/issues/14#issuecomment-555379810

krichardsson:
Looks good! 
It seems as github thinks something is wrong on line 56 though, https://github.com/bitcraze/crazyflie-firmware/actions/runs/2023276547

knmcguire:
ahhh so it does check the yml workflow file everytime. interesting! I've fixed it so let's see

Issue #915, opened 2022-01-14
SD card config.txt full comment line parsing


dominiknatter:
Hi everyone,

I'm currently trying to ramp up logging via the SD deck. I get to the point where log files are created on the SD card using cfclient to start and end the logging. However, when analyzing the log files using your tools (cfusdlog.py) it seems like only timestamps are being logged. We use a slightly individually adapted firmware, but those changes should mainly regard the UWB code. I've already removed all the other decks to avoid interference via the SPI. Here is the config.txt file, the console output, and the resulting log file:

config.txt

```
1     # version
512   # buffer size in bytes
sd_log   # file name
0     # enable on startup (0/1)
on:fixedFrequency
1000     # frequency
1     # mode (0: disabled, 1: synchronous stabilizer, 2: asynchronous)
# IMU
acc.x
acc.y
acc.z
gyro.x
gyro.y
gyro.z
```

console output (... replaces some irrelevant output from our own firmware):
```
SYS: Crazyflie 2.1 is up and running!
SYS: Build 399:29116e6eee76 (2020.06 +399) CLEAN
SYS: I am 0x20303743575350170023001F and I have 1024KB of flash!
CFGBLK: v1, verification [OK]
DECK_CORE: 1 deck(s) found
DECK_CORE: Calling INIT on driver bcUSD for deck 0
uSD: mount SD-Card [OK].
IMU: BMI088: Using I2C interface.
uSD: wait for sensors
IMU: BMI088 Gyro connection [OK].
IMU: BMI088 Accel connection [OK]
IMU: BMP388 I2C connection [OK]
ESTIMATOR: Using Complementary (1) estimator
CONTROLLER: Using PID (1) controller
MTR-DRV: Using brushed motor driver
EEPROM: I2C connection [OK].
STORAGE: Storage check [OK].
IMU: BMI088 gyro self-test [OK]
DECK_CORE: Deck 0 test [OK].
STAB: Wait for sensor calibration...
SYS: Free heap: 18808 bytes
...
STAB: Ready to fly.
uSD: Config read [OK].
uSD: malloc buffer 512 bytes uSD: [OK].
uSD: Logging to: sd_log08
uSD: Wrote 1315898 B to: sd_log08 (131587 of 133000 events)
```

log file (I had to add .txt in order to upload it on GitHub, but it's a binary without file extension):
[sd_log08.txt](https://github.com/bitcraze/crazyflie-firmware/files/7872025/sd_log08.txt)

[...]

PR #1004, opened 2022-03-23
Update usddeck.c

roeeSch:
Support config.txt files with lines which are pure comments such as the first line here:
```
# Config Vesion 1.2.3
100   # frequency
3    # buffer size
...
gyro.x
gyro.y
gyro.z
acc.x
```

### DIFF
--- a/src/deck/drivers/src/usddeck.c
+++ b/src/deck/drivers/src/usddeck.c
@@ -411,6 +411,7 @@ TCHAR* f_gets_without_comments (
   TCHAR c, *p = buff;
   UINT rc;
   bool isComment = false;
+  bool isPureComment = false;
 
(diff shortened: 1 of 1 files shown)


knmcguire:
Ah great, thanks!

This will fix this following issue #915? I wanted to look at this for a triage meeting but perhaps not necessary anymore!

We are in the middle or a release so we might wait a bit with merging this, but in the mean time @tobbeanton could take a look at it!

roeeSch:
@knmcguire, I think this will fix #915.
I verified this functionality by updating the firmware and adding a comment to config.txt and checking the logs on the SD.

krichardsson:
Merged
Thanks!

krichardsson:
Fixes #915

PR #1009, opened 2022-03-25
Blocking GAP8 reading to avoid printing 0 in console

evoggy:
There was a timeout on the GAP8 UART reception which caused 0 to be printed in the Crazyflie client console, the return was not checked. This replaces it with an infinite timeout.


knmcguire:
I guess you mean finite timeout? looks good!

knmcguire (approved):

evoggy:
@knmcguire There used to be a default timeout but this replaced it with the special timeout that means blocking (i.e no timeout). Thanks for having a look!

PR #1020, opened 2022-04-14
CPX documentation

evoggy:
This documents the inner workings of CPX.

### DIFF
--- a/docs/_data/menu.yml
+++ b/docs/_data/menu.yml
@@ -12,6 +12,7 @@
 - title: Functional areas
   subs:
     - {page_id: crtp_index}
+    - {page_id: cpx_index}
     - {page_id: memory_subsystem}
     - {page_id: stabilizer_index}
     - {page_id: deck_memory_format}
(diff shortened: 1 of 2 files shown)


evoggy:
This is still work in progress, but I'll leave it here so we can have a look at it. It also contains some stuff that's not true yet, like CPX over CRTP. Let's see how this progresses otherwise we remove it :smile: 

Should we also add some information about the APIs at the end (like structs and C/Python APIs) or this should be in each of the repositories where this is implemented?

krichardsson (commented):
Looks good!

Diitaa is supported for repo docs if you want fancier diagrams https://www.bitcraze.io/development/contribute/repo-doc-guidelines/#diagrams

evoggy:
Nice, forgot about that! Updated the diagrams now to use ditaa instead.

The MTU of a packet which will not be split is still TDB. What about setting this to 30 bytes?

evoggy:
30 bytes was decided for the MTU.

PR #1023, opened 2022-04-14
Python bindings for power distribution

whoenig:
The current firmware code mixes power distribution and motor control,
so this change also includes refactoring for improved separation:

* powerDistribution() now outputs motors_thrust_t
* All low-level motor functionality (including logging and motorSet
  params) are moved to motors.c
* powerStop() is renamed to motorsStop() to reflect the source module
* stabilizer initializes the motors and has a more verbose call for the
  actual actuation

### DIFF
--- a/bindings/cffirmware.i
+++ b/bindings/cffirmware.i
@@ -18,6 +18,7 @@
 #include "filter.h"
 #include "num.h"
 #include "controller_mellinger.h"
+#include "power_distribution.h"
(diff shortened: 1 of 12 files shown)


knmcguire (approved):
#1024 mentions the enabling of more than 4 motors. Hopefully by that time, the python bindings are already part of the CI and we are aware if anything goes wrong.

PR #1027, opened 2022-04-20
Add assert dump functionality

krichardsson:
After a faild assert, the Crazyflie restarts and information about the assert is  logged to the console. In some cases when a client already was connected to the Crazyflie before the assert, it will re-connect and consume the logs. This makes it hard to debug as the assert infomation is lost.

This PR adds functionality to trigger a dump of the assert information again to the console log. This way a client can re-connect to the Crazyflie and trigger a dump for post-mortum investigations.


evoggy (approved):
Looks good if the power on/off will clear the flag.

PR #1037, opened 2022-05-03
Kalman: only initialise param vars one time

### LINKED ISSUE #1014: Position estimate can no longer be initialized to non-zero value
It used to be possible to initialize / overwrite the drones estimated position by setting the parameters
```
kalman.initialX
kalman.initialY
kalman.initialZ
```
before setting 
`kalman.resetEstimation`
(see also describing [forum entry](url))

Commit d8d4878e4a1ebe79d7e1bbef24d81ec0c95589f9 unfortunatley 'destroyed' this behaviour.

In [estimator_kalman.c:437](https://github.com/bitcraze/crazyflie-firmware/blob/d8d4878e4a1ebe79d7e1bbef24d81ec0c95589f9/src/modules/src/estimator_kalman.c#L437) `kalmanCoreDefaultParams()` now ignores these values and uses zero instead:
https://github.com/bitcraze/crazyflie-firmware/blob/d8d4878e4a1ebe79d7e1bbef24d81ec0c95589f9/src/modules/src/estimator_kalman.c#L437
https://github.com/bitcraze/crazyflie-firmware/blob/d8d4878e4a1ebe79d7e1bbef24d81ec0c95589f9/src/modules/src/kalman_core/kalman_core.c#L125-L153

It would be nice to either restore the behaviour via the parameters or have another possibility to initialize a non-zero position

krichardsson:
The parameters used in kalman core were extracted into a struct in #869.
The variables used for the parameters are reset to default values when the kalman estimator is reset, and this causes two problems:
1. Parameter values that were set earlier are lost, for instance the starting point used in the Position commander (see #1014)
2. The contract for parameters is that the underlying variable should not be changed directly by the firmware code, as the client can not detect the change. Re-initializing the parameters from time to time breaks this contract.

In this PR the initialization of the parameter data is moved from `estimatorKalmanInit()` to `estimatorKalmanTaskInit()` to make sure it is only done one time.

Fixes #1014

### DIFF
--- a/src/modules/interface/kalman_core/kalman_core.h
+++ b/src/modules/interface/kalman_core/kalman_core.h
@@ -141,7 +141,7 @@ void kalmanCoreUpdateWithBaro(kalmanCoreData_t *this, const kalmanCoreParams_t *
  *
  * The filter progresses as:
  *  - Predicting the current state forward */
-void kalmanCorePredict(kalmanCoreData_t *this, const kalmanCoreParams_t *params, Axis3f *acc, Axis3f *gyro, float dt, bool quadIsFlying);
+void kalmanCorePredict(kalmanCoreData_t *this, Axis3f *acc, Axis3f *gyro, float dt, bool quadIsFlying);
 
 void kalmanCoreAddProcessNoise(kalmanCoreData_t *this, const kalmanCoreParams_t *params, float dt);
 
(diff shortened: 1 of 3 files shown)


tobbeanton (approved):
Good change.

krichardsson:
@whoenig We decided to merge this even though you did not review it, please let us know if this causes any issues for the python bindings

PR #1040, opened 2022-05-09
Update DFU flashing and new docs

C-Glick:
From issue #923 

With this update I've fixed a few issues with the automatic DFU flashing. 
- I add the $(srctree) variable to python script call
- I added a `flash_dfu_manual` make target to flash the firmware assuming the crazyflie is already in DFU mode. This makes it easier for users to flash using DFU mode when the crazyflie cannot run the normal firmware to support the automatic reboot into DFU mode
- Finally, I updated the documentation to reflect the new make targets. I decided to leave the warning at the top of the documentation as DFU mode can still be dangerous for new/inexperienced users.

One issue that I've noted is dfu-util complains about an invalid suffix signature. This doesn't seem to cause any issues now but could in the future.

```
dfu-util: Invalid DFU suffix signature
dfu-util: A valid DFU suffix will be required in a future dfu-util release!!!
```

I've also noticed that the make file complains about old recipes for flash_dfu, this may be due to the .PHONY target, I'm not familiar with what its purpose is so I left it as is.

Let me know if any changes need to be made before merging. Thanks!

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -3,5 +3,8 @@
         "**/generated": true,
         "**/node_modules": true,
         "**/bower_components": true
+    },
+    "files.associations": {
+        "bootloader.h": "c"
     }
 }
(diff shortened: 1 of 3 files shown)


tobbeanton (approved):
I think it all looks good, many thanks!

The warnings:
```
dfu-util: Warning: Invalid DFU suffix signature
dfu-util: A valid DFU suffix will be required in a future dfu-util release!!!
```
I think is because we are flashing a binary file.

Issue #1047, opened 2022-05-16
Crazyflie 2.1 listed as 2.0 in windows


knmcguire:
When connecting a crazyflie to a windows computer, the USB device is listed as 2.0 even if it is a 2.1.

tobbeanton:
Fixed in https://github.com/bitcraze/crazyflie-firmware/pull/1050

PR #1050, opened 2022-05-20
Implementation of Virtual Com Port (composite USB) and DSHOT motor signal protocol

tobbeanton:
This is a pretty hefty PR but it was hard to break it down...

- ST USB device driver updated to V2.2.1
- Virtual com port (VCP) has been implemented alongside the Crazyflie interface making a  composite USB device.
- VCP is used to support BLHeli 4-way passthough interface. Works with the BLHeliConfiguratior for windows (in wine in linux).
- DSHOT implementation for CF2.1-BL prototype and Bolt 1.1
- 4way protocol and logic adapted from Betaflight.
- FreeRTOS task notifications enabled and used for 4-wire interface.
- 4way task created with max priority since the half-duplex uart protocol used (AVRootloader) to communicate with the ESCs is bit-banged.
- Crazyflie 2.1 Brushless config added to Crazyflie platform.

Documentation is still lacking and will be made when things are a bit more clear and figured out as part of getting Bolt out of early access.

The crazyflie-lib-python cfusb driver needs to be new for the cfclient to be able to connect using USB.


knmcguire:
I'm getting a build error for the cfbl.

```
make cfbl_defconfig
make -j16
```

Results in this error:
```
../src/drivers/src/motors.c: In function 'motorsPrepareDshot':
../src/drivers/src/motors.c:403:5: error: 'dmaWait' undeclared (first use in this function)
  403 |     dmaWait++;
      |     ^~~~~~~
../src/drivers/src/motors.c:403:5: note: each undeclared identifier is reported only once for each function it appears in
../src/drivers/src/motors.c: At top level:
cc1: error: unrecognized command line option '-Wno-stringop-overread' [-Werror]
cc1: all warnings being treated as errors
make[4]: *** [../scripts/Makefile.build:190: src/drivers/src/motors.o] Error 1
make[3]: *** [../scripts/Makefile.build:311: src/drivers/src] Error 2
make[2]: *** [../scripts/Makefile.build:311: src/drivers] Error 2
make[2]: *** Waiting for unfinished jobs....
```

The CI doesn't use this config yet so that is why it didn't catch this

configs/cfbl_defconfig
```diff
@@ -0,0 +1,5 @@
+CONFIG_MOTORS_ESC_PROTOCOL_DSHOT=y
```
knmcguire, line 1: the configfile is called cfbl_defconfig eventhough the platform calls it cf21b. Should we try to uniform it somehow?

src/drivers/interface/motors.h
```diff
@@ -95,6 +95,26 @@
   #define MOTORS_BL_PWM_PERIOD         MOTORS_BL_PWM_CNT_FOR_PERIOD
   #define MOTORS_BL_PWM_PRESCALE       (uint16_t)(MOTORS_BL_PWM_PRESCALE_RAW - 1)
   #define MOTORS_BL_POLARITY           TIM_OCPolarity_Low
+#elif defined(CONFIG_MOTORS_ESC_PROTOCOL_DSHOT)
+/**
+ * *VARNING* Make sure the brushless driver is configured correctly as on the Crazyflie with normal
```
knmcguire, line 100: typo: Warning

src/drivers/interface/motors.h
```diff
@@ -95,6 +95,26 @@
...
+ * brushed motors connected they can turn on at full speed when it is powered on!
+ *
+ */
+//  #define MOTORS_BL_PWM_PERIOD         (TIM_CLOCK_HZ / 150000) // 150kHz bitrate DHSOT150
```
knmcguire, line 104: do we keep the uncommented values here? Shouldn't it be part of kbuild menuconfig if these need to be configured?

src/hal/src/usbd_desc.c
```diff
@@ -68,7 +69,7 @@
 #define USBD_PRODUCT_HS_STRING          "STM32 Virtual ComPort in HS mode"
 #define USBD_SERIALNUMBER_HS_STRING     "00000000050B"
 
-#define USBD_PRODUCT_FS_STRING          "Crazyflie 2.0"
+#define USBD_PRODUCT_FS_STRING          "Crazyflie 2.X"
```
knmcguire, line 72: I guess this will fix #1047 ?

knmcguire (commented):
It's a lot... and I won't be able to review it all, for the communication part @evoggy is probably the best. 

Maybe we need to do release testing width for the size of this PR

src/drivers/src/uart_syslink.c
```diff
@@ -56,7 +56,7 @@ static StaticSemaphore_t waitUntilSendDoneBuffer;
 static xSemaphoreHandle uartBusy;
 static StaticSemaphore_t uartBusyBuffer;
 static xQueueHandle syslinkPacketDelivery;
-STATIC_MEM_QUEUE_ALLOC(syslinkPacketDelivery, 8, sizeof(SyslinkPacket));
+STATIC_MEM_QUEUE_ALLOC(syslinkPacketDelivery, 16, sizeof(SyslinkPacket));
```
knmcguire, line 59: Not sure if this will affect the rest of the STM at all?

src/hal/src/usbd_desc.c
tobbeanton: Yes

src/drivers/src/uart_syslink.c
tobbeanton: I will revert this since the increased queue size was to avoid asserts but this is handled by pausing interrupts instead which works well.

PR #1052, opened 2022-05-25
CI testing dynamic matrix

knmcguire:
I just created this PR as I thought it might be a little less annoying to constantly change in master in order to test out the CI actions. let's see how it goes!


knmcguire:
ah never mind... it doesn't really matter.

Issue #1054, opened 2022-06-02
Platform build configurations seems to be broken


krichardsson:
The default kbuild konfigurations seems to be a bit off and needs a re-check

The way estimators are set does not work
Current: CONFIG_ESTIMATOR=any
Should be: ESTIMATOR=any

CONFIG_POWER_DISTRIBUTION does not work at all

Deck force does not work
Current: CONFIG_DECK_FORCE=bcDWM1000
Should be: CONFIG_DECK_FORCE="bcDWM1000"

Since the tag def config is wrong, the road runners do not have the Loco deck enabled by default

krichardsson:
Fixed in #1056

PR #1056, opened 2022-06-08
Kconfig updates

krichardsson:
This PR fixes some issues in the Kconfig setup found in #1054


krichardsson:
For the estimator and controller settings the changes are pretty straight forward, simply set the appropriate estimator or controller config to `y` in the default config files. For instance
```
CONFIG_ESTIMATOR_ANY=y
```

If the kalman estimator is required, one would use
```
CONFIG_ESTIMATOR_KALMAN=y
```

krichardsson:
I changed the power distribution a bit. I could not figure out a way to use kbuild in a similar way as we did earlier (with a config the names the file to use in the build). Instead I added a power distribution choice to the configuration that only contains the normal quad power distribution. If someone wants a new power distribution for a new platform, a new option can be added to the choice that is used to build the correct file. The setup is documented in the docs files.

tobbeanton (approved):
Nothing to common on really, looks nice!

PR #1057, opened 2022-06-09
Autonomous swarm demo update and working 

marios-stam:
Fixed `set.value(trajcount)` timeout error  and removed not needed lighthouse stations configuration


krichardsson:
Sweet!

Issue #998, opened 2022-03-18
Channel/address not always set at start up


krichardsson:
I have seen that the channel and/or address is not always set at start up. This problem does not happen often and is not easily repeatable.

Address and channel is sent from the STM to the NRF here https://github.com/bitcraze/crazyflie-firmware/blob/master/src/hal/src/radiolink.c#L99-L101 as syslink packets

Maybe it is possible that one or more of the packets is lost?

I have seen cases where no data is set (the NRF is still on the default 80/0xE7E7E7E7E7), but also where the channel is set but not the address.

Maybe #703 will solve this problem as well?

krichardsson:
This happens in the stab lab from time to time. 

It seems as sometimes only one of the channel or address is set

krichardsson:
We are note sure what is causing this problem but one suspect is the syslink. The STM is sending 3 syslink messages to the NRF early in the start up phase, with the radio channel, address and rate. If one or more of theses messages are lost we could get the symptoms we see.

The first step to solve this is to add a debug probe in the NRF to collect information that will show if the messages arrived or not in the NRF. Functionality is also added to the STM to request the information and dump it to the console. Since the issue is observed in the stab lab from time to time, we will hopefully be able to inspect a Crazyflie where the problem has occurred.

The probe is implemented in #1061 and bitcraze/crazyflie2-nrf-firmware#71

krichardsson:
Finally we got a CF ending up with this problem in the stab lab after the probe was installed.

The result from the NRF is:
```
SL: Syslink debug probe:
SL: Address received: 0
SL: Chan received: 1
SL: Rate received: 0
SL: Dropped: 0
```

Looking at the code in [radiolink.c](https://github.com/bitcraze/crazyflie-firmware/blob/master/src/hal/src/radiolink.c#L99-L101) we see

``` C
  radiolinkSetChannel(configblockGetRadioChannel());
  radiolinkSetDatarate(configblockGetRadioSpeed());
  radiolinkSetAddress(configblockGetRadioAddress());
```

This means that the two last messages were lost for some reason.

krichardsson:
Another example from the stab lab shows the same pattern:
```
SL: Syslink debug probe:
SL: Address received: 0
SL: Chan received: 1
SL: Rate received: 0
SL: Dropped: 0
```

PR #1061, opened 2022-06-10
Add debug probe for radio setup

krichardsson:
Add code to collect debug data for #998.
Added functionality to request and dump data from the NRF that can be used on a CF that has a half configured radio.

We do not know what is causing this problem and this is a first step to collect data. This code should probably be removed when the problem is solved.


krichardsson:
Paired with https://github.com/bitcraze/crazyflie2-nrf-firmware/pull/71

tobbeanton (approved):

PR #1064, opened 2022-06-20
Enable setting UART1 debug printing baudrate from Kconfig

evoggy:
This change enables setting the UART1 baudrate from Kconfig. The default is still 115200.


krichardsson:
Nice!

PR #1068, opened 2022-06-23
Tunnel CRTP over CPX (WiFi and serial port)

### LINKED ISSUE #1065: Optimize CPX communication on UART2
The AI deck uses UART2 to communicate with the STM32 when using CPX. Unfortunately the current implementation of reading/writing the UART is too slow and this causes issues (like asserting in syslink).

In the current implementation we're busy sending (i.e we're looping until TXE) and when receiving we're queuing one byte at a time. Ideally this would be moved to DMA instead, but since this conflicts with both the LED-ring and DSHOT on the Bolt/BL this is not a good option. Instead we should rewrite it to use interrupts for sending data buffers and assemble full CPX packets in the ISR before queuing them and switch task to CPX.

evoggy:
These changes enables CRTP tunneling over CPX. It can either be used over WiFi or over the UART2 serial port on the expansion connector. It also improves CPX stability.

* Increased UART2 interrupt priority
* Allow debug printouts on UART1 together with AI deck (STM32->GAP8 UART now disabled by default)
* Improve UART2 performance (read/write is now done in interrupts instead of busy looping RXNE/TXE
* Break out CPX and transports from AI deck and make generic
* Increased UART2 baud rate to 576000

Note, this removes the older functionality of using syslink over UART2 and replaces it with using CRTP-over-CPX over UART2. This change will not effect the application using the library, but it does require an update of the crazyflie-lib-python where the serial driver uses CPX instead.

Requires updated crazyflie-lib-python with PR https://github.com/bitcraze/crazyflie-lib-python/pull/342
Requires updated aideck-esp-firmware with PR https://github.com/bitcraze/aideck-esp-firmware/pull/16

This closes #1065.


src/deck/drivers/src/aideck.c
```diff
@@ -554,7 +353,7 @@ static const DeckDriver aideck_deck = {
     .name = "bcAI",
 
     .usedGpio = DECK_USING_IO_4,
-    .usedPeriph = DECK_USING_UART1,
+    //.usedPeriph = DECK_USING_UART1,
```
knmcguire, line 356: shouldn't this be completely removed or actually be DECK_USING_UART2 ?

src/drivers/src/uart2.c
```diff
@@ -242,134 +223,53 @@ int uart2Putchar(int ch)
...
 
-bool uart2GetDataWithDefaultTimeout(uint8_t *c)
-{
-  return uart2GetDataWithTimeout(c, UART2_DATA_TIMEOUT_TICKS);
-}
+int uart2GetData(size_t size, uint8_t * buffer) {
+  size_t sizeLeft = size;
+  while (sizeLeft > 0) {
+    xStreamBufferSetTriggerLevel(rxStream, sizeLeft);
+    // TODO: Investigate why this loop is needed?
```
knmcguire, line 267: let's make an issue for this

src/hal/src/Kconfig
```diff
@@ -21,21 +21,18 @@ config SYSLINK_RX_DMA
...
-config CRTP_OVER_UART2
-    bool "Move CRTP communication to UART2 instead of over radio"
-    default n
-    help
-        The CRTP (Crazy Real Time Protocol) communication is normally
-        sent over the radio link (via Syslink). By enabling this the
-        communication will instead be sent over UART2 to communicate
-        with e.g. a companion computer instead.
+config ENABLE_CPX
+  bool "Enable CPX on UART2"
```
knmcguire, line 25: Does this only CPX for UART2? or is the entire stack enabled? Perhaps this should be a bit clearer

src/modules/src/cpx/cpx.c
```diff
@@ -0,0 +1,107 @@
...
+#include "semphr.h"
+#include "queue.h"
+#include "event_groups.h"
+#include "debug.h"
+#include "system.h"
+#include "radiolink.h"
+
+#include "cpxlink.h"
+#include "cpx_internal_router.h"
+#include "aideck.h"
```
knmcguire, line 17: Ideally we want to remove this dependency. let's make an issue for this

src/modules/src/cpx/cpx_external_router.c
```diff
@@ -0,0 +1,136 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
```
knmcguire, line 24: Perhaps an explanation of this file?

src/modules/src/cpx/cpx_internal_router.c
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
```
knmcguire, line 24: same as above, maybe just one line of comment to explain the functionality

src/modules/src/cpx/cpxlink.c
```diff
@@ -0,0 +1,127 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ * syslink.c: nRF24L01 implementation of the CRTP link
```
knmcguire, line 24: copy paste. Probably should be different

src/hal/src/Kconfig
evoggy: This has been split up on two configs, but one forces the other so it's the same behavior as before but more clear.

knmcguire:
@s9meboeh this removes the old uart2_com functionality and replaces it with the new CPX framework, just that you are aware of this.

s9meboeh:
> 

Thank you for your ping. I see if we can have a look into updating our functionality for the new framework.

PR #1072, opened 2022-06-30
Specify wifi password minimum size

Laatikk01234:
If given smaller password than 8 the esp32 on ai-deck appears to die silently.

Or not initialize properly.


evoggy:
Thanks for the change and reporting this! It would be great if we could set the allowed length of the string in Kconfig, but as far as I can see this is not supported. I've opened an issue for handling passwords with < 8 chars gracefully in the ESP (https://github.com/bitcraze/aideck-esp-firmware/issues/15)

PR #1075, opened 2022-07-05
convert alternative pins loco deck to kbuild

knmcguire:
The loco deck alternative pins setting was not yet configured to kbuild.

I also noticed that somewhere the define had a typo initially, so probably was broken anyway.


src/deck/drivers/src/locodeck.c
```diff
@@ -569,7 +569,7 @@ static const DeckDriver dwm1000_deck = {
   .pid = 0x06,
   .name = "bcDWM1000",
 
-#ifdef LOCODEC_USE_ALT_PINS
```
knmcguire, line 572: this was the typo I was referring about

tobbeanton (approved):
Nice!

PR #1077, opened 2022-07-07
Improved internal parameter set/get functions

### LINKED ISSUE #1073: paramSetFloat() seems to be broken for persistent parameters
It seems like paramSetFloat() does not work when it is invoked for persistent parameters because the `(params[varid.index].type & (~PARAM_CORE)) == PARAM_FLOAT` condition does not hold if the `PARAM_EXTENDED` bit is set.

I worked around it with the following patch, but it's unclear to me whether this is the right thing to do as this would happily attempt to set an int parameter with a `float` value bit-by-bit; on the other hand, `paramSetInt()` does the same thing the opposite way (allowing the user to set a float parameter with an int value, again bit-by-bit). Furthermore, both variants allow the user to overwrite parameters marked as read only.

```patch
--- a/src/modules/src/param_logic.c
+++ b/src/modules/src/param_logic.c
@@ -581,14 +581,9 @@ void paramSetFloat(paramVarId_t varid, float valuef)
   pk.data[0] = MISC_VALUE_UPDATED;
   pk.data[1] = varid.id & 0xffu;
   pk.data[2] = (varid.id >> 8) & 0xffu;
-  pk.size=3;
-
-  if ((params[varid.index].type & (~PARAM_CORE)) == PARAM_FLOAT) {
-      *(float *)params[varid.index].address = valuef;
+  pk.size = 3;

-      memcpy(&pk.data[2], &valuef, 4);
-      pk.size += 4;
-  }
+  pk.size += paramSet(varid.index, (void *)&valuef);
```

Ultimately, it should be decided where the type and readonly checks belong; if they do not belong to `paramSetInt()` and `paramSetFloat()`, then the patch above would work, but it then allows the user to do all sorts of nasty things to parameters from within an app. If they do belong there, then the checks should be added so my patch is not adequate.

tobbeanton:
As a first step this PR will fix #1073 but it will need more work as the functions to internally set/get params need a clearer definition and error handling.

The checks to evaluate the type of param has been moved to an assert and made simpler. For the future I'm not so sure asserts are such a good idea but this would have to be part of a bigger redesign of the firmware. I am now only checking the type (PARAM_TYPE_INT and PARAM_TYPE_FLOAT) and if it is setting a param with PARAM_RONLY.

### DIFF
--- a/src/modules/interface/param_logic.h
+++ b/src/modules/interface/param_logic.h
@@ -114,24 +114,14 @@ int paramGetInt(paramVarId_t varid);
  */
 unsigned int paramGetUint(paramVarId_t varid);
 
-/**
- * Set param with [index] to data
- *
- * @param index  The param index
- * @param data  The variable data
- *
- * @return number of bytes set
- **/
-int paramSet(uint16_t index, void *data);
-
-/** Set int value of a parameter
+/** Set int value of an int parameter (1-4 bytes)
  *
  * @param varId variable ID, returned by paramGetVarId()
  * @param valuei Value to set in the variable
  */
 void paramSetInt(paramVarId_t varid, int valuei);
 
-/** Set float value of a parameter
+/** Set float value of a float parameter
  *
  * @param varId variable ID, returned by paramGetVarId()
  * @param valuef Value to set in the variable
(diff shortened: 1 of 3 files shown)


ntamas:
Looks good to me!

PR #1088, opened 2022-07-29
multiranger discard range measurements larger with bad status

### LINKED ISSUE #874: Multiranger deck gives odd values when out of range
Reported in https://forum.bitcraze.io/viewtopic.php?f=6&t=5086

We have also seen odd behavior when the multiranger sensors are out of range. A user probably expects some know value (or other mechanism) to be able to distinguish out of range from real distance measurements.

I think the Z-ranger behaves differently and reports a fixed distance when out of range. The sensor has a wide range of configurations and the Z-ranger is not set up in the same way as the multi ranger (short-range VS long-range).

knmcguire:
This should fix #874. Here in the multiranger, it will first check if the vl53l1 returns a valid range measurements status, and if not it will output the maximum value of an int16_t

Two questions here: 

1. Should I do this also for the z-ranger/flowdeck?
2. Do we want to still output the measurement but just include the ranging status in the logging variable?

### DIFF
--- a/src/deck/drivers/src/multiranger.c
+++ b/src/deck/drivers/src/multiranger.c
@@ -46,6 +46,7 @@
 static bool isInit = false;
 static bool isTested = false;
 static bool isPassed = false;
+static uint16_t filterMask = 1 << VL53L1_RANGESTATUS_RANGE_VALID;
 
 #define MR_PIN_UP PCA95X4_P0
 #define MR_PIN_FRONT PCA95X4_P4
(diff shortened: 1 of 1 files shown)


src/deck/drivers/src/multiranger.c
```diff
@@ -96,7 +96,11 @@ static uint16_t mrGetMeasurementAndRestart(VL53L1_Dev_t *dev)
     }
 
     status = VL53L1_GetRangingMeasurementData(dev, &rangingData);
-    range = rangingData.RangeMilliMeter;
```
knmcguire, line 99: this is the most important change, the rest is just formatting

knmcguire:
with discussion with @krichardsson, we decided to add a parameter which presents a filter flag, that indicates for which statuses the measurement, if it should let the measurement through or not, based on these values: https://github.com/bitcraze/crazyflie-firmware/blob/46f497a1ef058750f59277c88578f9f2c13ea670/src/lib/vl53l1/core/inc/vl53l1_def.h#L502-L535

krichardsson (approved):
Sweet!

PR #1089, opened 2022-08-02
Fixed estimator position stuck when no connection to client

marios-stam:
Calling ```paramSetInt()``` and ```paramSetFloat()``` while having no connection to the client established leads to the output queue being filled up and the position being stuck after the resetting of the Kalman estimator. This can be handled by setting the CONFIG_PARAM_SILENT_UPDATES flag and the user should be careful when using these functions.


krichardsson:
This PR addresses the problem of locking the kalman estimator if the estimator is reset multiple times when not connected to a client. This could, for instance happen if no positioning system was present which let to the kalman supervisor resetting the kalman filter repeatedly.

PR #1092, opened 2022-08-11
Multi bs non experimental

knmcguire:
This PR does two things

- Updating the default amount of basestations to 4
- Triming the dev documentation and just give instructions to have more than 4 basestations


knmcguire:
Firmware compiled with 4:
```
Version extracted from git
Flash |  253156/1032192 (25%),  779036 free | text: 245808, data: 7348, ccmdata: 0
RAM   |   83336/131072  (64%),   47736 free | bss: 75988, data: 7348
CCM   |   59408/65536   (91%),    6128 free | ccmbss: 59408, ccmdata: 0
```

compared to when compiling for 2:
```
Flash |  253156/1032192 (25%),  779036 free | text: 245808, data: 7348, ccmdata: 0
RAM   |   82400/131072  (63%),   48672 free | bss: 75052, data: 7348
CCM   |   59408/65536   (91%),    6128 free | ccmbss: 59408, ccmdata: 0
```

krichardsson (commented):
My only worry is what happens in a LH1 system that only supports 2 base stations. 
* Does the code work for LH1 when set to 4 base stations?
* I suppose the client will show 4 base stations also for LH1, is that OK?

knmcguire:
This is what it looks like in the cfclient:
image

knmcguire:
If we would need to change it further in the firmware, it will be either changing all DECK_LIGHTHOUSE_MAX_N_BS to a variable or we do something with the baseStationAvailabledMap parameter?

https://github.com/bitcraze/crazyflie-firmware/blob/59044cecc5b4e263226968bb6362d5a150560554/src/modules/src/lighthouse/lighthouse_core.c#L846

The problem is that it is only being set and used in the lighthouseCoreInit()
```
void lighthouseCoreInit() {
  lighthouseStorageInitializeSystemTypeFromStorage();
  lighthousePositionEstInit();

  for (int i = 0; i < CONFIG_DECK_LIGHTHOUSE_MAX_N_BS; i++) {
    modifyBit(&baseStationAvailabledMap, i, true);
  }
}
```

Maybe we can continously modify the bit everytime the lighthouse type switched:

https://github.com/bitcraze/crazyflie-firmware/blob/59044cecc5b4e263226968bb6362d5a150560554/src/modules/src/lighthouse/lighthouse_core.c#L180-L188

knmcguire:
Tried this in the client and it does not work... https://github.com/bitcraze/crazyflie-clients-python/pull/623 

Read only params and get_param is not meant for this functionality so I will need to turn it into a logging variable as well (since the param one is a core variable). Let's see how it goes.

Issue #1094, opened 2022-08-15
Support 2+ base stations for crossing beam method


krichardsson:
In the Lighthouse system, there is the option to use the crossing beam method to compute the current position. This was originally implemented for LH1 and currently it only works for 2 base stations, id 1 and 2. In a LH2 it is now possible to use more than 2 base stations, but base stations with an id > 2 will be ignored. 

It is probably not very hard to extend the implementation to a crossing beam computation between arbitrary base stations. 

For the user, it is not obvious why the CF does not get any position estimate when base station 1 & 2 are not visible.

PR #1095, opened 2022-08-17
LH crossing beam support for 2+ base stations

krichardsson:
The current crossing beam method in the lighthouse system was originally implemented for LH1 and only uses base stations 1 and 2. This PR makes it possible to use the crossing beam method with more than 2 base stations in a LH2 system.


krichardsson:
Closes #1094

knmcguire (approved):

PR #1100, opened 2022-08-25
Add support for fullStateSetpoint in PID controller

whoenig:
When using fullState CRTP package, the desired rotation is encoded as a
quaternion. The Mellinger controller correctly computes the desired yaw
in this case. In the PID controller, this is currently ignored. This
change adds the same logic.

Test case: Crazyswarm2 teleoperation.


krichardsson:
Nice, thanks!

PR #1101, opened 2022-08-25
High-Level flight: plan shortest rotation to reach the desired yaw

whoenig:
* Adds new helper functions (and tests) to math3d:
  normalize_radians and shortest_signed_angle_radians
* For takeoff, landing, and goTo: plan the yaw motion such that always
  the shortest rotation is used.


krichardsson:
Thanks!

PR #1117, opened 2022-09-16
Add throttling for LH2

krichardsson:
When 2+ LH V2 base stations are visible a lot of samples are sent to the kalman estimator. This PR adds functionality to discard samples to keep the rate around 50 samples/s


tobbeanton (commented):
Code looks good! 
What I'm interested in is how well the throttling works. 
1. I guess you have some graphs you could show?
2. Should we verify that the samples that are discarded are well distributed?
3. Are there any edge cases that could be problematic?

src/modules/src/lighthouse/lighthouse_throttle.c
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * When many LH V2 base stations are available in a system, the over all rate of samples sent to the estimator might be
+ * too high to handle. This parameter sets the (approximate) maximum rate (samples/s). 50 By default.
+ */
+PARAM_ADD(PARAM_UINT16, lh2maxRate, &maxRate)
+
+PARAM_GROUP_STOP(lighthouse)
+
+LOG_GROUP_START(lighthouse)
+LOG_ADD(LOG_FLOAT, disProb, &discardProbability)
```
tobbeanton, line 74: Should this be a debug log variable?

krichardsson: Yes. Fixed

PR #1118, opened 2022-09-20
Add debug log macros

krichardsson:
Sometimes it is useful to log many variables or rates when debuging a module, but it is bad to leave the logs when done since it will grow the size of the TOC. On the other hand it is nice to be able to leave them as they usually takes some time to set up and it can be hard to understand what to log. This PR adds macros that makes it easier to use logs that are enabled by defining a a macro. 

See the documentation in logparam.md for documentation.


knmcguire (approved):

tobbeanton (commented):
Looks good!

PR #1122, opened 2022-09-29
Migrating default PID gains and filter settings to platform defaults and Kconfig

matejkarasek:
Until now, PID values have been hardcoded, and could only be changed via permanent_parameters, or in a custom fork. 
This PR enables using platform-specific PID gain values and filter settings.

- PID gains are now defined for each platform individually in the [specific_patform]_defaults.h
- defaults for low pass filter settings are in platform_defaults.h, and can be overridden inside [specific_patform]_defaults.h for each platform if needed
- defines PID_FILTER_ALL and IMPROVED_BARO_Z_HOLD moved to Kconfig

### DIFF
--- a/src/modules/src/attitude_pid_controller.c
+++ b/src/modules/src/attitude_pid_controller.c
@@ -145,12 +145,7 @@ void attitudeControllerCorrectRatePID(
 
   pidSetDesired(&pidYawRate, yawRateDesired);
 
-  // there is probably a more elegant way to get the yaw rate setpoint...
-  static setpoint_t setpoint;
-  static state_t state;
-  commanderGetSetpoint(&setpoint, &state);
-  // adding a feedforward term
-  yawOutput = saturateSignedInt16(pidUpdate(&pidYawRate, yawRateActual, true) + yawFeedForw*setpoint.attitudeRate.yaw);
+  yawOutput = saturateSignedInt16(pidUpdate(&pidYawRate, yawRateActual, true));
 }
 
 void attitudeControllerCorrectAttitudePID(
@@ -334,10 +329,6 @@ PARAM_ADD(PARAM_FLOAT | PARAM_PERSISTENT, yaw_ki, &pidYaw.ki)
  * @brief Derivative gain for the PID yaw controller
  */
 PARAM_ADD(PARAM_FLOAT | PARAM_PERSISTENT, yaw_kd, &pidYaw.kd)
-/**
- * @brief Feedforward gain for the yaw controller
- */
-PARAM_ADD(PARAM_FLOAT | PARAM_PERSISTENT, yawFeedForw, &yawFeedForw)
 /**
  * @brief If nonzero, yaw setpoint can only be set within +/- yawMaxDelta from the current yaw
  */
--- a/src/modules/src/attitude_pid_controller.c
+++ b/src/modules/src/attitude_pid_controller.c
@@ -31,12 +31,18 @@
 #include "pid.h"
 #include "param.h"
 #include "log.h"
+#include "commander.h"
+#include "platform_defaults.h"
 
-#define ATTITUDE_LPF_CUTOFF_FREQ      15.0f
-#define ATTITUDE_LPF_ENABLE false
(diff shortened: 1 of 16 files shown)


matejkarasek:
Not sure what is going wrong with the build checks, all platforms build fine locally... Any ideas?

krichardsson:
After a very quick look (that is it might be completely wrong) I think the problem is that the unit tests can not include the platform_defaults.h file. You probably have to add it to [the unit test config](https://github.com/bitcraze/crazyflie-firmware/blob/master/tools/test/gcc.yml#L23-L71)

You can run the unit tests with `tb test`

I will have mote time to look at the PRs next week

knmcguire:
Same here:) Please check what is causing the CI to fail

matejkarasek:
> After a very quick look (that is it might be completely wrong) I think the problem is that the unit tests can not include the platform_defaults.h file. You probably have to add it to [the unit test config](https://github.com/bitcraze/crazyflie-firmware/blob/master/tools/test/gcc.yml#L23-L71)
> 
> You can run the unit tests with `tb test`
> 
> I will have mote time to look at the PRs next week

Thanks for the pointer @krichardsson, will check it out!

matejkarasek:
So running `tb test` in docker passes all the tests.
But `tb build` fails at same place as the CI.

The unit test config seems to already include the folder where platform_defaults.h is included, so it must be something else 🤔
https://github.com/bitcraze/crazyflie-firmware/blob/b370e20e50106dc10837c97e28404ff5d3efc5ea/tools/test/gcc.yml#L56

krichardsson:
The problem is not the unit tests, but building the python bindings. You can run them with `tb test_python`.

I think you need to add `"src/platform/interface",` to the include section of bindings/setup.py

krichardsson:
The call to `commanderGetSetpoint()` in attitude_pid_controller.c is also causing problems, in general it feels a bit "backwards". Maybe you can pass in the value of setpoint.attitudeRate.yaw as a parameter to the function instead?

matejkarasek:
The build errors are fixed.

> The call to `commanderGetSetpoint()` in attitude_pid_controller.c is also causing problems, in general it feels a bit "backwards". Maybe you can pass in the value of setpoint.attitudeRate.yaw as a parameter to the function instead?

I have removed the feedforward gain (where this is needed) from the current pull request as it can be treated separately.

How would you feel about updating the entire PID structure, such that it includes an extra feedforward gain, and everything is handled by the `pidUpdate(...)` function?
Next to yaw control, we us feed forward gains also in the velocity controller:
https://github.com/bitcraze/crazyflie-firmware/blob/b370e20e50106dc10837c97e28404ff5d3efc5ea/src/modules/src/position_controller_pid.c#L242-L243

matejkarasek:
The last 2 commits include the feedforward gain in the PID structure, how do you feel about this solution?

src/modules/src/attitude_pid_controller.c
```diff
@@ -31,12 +31,18 @@
...
 #include "log.h"
+#include "commander.h"
+#include "platform_defaults.h"
 
-#define ATTITUDE_LPF_CUTOFF_FREQ      15.0f
-#define ATTITUDE_LPF_ENABLE false
-#define ATTITUDE_RATE_LPF_CUTOFF_FREQ 30.0f
-#define ATTITUDE_RATE_LPF_ENABLE false
 
+bool attFiltEnable = ATTITUDE_LPF_ENABLE;
```
krichardsson, line 38: Should these perhaps be static?

krichardsson (changes requested):
I think it looks good!
The feed forward addition to the the PIDs is a nice feature

src/modules/src/attitude_pid_controller.c
matejkarasek: Yes indeed.... Fixed!

matejkarasek:
Ok, great to hear!

krichardsson (approved):
Nice!

krichardsson:
Merged. Thanks!

PR #1124, opened 2022-10-03
Improve build scripts

krichardsson:
There is a lot of repetition for platforms in our build scripts. Ideally there should be one list of platforms (possibly in module.json) that is used by all scripts that requires the information


krichardsson:
This PR was mainly used for the development of github actions scripts. Closing

PR #1127, opened 2022-10-05
Implement escaped percentage support in eprintf

ataffanel:
The current implementation of eprintf, used among other things to print in the Crazyflie console, does not support printing the percen (%) symbol.

Traditionally, with printf, the percent symbol can be printed by doubling it "%%". This currently makes the Crazyflie crash and the unit-test framework freeze.

This PR fixes the issue by implementing the handling of "%%".


krichardsson (approved):
Merge it!

PR #1131, opened 2022-10-07
Bigquad early access fix

victorhook:
Enabled external battery voltage measurements to be used as system votlage (useful for Big Quad deck).
Updated power state handling to support lowPower state regardless of platform and if Big Quad deck is connected.
Moved constants from source code to kbuild.


src/hal/src/pm_stm32f4.c
```diff
@@ -273,31 +285,38 @@ void pmSetChargeState(PMChargeStates chgState)
...
-  uint32_t batteryLowTime;
+  bool usbPluggedIn = pmSyslinkInfo.usbPluggedIn;
+  bool isCharging = pmSyslinkInfo.isCharging;
+  PMStates nextState;
 
-  batteryLowTime = xTaskGetTickCount() - batteryLowTimeStamp;
+  uint32_t batteryLowTime = xTaskGetTickCount() - batteryLowTimeStamp;
 
-  if (isPgood && !isCharging)
+  if (usingBigQuadDeck)
```
tobbeanton, line 294: I would like to remove this dependency and have the BiqQuaddeck driver call a function to enable this behavior instead. Something like pmIgnoreChargedState()

tobbeanton (changes requested):

victorhook:
I agree that it makes more sense to **not** have the dependency in the `pm`.
I made a new public `void pmIgnoreChargedState(bool ignore)` method in `pm` to allow other modules to ignore charging state if needed.

tobbeanton (approved):

tobbeanton:
The merge of PR #1132 might create some conflicts.

PR #1147, opened 2022-11-09
Better handling of motor power distribution

krichardsson:
This PR updates the power distribution.
 
Previously the motor power has been capped per motor if the desired thrust exceeds the maximum possible thrust. In this PR the capping has been updated to maintain the relative power of the motors and the trust of all motors are reduced equally much. The intention is to prioritize attitude over altitude to reduce the risk of flipping. 

The power distribution code has also been refactored to make it easier to understand and test.

### DIFF
--- a/src/modules/src/power_distribution_flapper.c
+++ b/src/modules/src/power_distribution_flapper.c
@@ -33,6 +33,7 @@
 #include "num.h"
 #include "autoconf.h"
 #include "config.h"
+#include "motors.h"
 
 #include "debug.h"
 #include "math.h"
@@ -104,7 +105,7 @@ int powerDistributionMotorType(uint32_t id)
 {
   int type = 1;
   if (id == idPitch || id == idYaw)
-  { 
+  {
     type = 0;
   }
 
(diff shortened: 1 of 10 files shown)


krichardsson:
@matejkarasek I have updated the power distribution a bit and thus modified the flapper power distribution file. It all looks good (in my opinion) but I have note tested on a real Flapper. The intention is that the functionality should be identical to before, would it be possible for you to take a look?
- Kristoffer

matejkarasek:
Thanks for the heads up, will try to test it and report back soon.

The new capping method, which prioritizes the attitude, could also be interesting for the Flapper :)

krichardsson:
Agree, I think it can be interesting for the Flapper as well. I did not want to change anything without checking with you first :-)

src/modules/src/power_distribution_flapper.c
```diff
@@ -142,40 +142,46 @@ bool powerDistributionTest(void)
...
+  #endif
+}
+
+void powerDistributionCap(const motors_thrust_uncapped_t* motorThrustBatCompUncapped, motors_thrust_pwm_t* motorPwm)
+{
+  #if CONFIG_POWER_DISTRIBUTION_FLAPPER_REVB
+    motorPwm->motors.m2 = limitThrust(motorThrustBatCompUncapped->motors.m2, 0, UINT16_MAX); // pitch servo
+    motorPwm->motors.m3 = limitThrust(motorThrustBatCompUncapped->motors.m3, 0, UINT16_MAX); // yaw servo
+    motorPwm->motors.m1 = limitThrust(motorThrustBatCompUncapped->motors.m1, idleThrust, UINT16_MAX); // left motor
+    motorPwm->motors.m4 = limitThrust(motorThrustBatCompUncapped->motors.m4, idleThrust, UINT16_MAX); // left motor
```
matejkarasek, line 179: copy&paste error in the comment, should be :
// right motor

matejkarasek:
So I tested this and all seems fine, thanks!

src/modules/src/power_distribution_flapper.c
krichardsson: Fixed. Thanks!

krichardsson:
Great!
Thanks @matejkarasek !

tobbeanton (approved):
Looks good and something we should have done a long time ago.

PR #1148, opened 2022-11-09
added actions to publish python bindings

knmcguire:
These are proposed actions to publish the cffirmware as a pypi package, shamelessly copied from @ataffanel's implementation on [crazyflie-link-cpp](https://github.com/bitcraze/crazyflie-link-cpp). For simulation we would like to make it easier to install the bindings so that we can implement it better into webots/crazyswarm2

This probably needs some discussion... so let's see :)

### DIFF
--- a/.github/workflows/build_bindings.yml
+++ b/.github/workflows/build_bindings.yml
@@ -1,0 +1,164 @@
+name: build bindings
+
+on:
+  push:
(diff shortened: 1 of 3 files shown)


knmcguire:
yeah... so that doesn't work 😄 I'll try locally first how to do this properly.

Issue #537, opened 2020-01-24
Option to use SI units in controller/power-distribution interface


jpreiss:
This issue refers to: https://github.com/bitcraze/crazyflie-firmware/compare/master...jpreiss:si-units. The code is not ready to merge into master yet, but it's enough to give an idea.

I propose to add a new interface between controller and power distribution using SI units. In the existing firmware, the units are related to the fixed-point motor power input unit. This is propagated into the controller gain parameters, which do not have any clear physical meaning.

In the new interface, the controller outputs a desired angular acceleration and a desired normalized thrust (thrust divided by mass). The power distribution section is now aware of the Crazyflie's mass, geometry, moments of inertia, and the torque:thrust ratio of the propellers. It uses this system-ID information to compute how much thrust each motor should produce in Newtons.

The result is that the PID controller gains have meaningful non-dimensionalized units; for example the P gain is (m/s^2)/m or just 1/s^2 - it tells how fast we should accelerate per unit of distance away from the setpoint.

The code is rough:
- I only applied the changes to the Mellinger controller in trajectory-tracking mode. There are `assert(false)`s in the other branches.
- I'm not sure if `platform.h` is the right place to put the system-ID info.
- I feel that the code for switching between SI and normal modes is clunky, but we could get rid of that if we convert all the controllers to SI.

If other people want this feature, I hope that other contributors who are more familiar with the other controllers could apply the same process. The main challenge is preserving the controller gain values so that the observable behavior of the CF doesn't change. I documented the process of how I did that [here](https://github.com/jpreiss/crazyflie-fix-units).

krichardsson:
Yes please! We have been discussing this for years, but have never got started.
I like the approach of adding a parallell API, even-though it might add some noise and duplication to the code. It will enable an orderly transition. 

> I'm not sure if platform.h is the right place to put the system-ID info.
First some background; The platform concept was added when we released the Roadrunner (a tag) and was intended to create an abstraction for multiple hardware platforms with different functionality, hw implementations, sensors and other properties. 
In our lingo we currently have two platforms: tag and CF (quad rotor). Within each platform there are multiple deviceTypes: for instance CF2.0, CF2.1 and the Bolt.
A FW that is built for one platform, should work on all deviceTypes of that platform, so the same CF FW works on all our quads.
The deviceType is stored permanently in the physical device and is used when booting the device to identify the device type, verify that the FW is compatible and pick the correct configuration.

[...]

PR #1154, opened 2022-11-17
Add SI-units to control structure

krichardsson:
A first small step to implement SI-units, see #537.

Basic implementation of force/thrust using code from Mike Hamer. Still old battery compensation.

### DIFF
--- a/src/modules/interface/stabilizer_types.h
+++ b/src/modules/interface/stabilizer_types.h
@@ -7,7 +7,7 @@
  *
  * Crazyflie control firmware
  *
- * Copyright (C) 2011-2012 Bitcraze AB
+ * Copyright (C) 2011-2022 Bitcraze AB
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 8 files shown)


ataffanel (approved):
Looks good to me, it is about time we add some SI units in the chain :-).

whoenig (changes requested):
Your change is of course nice and clean, is usual!

I am mostly concerned about the design decision of exposing SI units in the first place. Given that the tendency of the platform is now to support many different frames/motors, having SI units will be pretty painful to maintain (one would essentially need to rescale the gains for each possible platform). That's why I'd suggest to only expose normalized thrust/torque and force new controllers to use that mode. Then, as long as the thrust-to-weight ratio is similar, very little tuning needs to be done. Of course, one could add SI units and later the normalized ones, but there is a risk that controllers simply will stick around with the SI units and that those will be almost impossible to use on custom frames (since domain knowledge is needed for tuning).

See also https://github.com/bitcraze/crazyflie-firmware/issues/537#issuecomment-1064423960 (and discussion below).

krichardsson:
I see what you mean but I'm not sure it is a problem to support SI units in the API as well. To me the SI units API is something that can be used by anyone that finds it useful, and if you chose to use it, the controller will not be easy to use on a different platform. We are still in control of what we merge and can chose if we want to merge a contribution using the SI unit API, or not. 

As far as I can see in your implementation in the power distribution of the SI units/normalized forces, there is not a lot of extra code to support both APIs. When my fast copy/paste implementation is upgraded to something better (probably inspired by your code), I assume it will not be too much extra work to maintain both. 

What about adding a comment to the SI unit API describing the drawbacks of using it?

I'm of course arguing like this, partly because of self interest as I'm planning to add Mike's controller that uses SI-units. As I'm short on time, I want to minimize my work for now but I do hope to convert it to normalized forces later on :-)

whoenig:
I think that makes sense - just be careful when you merge more controllers - at some point it will be difficult to get rid of them:-)

jpreiss:
From browsing the current [cflib commander.py](https://github.com/bitcraze/crazyflie-lib-python/blob/master/cflib/crazyflie/commander.py) and the [CRTP commander docs](https://www.bitcraze.io/documentation/repository/crazyflie-firmware/master/functional-areas/crtp/crtp_commander/) it's not clear to me if there's any way to send a setpoint with physically meaningful thrust unit or not. Is that right?

If such a feature does not exist yet, would it be straightforward to add now that we have SI-unit power distribution?

ataffanel:
It seems that the setpoint structure in the Crazyflie is not updated for SI units yet, the thrust in there is still unit-less.

The way to implement a setpoint packet with SI unit would be to add new packets to the [generic commander](https://github.com/bitcraze/crazyflie-firmware/blob/e56b7fddf6e6ac9e10d9bacfffb4c4bfcd629219/src/modules/src/crtp_commander_generic.c#L39-L60). This should already be possible but the generic commander packet handler will currently have to do the conversion from SI to the set-point structure in the Crazyflie. Updating the setpoint structure first would be the cleanest but is not absolutely required to expose a setpoint packet with SI thrust.

PR #1156, opened 2022-11-21
Add Brescianini controller

krichardsson:
This PR adds a non linear attitude controller by Brescianini, originally written by Mike Hamer. The code has been adjusted to fit the current framework. Also, matrix, vector and quaternion operations use the math3d lib instead of the CMSIS lib to make it testable from python. 

Note: the controller uses the SI units in the power distribution system and it should be converted to use normalized forces instead to make it easier to adjust for different platforms.


whoenig (approved):
I think it's great to have this! I just don't like the naming: i) this seems to also be a position controller, not only attitude; ii) most other controllers are also non-linear (like Mellinger or INDI). Perhaps we can just name it by the name again: controller_brescianini?

krichardsson:
Sounds like a good name, I'll change it. I started with Mike's controller but that did not feel right :-)

ataffanel (approved):

PR #1162, opened 2022-12-03
Fix GAP8 flashing problems

krichardsson:
This PR fixes bugs in the GAP8 flashing of the AI-deck. 

In teh current implementation some file sizes will fail to flash due to a buffer problem.

This PR also intriduces a generic utility that handles copying between buffers of different sizes in a structured way.

Fixes bitcraze/aideck-gap8-examples#104

### DIFF
--- a/src/deck/drivers/src/aideck.c
+++ b/src/deck/drivers/src/aideck.c
@@ -54,6 +54,7 @@
 #include "queue.h"
 #include "stm32fxxx.h"
 #include "system.h"
+#include "buf2buf.h"
 
 #include "cpx_internal_router.h"
 #include "cpx_external_router.h"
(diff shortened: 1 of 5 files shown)


evoggy (approved):

ataffanel (approved):
Looks good to me.

whoenig:
We have frequently flashing issues over the radio as well, often hanging at 3%, not 99%. In some cases, this gets resolved by a more stable power supply (as in battery + USB cable). However, I have one AI deck that simply does not flash wirelessly (but fine over JTAG).

In the "old" firmware, it simply hangs. In this new firmware, the AI deck seems to hit one of these handy new asserts. However, there is no way for me to see which assert it hit, because cfloader doesn't print anything and if I connect with cfclient afterwards, the assert information is already gone. How should one identify the assert? Perhaps this requires a cfloader change/setting to print console messages? @krichardsson .

krichardsson:
@whoenig This is a different problem, as far as I can understand. I have created a new issue here https://github.com/bitcraze/aideck-gap8-examples/issues/105
Please add any information that you have. 
I also added some information on how to get the assert data out.

PR #1165, opened 2022-12-08
Unify controller and estimator function names

whoenig:
getControllerType -> controllerGetType
getStateEstimator -> stateEstimatorGetType

so that global/public functions start with a common prefix


krichardsson:
Much better :-) 
Thanks!

PR #1166, opened 2022-12-08
Refactor controller update to take const setpoint

whoenig:
The update function of controllers should not alter the setpoint itself.
This could lead to unexpected behavior and makes the Python interface for
software-in-the-loop simulation less clean.

The change itself is simple, apart from the PID velocity controller.
Here, there is a small behavior change: when logging, the user-specified setpoint will be shown, not
the intermedate value altered by the controller.
However, this seems to be the more logical behavior and is consistent
with the behavior of all other controllers that compute intermediate
setpoints (e.g., position controller of mellinger
 computes intermediate angles).

### DIFF
--- a/src/modules/interface/controller.h
+++ b/src/modules/interface/controller.h
@@ -39,7 +39,7 @@ typedef enum {
 
 void controllerInit(ControllerType controller);
 bool controllerTest(void);
-void controller(control_t *control, setpoint_t *setpoint,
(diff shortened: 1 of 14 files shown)


krichardsson:
Looks good to me, I love `const` :-)
I think the setpoint is modified because that is the way the code was written a long time ago. Nice to get it fixed!

krichardsson:
Thanks @whoenig !

PR #1168, opened 2022-12-08
Added function to start battery autoupdate to avoid syslink queue overflow.

tobbeanton:
Part of fixing https://github.com/bitcraze/crazyflie2-nrf-firmware/issues/78.

Before the nRF was sending battery status and RSSI messages to STM over syslink after a startup delay. Now a activation message must be sent to avoid a syslink queue overflow.


victorhook:
Looks good :)

PR #1169, opened 2022-12-08
Discard incoming syslink packets until system is ready

### LINKED ISSUE #1087: Assert does not seem to work from interrupts
If an ASSERT() statement in an interrupt fails and triggers a re-boot, it seems as the STM freezes instead of re-boots.
This has been observed in the syslink UART DMA interrupt (current version 220726), in the uartslkDmaRXIsr() function in uart_syslink.c.

I tried to reproduce it by triggering an assert at a know time, I used the protocol version request message (that is send during connection) to trigger the assert.

Insert the following code into uartslkDmaRXIsr() (or uartslkHandleDataFromISR()) and connect from the client to trigger the assert.

```C
    if (slp.type == SYSLINK_RADIO_RAW || slp.type == SYSLINK_RADIO_RAW_BROADCAST) {
      uint8_t header = slp.data[0];
      if(header == 0xFD) {
        ASSERT(0);
        // while(true);
      }
    }
```

Expected result: A re-boot and ending up in failed assert mode
Actual result: STM freeze

I have not tested if the same problem exists in other interrupts, or if it only is in the UART DMA interrupt.

This is issue is probably one of the current problems in the stability lab. I suspect that the syslink detects a check sum error and asserts, but the problem is hidden by the fact that the STM freezes.

krichardsson:
See description in #1087
This PR implements solution 2.

Fixes #1087

### DIFF
--- a/src/drivers/src/uart_syslink.c
+++ b/src/drivers/src/uart_syslink.c
@@ -62,6 +62,7 @@ static xSemaphoreHandle uartBusy;
 static StaticSemaphore_t uartBusyBuffer;
 static xQueueHandle syslinkPacketDelivery;
 STATIC_MEM_QUEUE_ALLOC(syslinkPacketDelivery, 8, sizeof(SyslinkPacket));
+static bool syslinkPacketDeliveryReadyToReceive = false;
 
 #ifdef CONFIG_SYSLINK_RX_DMA
 static uint8_t dmaRXBuffer[64];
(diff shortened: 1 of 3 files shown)


src/drivers/src/uart_syslink.c
```diff
@@ -255,6 +256,10 @@ void uartslkResumeRx(void)
   NVIC_EnableIRQ(UARTSLK_IRQ);
 }
 
+void uartslkEnableIncoming() {
```
tobbeanton, line 259: Should we keep the same style in the file?

krichardsson: Fixed

PR #1170, opened 2022-12-12
Support to configure ESCs with BLHeli Configurator

victorhook:
This PR adds the ability to configure ESCs through BLHeli Configurator, which has a nicer UI than BLHeliSuite. BLHeli Configurator also runs nicely on Linux, compared to BLHeliSuite which must be used with something like `wine` to work.

To make the BLHeli Configurator recognize the Crazyflie, a "handshake" must first be done upon serial connection. This handshake follows the MSP extension protocol. I have reveresed engineered the protocol using BLHeli Configurator, so I'm not exactly sure which version I have really implemented.

I have tested to configure and flash the ESCs with the BOLT platform and it works OK.

Future improvements:
1. Know what version of the MSP protocol I've implemented.
2. Ensure that the information sent during the handshake (build version, FC name etc) is correct.
3. Refactor some of the request-handlers in `msp.c` to use `mspMakeTxPacket()` that I have added.

### DIFF
--- a/src/modules/interface/msp.h
+++ b/src/modules/interface/msp.h
@@ -93,4 +93,8 @@ void mspInit(MspObject* pMspObject, const MspResponseCallback callback);
  */
 void mspProcessByte(MspObject* pMspObject, const uint8_t data);
 
+/*
+*/
+bool mspHasSet4WayIf();
+
 #endif /* MSP_H_ */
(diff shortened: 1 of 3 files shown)


victorhook:
I am not sure if line 187: in `vcp_esc_pasthrough.c` is necessary. It is is just a small delay between reads: `vTaskDelay(M2T(2));`, which I used during development but never removed. It doesn't really make a difference but I think it can be safely removed.

tobbeanton (approved):
Basic functionality is there and this is still under development. Code looks good so let's merge it to ease further development.

PR #1180, opened 2023-01-04
Mellinger: support to change gains in bindings

whoenig:
To use the controller in a simulator (and for testing), we need two
changes:
1.) have a data structure that holds the state, so we can instantiate
 multiple controllers.
2.) the ability to change gains without recompiling.

This change introduces a state/param struct to achieve both objectives,
similar to the logic in collision_avoidance.c.

### DIFF
--- a/src/modules/interface/controller_mellinger.h
+++ b/src/modules/interface/controller_mellinger.h
@@ -27,12 +27,82 @@
 #define __CONTROLLER_MELLINGER_H__
 
 #include "stabilizer_types.h"
+#include "math3d.h"
 
-void controllerMellingerInit(void);
-bool controllerMellingerTest(void);
-void controllerMellinger(control_t *control, const setpoint_t *setpoint,
+typedef struct {
+    float mass;
+    float massThrust;
+
+    // XY Position PID
+    float kp_xy;      // P
+    float kd_xy;      // D
+    float ki_xy;      // I
+    float i_range_xy;
+
+    // Z Position
+    float kp_z;       // P
+    float kd_z;       // D
+    float ki_z;       // I
+    float i_range_z;
+
+    // Attitude
+    float kR_xy;      // P
+    float kw_xy;      // D
+    float ki_m_xy;    // I
+    float i_range_m_xy;
+
+    // Yaw
+    float kR_z;       // P
+    float kw_z;       // D
+    float ki_m_z;     // I
+    float i_range_m_z;
+
+    // roll and pitch angular velocity
+    float kd_omega_rp; // D
+
+    // Helper variables
+    float i_error_x;
+    float i_error_y;
+    float i_error_z;
+
+    float prev_omega_roll;
+    float prev_omega_pitch;
+    float prev_setpoint_omega_roll;
+    float prev_setpoint_omega_pitch;
+
+    float i_error_m_x;
+    float i_error_m_y;
+    float i_error_m_z;
+
+    // Logging variables
+    struct vec z_axis_desired;
+
+    float cmd_thrust;
+    float cmd_roll;
+    float cmd_pitch;
+    float cmd_yaw;
+    float r_roll;
+    float r_pitch;
+    float r_yaw;
+    float accelz;
+} controllerMellinger_t;
+
+void controllerMellingerInit(controllerMellinger_t* self);
+bool controllerMellingerTest(controllerMellinger_t* self);
+void controllerMellinger(controllerMellinger_t* self, control_t *control, const setpoint_t *setpoint,
                                          const sensorData_t *sensors,
                                          const state_t *state,
                                          const uint32_t tick);
+
+#ifdef CRAZYFLIE_FW
+
+void controllerMellingerFirmwareInit(void);
+bool controllerMellingerFirmwareTest(void);
+void controllerMellingerFirmware(control_t *control, const setpoint_t *setpoint,
+                                         const sensorData_t *sensors,
+                                         const state_t *state,
+                                         const uint32_t tick);
+
+#endif // CRAZYFLIE_FW
 
 #endif //__CONTROLLER_MELLINGER_H__
--- a/src/modules/src/controller.c
+++ b/src/modules/src/controller.c
@@ -25,7 +25,7 @@ typedef struct {
 static ControllerFcns controllerFunctions[] = {
   {.init = 0, .test = 0, .update = 0, .name = "None"}, // Any
   {.init = controllerPidInit, .test = controllerPidTest, .update = controllerPid, .name = "PID"},
-  {.init = controllerMellingerInit, .test = controllerMellingerTest, .update = controllerMellinger, .name = "Mellinger"},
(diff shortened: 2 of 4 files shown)


src/modules/interface/controller_mellinger.h
```diff
@@ -27,12 +27,82 @@
...
+} controllerMellinger_t;
+
+void controllerMellingerInit(controllerMellinger_t* self);
+bool controllerMellingerTest(controllerMellinger_t* self);
+void controllerMellinger(controllerMellinger_t* self, control_t *control, const setpoint_t *setpoint,
+                                         const sensorData_t *sensors,
+                                         const state_t *state,
+                                         const uint32_t tick);
+
+#ifdef CRAZYFLIE_FW
```
knmcguire, line 97: I actually never noticed the CRAZYFLIE_FW define before. Is this to separate functionality for the python bindings build and the firmware build? Will there be an build error if you don't use it?

src/modules/src/controller.c
```diff
@@ -25,7 +25,7 @@ typedef struct {
 static ControllerFcns controllerFunctions[] = {
   {.init = 0, .test = 0, .update = 0, .name = "None"}, // Any
   {.init = controllerPidInit, .test = controllerPidTest, .update = controllerPid, .name = "PID"},
-  {.init = controllerMellingerInit, .test = controllerMellingerTest, .update = controllerMellinger, .name = "Mellinger"},
+  {.init = controllerMellingerFirmwareInit, .test = controllerMellingerFirmwareTest, .update = controllerMellingerFirmware, .name = "Mellinger"},
```
knmcguire, line 28: For the sake of consistency, is it perhaps better to switch the names of these functions? Call these controllerMellinger* like with the others and those that take the controller struct  controllerMellingerBase* or something?

src/modules/src/controller_mellinger.c
```diff
@@ -331,98 +342,100 @@ PARAM_GROUP_START(ctrlMel)
...
+LOG_ADD(LOG_FLOAT, accelz, &g_self.accelz)
+LOG_ADD(LOG_FLOAT, zdx, &g_self.z_axis_desired.x)
+LOG_ADD(LOG_FLOAT, zdy, &g_self.z_axis_desired.y)
+LOG_ADD(LOG_FLOAT, zdz, &g_self.z_axis_desired.z)
+LOG_ADD(LOG_FLOAT, i_err_x, &g_self.i_error_x)
+LOG_ADD(LOG_FLOAT, i_err_y, &g_self.i_error_y)
+LOG_ADD(LOG_FLOAT, i_err_z, &g_self.i_error_z)
 LOG_GROUP_STOP(ctrlMel)
+
+#endif // CRAZYFLIE_FW
```
knmcguire, line 441: It is not ideal that we put all this code and logging and variables in the CRAZYFLIE_FW define... Technically the struct exists but it is just not written in ? I had a question about the use of this define in one of my earlier questions.

knmcguire (commented):
I have some larger generic comments, just to enhance my understanding of what you are trying to achieve in the PR 😄

src/modules/interface/controller_mellinger.h
whoenig: I copied this logic from collision_avoidance.c - might have been introduced there by @jpreiss. We want it here to avoid that these functions end up in the bindings, since they are "internal". A cleaner way could be to have separate files, but that makes code navigation more annoying.

src/modules/src/controller.c
whoenig: The goal was to keep using the "controllerMellinger*" name for the "public" functions that also end up in the bindings. Ultimately, this sort of structural change will need to be done for all controllers that we want to have bindings for.

src/modules/src/controller_mellinger.c
whoenig: The if/endif preprocessor doesn't need to go around logging/params. In this file, it might not technically be needed at all, since the binding compilation process only generates bindings from a given header file.

knmcguire: Ah yeah that is true. Would it be okay for you to remove this define in case it is not necessary? I just want to prevent more ifdefs to be added if necessary as it complicates the code eventually.

src/modules/interface/controller_mellinger.h
jpreiss: I considered both choices. IIRC, I didn't have particularly strong motivation for choosing one file with `#ifdef` over two files. I probably just did it to reduce the number of files. I don't see any discussion in https://github.com/bitcraze/crazyflie-firmware/issues/567 either.

src/modules/src/controller_mellinger.c
jpreiss: > Technically the struct exists but it is just not written in ?

In `collision_avoidance.c` the global state struct is defined inside the `#ifdef CRAZYFLIE_FW`. The bindings shouldn't need it at all.

> The if/endif preprocessor doesn't need to go around logging/params.

If they are moved outside the `#ifdef CRAZYFLIE_FW`, then the C compiler for the bindings will need to know about the logging/param system, right? When I was working on `collision_avoidance.c` I assumed we wouldn't want that.

jpreiss:
Repeating my inline comment: There was no strong motivation for choosing `#ifdef CRAZYFLIE_FW` over using two files. If the Crazyflie firmware team prefers the latter, now would be a good time to change the collision avoidance module accordingly and apply the multi-file design for this & other future bindings.

src/modules/src/controller.c
knmcguire: sounds fine by me!

knmcguire:
Next to the question on what to do with define, it all looks good to me! 

But for obvious reasons, I do want to flight test this 😄 Tomorrow it's a holiday in Sweden so that will be next week.

whoenig:
I got rid of the define in the *.c file (and made sure it still works fine). For the one in the header I see four options:

1. Keep as is, i.e., use CRAZYFLIE_FW to not include the *Firmware functions in the bindings
2. Remove the guard. Then, the *Firmware functions will also be in the bindings (but are not useful IMO)
3. Have a separate header file for the *Firmware functions, e.g., controller_mellinger_firmware.h
4. Remove the *Firmware functions from the header. Use "extern" declarations in controller.c, rather than a include.

I think options 1 or 3 would be best, but perhaps best if the firmware maintainers raise their preferences.

Flight test: I did one before submitting the PR, but a second one is certainly even better!

krichardsson (approved):
Looking good!

knmcguire:
I flight tested btw, so I'll give the approve! @whoenig do you want to do something still with @krichardsson  comments or do you like us to merge it?

knmcguire (approved):

whoenig:
I like @krichardsson ideas, just not sure about the naming `utils` as this usually implies little helper functions. Since this version is flight tested, perhaps we should merge it and then have a separate change for moving files around, without touching any computation.

krichardsson:
@whoenig I agree, let's keep this PR as it is and refactor later.

PR #1183, opened 2023-01-06
Test framework improvements

krichardsson:
This PR aims at improving testability, both in unit tests and the python tests (using python bindings). A secondary goal is also to make it possible to add the kalman estimator to the python bindings which would make it possible to use it in simulations and test.

The main changes include:
* `ASSERT()` in the firmware is propagated to the test frameworks and fails the test if an assert is false 
* `DEBUG_PRINT()` in the firmware is now printed in the test console output
* CMSIS matrix math functions are supported from tests, that is the matrix operations in `cf_math.h` works when called from a test
* Fixes a bug related to the `@IGNORE_IF_NOT` annotation that prevented some tests from running

### DIFF
--- a/bindings/cffirmware.i
+++ b/bindings/cffirmware.i
@@ -89,6 +89,12 @@ void collisionAvoidanceUpdateSetpointWrap(
     free(workspace);
 }
 
+void assertFail(char *exp, char *file, int line) {
+    char buf[150];
+    sprintf(buf, "%s in File: \"%s\", line %d\n", exp, file, line);
+
+    PyErr_SetString(PyExc_AssertionError, buf);
+}
 %}
 
 %pythoncode %{
--- a/bindings/setup.py
+++ b/bindings/setup.py
@@ -54,6 +54,7 @@ cffirmware = Extension(
         # The following flags are also used for compiling the actual firmware
         "-fno-strict-aliasing",
         "-Wno-address-of-packed-member",
+        "-DUNIT_TEST_MODE",
     ],
 )
 
--- a/bindings/setup.py
+++ b/bindings/setup.py
@@ -13,9 +13,22 @@ include = [
     "src/config",
     "src/drivers/interface",
     "src/platform/interface",
+    "vendor/CMSIS/CMSIS/DSP/Include",
+    "vendor/CMSIS/CMSIS/Core/Include",
 ]
 
 fw_sources = [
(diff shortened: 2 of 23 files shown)


krichardsson:
Hmm, there seems to be some sort of problem when building the `examples/app_hello_world/src/hello_world.c` OOT app. I think the problem is that the object files in the firmware are not properly linked and the linker can not find any functions in the firmware. Not sure why (`make` is not my best friend...). @ntamas I think you were involved in the c++ support, do you have any idea what is going on?

It seems as this is not a new problem caused by this PR, but it is rather just showing the problem.
Another way of triggering it is to go back to master and make a call from `appMain()` in `examples/app_hello_world/src/hello_world.c` to any random function in the firmware, for instance

``` C
#include "buzzer.h"
void appMain()
{
  buzzerInit();
}
```

The result is `/home/kristoffer/code/bitcraze/crazyflie-firmware/examples/app_hello_world-cpp/src/hello_world.cpp:58: undefined reference to `buzzerInit()'`

ntamas:
I think this is caused by differences in C and C++ linkage. If you want a header file to behave nicely, irrespectively of whether you include it from a C or a C++ source file, you need to put this around all declarations in the header to force C linkage for the functions:

```
#ifdef __cplusplus
extern "C" {
#endif

[...declarations come here...]

#ifdef __cplusplus
}
#endif
```

This either has to be done for all your headers that you ever want to include from C++ files, _or_ you need to wrap the `#include` directives in the .cpp files like this:

```
extern "C" {

[...include all the headers from the Crazyflie firmware here...]

}
```

For instance, this patch solves the compilation of `app_hello_world-cpp` in this PR:

```patch
diff --git a/examples/app_hello_world-cpp/src/hello_world.cpp b/examples/app_hello_world-cpp/src/hello_world.cpp
index 53de5c7c..d3607164 100644
--- a/examples/app_hello_world-cpp/src/hello_world.cpp
+++ b/examples/app_hello_world-cpp/src/hello_world.cpp
@@ -37,12 +37,12 @@
 extern "C"
 {
   #include "app.h"
-}

-#include "FreeRTOS.h"
-#include "task.h"
+  #include "FreeRTOS.h"
+  #include "task.h"

-#include "debug.h"
+  #include "debug.h"
+}

 #define DEBUG_MODULE "HELLOWORLD"
```

krichardsson:
Thanks @ntamas !

Issue #1067, opened 2022-06-22
release the python bindings as pip package


knmcguire:
If we release the python bindings of the firmware for every firmware release, we could use the functions in external projects more easily, like in the controller of Webots.

knmcguire:
ah good! Thanks! 

you don't think there will be any OS problems? That if the bindings is build for Ubuntu that they won't work for window for instance?

whoenig:
You have to build for each platform you want to support a separate wheel, which is uploaded to pip. The CI does take care of that, though (in the build part of it: https://github.com/bitcraze/crazyflie-link-cpp/blob/master/.github/workflows/CI.yml).

knmcguire:
ah great. I'll have a look at the actions for that. And keep it in mind for the next release, which will probably be in September.

knmcguire:
We will not do this for release 2022.09 due to lack of time, but a potential candidate for a release after that

knmcguire:
So I've looked at this more closely and with some trouble I need to conclude unfortunately it seems to be not as simple as copying the same structure as https://github.com/bitcraze/crazyflie-link-cpp. For Linux, and probably mac as well, it should be relatively simple with the following steps:
```
make cf2_defconfig && make
swig -python -Isrc/modules/interface -Isrc/hal/interface -Isrc/utils/interface -o build/cffirmware_wrap.c bindings/cffirmware.i
python3 bindings/setup.py bdist_wheel
```

But windows of course, this makes it more difficult. Building the firmware natively on windows is a pain so I'm using the bitcraze builder docker for that:
```
docker run -it -v %CD%:/module bitcraze/builder make cf2_defconfig
docker run -it -v %CD%:/module bitcraze/builder make
swig -python -Isrc/modules/interface -Isrc/hal/interface -Isrc/utils/interface -o build/cffirmware_wrap.c bindings/cffirmware.i
python bindings/setup.py bdist_wheel
```


First of all, the `"-Wno-address-of-packed-member"` can not be used but that is okay to remove as that only creates warnings if I'm correct.  So after disabling that I get the following types of errors, which does not happen in Ubuntu.

```
build/cffirmware_wrap.c(18254): error C2059: syntax error: ')'
build/cffirmware_wrap.c(18267): error C2065: 'arg1': undeclared identifier
build/cffirmware_wrap.c(18267): error C2065: 'sweepAngleMeasurement_t': undeclared identifier
```
[...]

PR #1184, opened 2023-01-10
Fix python wheel build of the python binding

ataffanel:
So far, generating a python wheel was not functional because it was missing the `cffirmware.py` file/module.

This PR fixes it and add a `python_wheel` Make target to make it easier to build the wheel. Building the python wheel now only requires calling the `tools/build/python_wheel` script.


ataffanel:
This is a required step for issue #1067

krichardsson (changes requested):
Looking good!
I would suggest to move the `python_wheel` script out of the `tools/build` directory and put it in a new one, something like `tools/python_bindings` or similar, as it can not be run from the toolbelt.

krichardsson (approved):
Sweet!

PR #1185, opened 2023-01-11
Params: add missing callback for writeByName

whoenig:
The callback function is used, e.g., for led.bitmask and currently only
implemented for paramWriteProcess, not paramWriteByNameProcess


krichardsson:
Thanks!

PR #1188, opened 2023-01-19
Add handling of rx fail for Loco

### LINKED ISSUE #1186: Loco decks on Roadrunners stop in test lab
I have observed that the integrated loco decks on the Roadrunners in our test lab stop running sometimes. The behavior is similar to the problem we had 2 years ago (#436 and #368). 

My feeling is that it started a couple of months ago, but that it has become worse over time. In the beginning I could see one or two that were not blinking, now it can be 9 out of 10 after 1 day.

krichardsson:
This PR adds handling of failed receives in the Loco system.
When a packed is not properly received, the underlying library calls the `receiveFailedHandler`. This handler has not been properly implemented with the result that the radio is not reset for a new receive. This in turn stops the loco deck from any further processing.

This PR fixes #1186

### DIFF
--- a/src/deck/drivers/src/locodeck.c
+++ b/src/deck/drivers/src/locodeck.c
@@ -176,6 +176,10 @@ static void rxTimeoutCallback(dwDevice_t * dev) {
   timeout = algorithm->onEvent(dev, eventReceiveTimeout);
 }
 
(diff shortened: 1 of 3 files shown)


knmcguire (approved):
looks good

PR #1193, opened 2023-01-21
Added paramNotifyChanged() to let apps notify the core when they modify the value of a parameter programmatically

ntamas:
This function intends to cater for the use-case when an app modifies the value of a parameter using `paramSetInt()`, `paramSetFloat()` or some similar mechanism (i.e. not externally via a CRTP packet but internally). Currently these functions do not call the callbacks associated to the parameter if there is one; `paramNotifyChanged()` allows the app code to trigger the callback programmatically.

Another option would be to change `paramSetInt()`, `paramSetFloat()` and similar functions to call the callback, but I don't know what other implications this change may have so I went for the surgical option instead.

### DIFF
--- a/src/modules/interface/param_logic.h
+++ b/src/modules/interface/param_logic.h
@@ -150,12 +150,6 @@ void paramLogicInit();
  */
 void paramLogicStorageInit();
 
-/**
- * @brief Notify the param subsystem that we changed the value of a param
- * programmatically.
- */
-void paramNotifyChanged(paramVarId_t varid);
-
 // The following functions SHALL NOT be called outside paramTask!
 void paramWriteProcess(CRTPPacket *p);
 void paramReadProcess(CRTPPacket *p);
--- a/src/modules/interface/param_logic.h
+++ b/src/modules/interface/param_logic.h
@@ -150,6 +150,12 @@ void paramLogicInit();
  */
 void paramLogicStorageInit();
 
+/**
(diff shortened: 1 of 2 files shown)


krichardsson:
I agree with @ataffanel 
Sounds reasonable that the callback is called also when a parameter is set from an app in the FW

ntamas:
First iteration. Maybe the call to `paramNotifyChanged()` could actually also be pushed down to `paramSet()`, but then the callbacks would also be called in `persistentParamFromStorage()` - do we want that or not?

krichardsson:
Oh, that is a good question! And a bit scary because it touches on the infinitely complex topic of initialization. The initialization process/sequence is slightly undefined at the moment and in my opinion we really should shape it up to clearly define the expected behavior. For how I would say that we do not call the callback when setting parameters from persisted values.

At the moment there is no case where we have a persistent parameter with a callback so we can decide on what ever solution we want I guess.

ataffanel:
I agree, one idea of stored parameter was to 'simulate' the behavior of default value, in that sense not calling the callback for stored parameter sounds like the right behavior.

krichardsson (approved):

krichardsson:
Thanks!

ntamas:
Note that the title of the PR became misleading now that `paramNotifyChanged()` is not a public function any more. The upside is that apps don't need to do anything explicitly now.

PR #1194, opened 2023-01-22
Refactor kalman estimator

krichardsson:
This PR contains refactoring of the kalman estimator to make it easier to run it from python bindings.

The code in estimator_kalman.c must be re-implemented (with modifications) in python when running thrugh the bindings as this code has lots of dependencies to FreeRTOS. The chaninges in this PR aims at moving as much as possible to the kalman core and other portable core to minimize and simplify the code that must be re-implemented.

Main changes:
* Add sub sampler for gyro and accelerometer
* Change all times in kalman core and related files from ticks to ms
* Make sure kalman core knows whether it has been updated and should be finalized (used to be handled outside)
* Make the kalman core aware of when it was previously updated to enable it to compute delta time

### DIFF
--- a/src/modules/interface/axis3fSubSampler.h
+++ b/src/modules/interface/axis3fSubSampler.h
@@ -1,0 +1,62 @@
+/**
+ * ,---------,       ____  _ __
+ * |  ,-^-,  |      / __ )(_) /_______________ _____  ___
+ * | (  O  ) |     / __  / / __/ ___/ ___/ __ `/_  / / _ \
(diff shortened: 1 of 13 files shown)


knmcguire:
I'll take a look now and review it.

@whoenig this might be of interest to you too since is meant for the estimator pythonbindings

knmcguire:
Me and @krichardsson has gone through it as the github difference is messing some parts up. But it checks out and it builds, but it still needs to be flighttested by @krichardsson

knmcguire (approved):
With the requirement that it needs to be flighttested

krichardsson:
Flight tested OK

PR #1199, opened 2023-01-24
Removed unused files

krichardsson:
Removed a directory that does not seem to be used.


knmcguire (approved):
Seems indeed to be useless

PR #1204, opened 2023-01-25
Add log for assert

krichardsson:
This PR adds a log that indicates if the system has failed an assert. This is mainly useful in test environments.


krichardsson:
There is already a similar functionality in the `system.selftestPassed` parameters. Closing this PR.

PR #1206, opened 2023-01-26
fix typo in name of Kalman estimator

ntamas:
There was a typo in the name of the Kalman estimator that prevented compilation if the Kalman estimator was selected explicitly in the configuration.


krichardsson:
Thanks!

PR #1208, opened 2023-01-30
Add reformat of storage

### LINKED ISSUE #1207: Add functionality to reset persistent data
Persistent data is stored in eprom, currently lighthouse geometry/calibration data and parameters. 
If persistent data is corrupt there is no way of clearing it in a simple way and the suggestion is to add a functionality where the user easily can reset the persistent storage.

krichardsson:
This PR adds a parameter that can be used to trigger a re-format of the storage. This will erase all stored data (persistent parameters and lighthouse geometry/calib) and can be useful if data in the storage is corrupt. 

Fixes #1207

### DIFF
--- a/src/hal/interface/storage.h
+++ b/src/hal/interface/storage.h
@@ -100,7 +100,17 @@ bool storageForeach(const char* prefix, storageFunc_t func);
 
 /**
  * Print storage information on the debug console
- * 
+ *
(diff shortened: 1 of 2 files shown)


ataffanel (approved):
Looks good, can be a useful functinality!

PR #1210, opened 2023-01-31
Always generate version information

### LINKED ISSUE #1049: Version information not updated in local build
The version information is handled by generating a c file, build/src/utils/src/version_gen.c that is compiled into the firmware.
This works fine when building from a newly cloned repo but when building locally in an "old" file tree, version_gen.c is not updated and the version that is compiled into the firmware is stale.
`make clean` does not remove the file either.

I found traces in tools/kbuild/Makefile.kbuild (search for 'version_h') that looks like an old implementation of forcing the build of version information, but it is targeted at version.h and does not work.

krichardsson:
This PR forces re-generation of the version_gen.c file when the firmware is built. Currently it is not always built which leads to invalid version information in the console log when a Crazyflie is booted.

Fixes #1049

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,4 +45,3 @@ build/
 **/*.cmd
 **/*.a
 *.elf
-src/utils/src/version_gen.c
--- a/src/utils/src/Kbuild
+++ b/src/utils/src/Kbuild
@@ -15,8 +15,8 @@ obj-y += malloc.o
 obj-y += version.o
 version-objs := version_gen.o
 
-src/utils/src/version_gen.c: src/utils/src/version.vtpl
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -45,4 +45,3 @@ build/
 **/*.cmd
 **/*.a
 *.elf
-src/utils/src/version_gen.c
```
ataffanel, line 48: Why is the version file removed from .gitignore? It likely should not be committed.

ataffanel (commented):
Good thing to regenerate this file for all the build but shouldn't it still ignored by git?

.gitignore
krichardsson: The path to the generated file is `build/src/utils/src/version_gen.c`, that is not in the `/src` dir which means that it is already ignored by the `build/` entry in `.gitignore`

PR #1211, opened 2023-02-01
Add OOT controller support

### LINKED ISSUE #922: Out of tree controller
Since we have an out of tree estimator option, symmetry asks for an out of tree controller as well!

Also, it would be good to add the documentation at the controller configuration page so users know that this is possible: https://www.bitcraze.io/documentation/repository/crazyflie-firmware/master/functional-areas/sensor-to-control/#configuring-controllers-and-estimators

krichardsson:
This PR adds support for OOT controllers.

Closes #922

### DIFF
--- a/configs/all.config
+++ b/configs/all.config
@@ -8,8 +8,9 @@
 # a linker error.
 # CONFIG_APP_ENABLE is not set
 #
-# Same is true for OOT estimator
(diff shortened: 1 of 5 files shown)


ataffanel (approved):
Looks good to me!

PR #1213, opened 2023-02-01
Rename "any" controller and estimator

### LINKED ISSUE #1007: Change of 'any' to autoselect
In the firmware and defconfig we indicate the word Any for auto selecting controllers and estimators. Unfortunately this is not very clear and we should change that notion to something like autoselect instead.

krichardsson:
This PR renames "any" to "auto select" for controller and estimator settings. Hopefully it will be easier to understand what they mean.

Note: Enums are also updated which might break apps the use them.

Closes #1007


knmcguire (approved):
Looks good!

PR #1214, opened 2023-02-05
Enable setting motor power level through MSP

victorhook:
This PR adds functionality to set the motor power level through MSP, which can be done directly from ESC Configurator for instance (using the motor sliders). This can be very handy when configuring the Crazyflie Bolt.
I also refactored some code in `msp.c`.

A concern I discovered when working with the Crazyflie Bolt and ESC Configurator:
1. Sometimes the ESC flashing procedure through ESC Configurator fails before it's finished. The Crazyflie crashes with watchdog timeout, which results in a broken ESC which has to be un-bricked manually (for instance with C2 interface). I assume that there is something in `esc4wayProcess()` function that takes too long, but have not investigated this further.

### DIFF
--- a/src/drivers/interface/motors.h
+++ b/src/drivers/interface/motors.h
@@ -315,6 +315,19 @@ int motorsESCIsLo(uint32_t id);
  */
 void motorsBurstDshot();
 
+/*
+ * Enables or disables the ability to set power levels of the motors directly
+ * which is done by calling motorPowerSetValue(id, value)
+*/
+void motorSetPowerEnabled(bool enable);
+
+/*
+ * Sets the power level for the given motor id.
+ * Note that the powerSetEnable functionality must be set to true
+ * in order for this to have any effect.
+*/
+void motorSetPowerValue(uint32_t id, uint16_t value);
+
 /**
  * Set the PWM ratio of the motor 'id'
  */
--- a/src/drivers/interface/motors.h
+++ b/src/drivers/interface/motors.h
@@ -315,19 +315,6 @@ int motorsESCIsLo(uint32_t id);
  */
 void motorsBurstDshot();
 
-/*
- * Enables or disables the ability to set power levels of the motors directly
- * which is done by calling motorPowerSetValue(id, value)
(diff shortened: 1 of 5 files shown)


tobbeanton:
Nice addition, something we should probably have in the cfclient as well.
What I'm thinking about is how to best change parameters from within the FW. In this example it is not really a problem since the 4way task hogs the whole system but one might imagine there could be a discrepancy between what e.g. the cfclient would know and what the FW is set to. Thus when a parameter is changed this should be notified to connected clients. Let us discuss internally and get back to you.

victorhook:
That's a good point. Wasn't sure if and how you wanted the public API (to `motors.h`) to enable and set power level of the motors so I did something that perhaps can serve as base.

ataffanel:
The best would be to set parameters using the [internal param API](https://github.com/bitcraze/crazyflie-firmware/blob/b89ae61f14ed096006c86d8f90563d1b0e3dca28/src/modules/interface/param_logic.h#L128). This will send updated value to a potentially connected client so it make sure everything stays in sync.

victorhook:
Good idea Arnaud, I made a new commit where I removed the newly added public functions in `motors.h` and instead used the internal parameter API to enable and set the motor power values.
However, it's worth mentioning that using the internal param API instead requires the Crazyflie to be connected to the cfclient, or the flag `CONFIG_PARAM_SILENT_UPDATES` must be set, as described [here](https://github.com/bitcraze/crazyflie-firmware/blob/b89ae61f14ed096006c86d8f90563d1b0e3dca28/src/modules/interface/param_logic.h#L128).

I've tried this with cf bolt, using [ESC Configurator](https://esc-configurator.com/) and both options:
1) Ensure Crazyflie is connected to client.
2) Set `CONFIG_PARAM_SILENT_UPDATES` build flag.

works :)

krichardsson:
It i s a bit of a problem that it does not work unless a client is connected. I added an issue for this #1234.

krichardsson:
Thanks!

PR #1217, opened 2023-02-07
Add .DS_Store to .gitignore

valeriyvan:


krichardsson:
Thanks for the pull request!

Since .DS_Store is a platform specific file, it is perhaps better to add it to your global gitignore file?

It would be possible to accept this change but I'm slightly hesitant, we are generally trying to avoid platform specific files. If we add it in this repo, we should probably add it to all other repos as well.

valeriyvan:
> Thanks for the pull request!
> 
> Since .DS_Store is a platform specific file, it is perhaps better to add it to your global gitignore file?
> 
> It would be possible to accept this change but I'm slightly hesitant, we are generally trying to avoid platform specific files. If we add it in this repo, we should probably add it to all other repos as well.

makes sense

PR #1221, opened 2023-02-08
Housekeeping 

valeriyvan:


valeriyvan:
> Great stuff! It is better to group multiple spelling changes into one PR, like this.
> 
> The only issue with this PR is the renaming of `appchannelHasOverflowOccured()` to `appchannelHasOverflowOccurred()`. This function is part of a stable app API (that's why it is called from app_api/src, to remind us) and the idea is that we will not change these functions without first deprecating them for a while. The renaming is good, but we must also keep the old function with a deprecation comment, see [`appchannelReceivePacket()`](https://github.com/bitcraze/crazyflie-firmware/blob/b89ae61f14ed096006c86d8f90563d1b0e3dca28/src/modules/interface/app_channel.h#L92-L108) for a similar case.

done

krichardsson:
Thanks!

PR #1237, opened 2023-02-24
Improved TDoA outlier filter

### LINKED ISSUE #1158: TDoA3 outlier filter need update for Long Range
The TDoA3 outlier filter was originally written for small spaces and the "normal" mode, when running the Loco system in TDoA3 Long Range mode in a large system, it is not doing a great job resulting in outliers getting through. It is probably a good idea to look at the implementation and see if it can be improved, for instance also using the direction to the sample and not only the distance. The outlier filter should be re-written but it will be easier to do after the kalman estimator has been added to the python bindings (#602).

---

A temporary workaround that seems to solve some of the problem in a large system is to [change the settings](https://github.com/bitcraze/crazyflie-firmware/blob/e82ba19aedcd75a5a95a94618c00cfe6865d8e89/src/modules/src/outlierFilter.c#L34-L35) to 
``` C
#define MAX_BUCKET_FILL 7
#define FILTER_CLOSE_DELAY_COUNT 20
```

Adding a maximum error distance that can be set from the ground when the filter has converged is also an option.
Conceptually something like the following code could be [inserted here](https://github.com/bitcraze/crazyflie-firmware/blob/e82ba19aedcd75a5a95a94618c00cfe6865d8e89/src/modules/src/outlierFilter.c#L101)
``` C
  if (errorDistance > maxFilterDistance) {
    sampleIsGood = false;
  }
```

krichardsson:
This PR contains an updated outlier filter for TDoA samples in the kalman filter. It has generally been simplified and now better supports various packet rates and nose levels in the system. This should make it work with TDoA long range and in systems with many anchors. 

Closes #1158


src/modules/src/outlierFilter.c
```diff
@@ -102,6 +101,84 @@ bool outlierFilterValidateTdoaSteps(const tdoaMeasurement_t* tdoa, const float e
...
+// larger than the acceptance level will be discarded. The acceptance level is based on the integrator, with a
+// hysteresis to avoid rapid changes. When the filter is open, all samples are let through, usually at startup to let
+// the kalman filter converge. When the filter is closed, only samples with an error < the acceptance level, this should
+// be most of the time.
+// The acceptance level is based on the standard deviation used for the tdoa samples, which should be based on the
+// noise level in the system.
+
+
+// The maximum size of the integrator. This size determines the time [in ms] needed to open/close the filter.
+static const float INTEGRATOR_SIZE = 300.0f;
```
knmcguire, line 121: I'm wondering if we should make a struct for these values and initialize the avalues when the kalman filter is initialize? What do you think? It will make it more suitable for wrapping and the bindings?

knmcguire (commented):
Just a couple of comments on these!

src/modules/src/outlierFilter.c
knmcguire: Also it would make it possible to use this filter for something else

src/modules/src/kalman_core/outlierFilterLighthouse.c
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ * outlierFilter.c: Outlier rejection filter for the LPS system
```
knmcguire, line 24: for the lighthouse system

knmcguire (commented):
I'll also give it a quick flight test

knmcguire (approved):
flight test is okay! If the copy paste error I mentioned before is fixed go ahead and merge it

krichardsson:
Fixed the comment, and some other I found as well

Issue #588, opened 2020-06-04
Output useful error information when log and param names that are too long


krichardsson:
See #587 for a background.

Asserts that fail before the watchdog is started are hard to understand for a user since the Crazyflie does not reboot and output assert information to the console. In these cases the Crazyflie turns on the two red LEDs and hangs.

Check if it is possible to start the watchdog earlier in the start up sequence to reduce the number of possible asserts that can be triggered before the watchdog is started.

Specifically make sure log and param init outputs useful error information if names are too long.

jonasdn:
We know have a script `tools/verify/elf_sanity.py` to detect these cases, it is run in CI as well.

PR #1240, opened 2023-02-28
Added limitation of log variable name length

pacificlion:
Faced issue while creating custom log variables for crazyflie-firmware. 
There seems to be script already for run time checking (https://github.com/bitcraze/crazyflie-firmware/issues/588)
Adding this information in document for logging seems appropriate for users to create correct variables in start.


krichardsson:
Thanks!

PR #1245, opened 2023-03-04
[CPX] custom data types from gap8 to stm32

matthewoots:
This PR addresses https://github.com/bitcraze/aideck-gap8-examples/issues/116. A look at it is as shown below, using an example of `sending apriltag detection of corners` and `pose estimation` related data
cpx

### DIFF
--- a/src/modules/src/cpx/cpx.c
+++ b/src/modules/src/cpx/cpx.c
@@ -126,8 +126,14 @@ static void cpx(void* _param) {
           }
         }
         break;
+      case CPX_F_APP:
+      {
+        messageReceive(cpxRx);
+        break;
+      }
       default:
         DEBUG_PRINT("Not handling function [0x%02X] from [0x%02X]\n", cpxRx.route.function, cpxRx.route.source);
+        break;
     }
   }
 }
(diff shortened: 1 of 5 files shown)


src/modules/src/cpx/cpx.c
```diff
@@ -126,8 +126,14 @@ static void cpx(void* _param) {
           }
         }
         break;
+      case CPX_F_APP:
+      {
+        messageReceive(cpxRx);
```
ataffanel, line 131: There should likely be an API to register a callback for app messages or to receive packet from it. A similar things is done in CRTP with the [appchannel](https://github.com/bitcraze/crazyflie-firmware/blob/23af53b3aae324e8b8da95a29ab630dc30c2bc32/src/modules/interface/app_channel.h#L35). I feel like a similar API could be used there since it is a very similar concept.

ataffanel (changes requested):
It is great to have a documented easy-way for app to communicate between the Gap8 and the STM32!

However, most of this code is application specific, so it should live in a Crazyflie app in the `example\` sub-folder of the firmware. The only thing that looks like it should be in the firmware main code (in the `src\` folder) is the packet routing in `cpx\`, the rest should be moved to an example app.

matthewoots:
> @matthewoots could you address these changes? If we wait any longer then perhaps it no longer can be merged and we would like to prevent that.

Sorry I wasn't with the hardware to do any testing recently, Will try to get it up by this week, thank you!

Issue #1239, opened 2023-02-27
TOF measurement model comments unclear 


knmcguire:
The code comments of the TOF measurement model is a bit unclear. If we post an equation there we should at least say what the symbols mean or point towards the source equation in the paper

https://github.com/bitcraze/crazyflie-firmware/blob/0f2f0f07d2c2463cb07e4b187f9a9a58b689d124/src/modules/src/kalman_core/mm_tof.c#L46

The explanation is found in [this masterthesis](https://lup.lub.lu.se/student-papers/search/publication/8905295) in page 95 but we can add it also to our documentation: https://www.bitcraze.io/documentation/repository/crazyflie-firmware/master/functional-areas/sensor-to-control/state_estimators/

knmcguire:
This is already fixed a while ago

Issue #474, opened 2019-09-25
Implement tilt compensation for Z-ranger2


tobbeanton:
When the CF2 is tilted the height measurement is increased but the height above ground is roughly the same. Compensation should be added to handle this to improve flight characteristics.

ataffanel:
Tilt compensation is already implemented in the EKF (https://github.com/bitcraze/crazyflie-firmware/blob/31bbab58e9aea5f9378cd940e908f4be35400d7f/src/modules/src/kalman_core.c#L540). It seems that the measurement cone is hardcoded in the EKF though and this most likely needs to be parameterized.

Similar code should also be added to the complementary filter.

jonasdn:
I will have a stab at fixing this.

knmcguire:
So... I guess we would like to add this to the complementary filter then, since it has already been implemented into the EKF, and further clarified with PR #1251. But nobody else has really complained about this so I'll close it.

PR #1251, opened 2023-03-18
Clarify documentation for ToF laser ranging

nikhil-sethi:
This basic PR relates to #1239 and adds comments to make the ToF laser measurement model clear. The image for the online documentation has also been changed to reflect the concerned formula and diagram from the thesis.


hmllr:
Hi! Thanks for your pull request. However, it seems as if the documentation is now explaining the case of theta_px going towards zero, which is not the case with our sensor (it unfortunately is in some commented out code/comment which should be removed). Do you want to adapt your PR to explain the used measurement model (6.53 in the masterthesis but assuming no pitch/roll over theta_px / 2)? Or should I go ahead and clean up the misleading comments?

nikhil-sethi:
Hey @hmllr!
Thanks for checking the PR out. The commented-out code and the funky history of this file (https://github.com/bitcraze/crazyflie-firmware/commits/557dc37f7c8f215e10360f8a0b53372c94d7be3b/src/modules/src/estimator_kalman.c) tripped me up while writing the doc, but your explanation did make it somewhat clear. 
It seems like the `cos(angle)` was removed at some point because of a small aperture approximation and a NaN bug (https://github.com/bitcraze/crazyflie-firmware/commit/f3ba4c0cf13b882501fa59d92a56f828841d3cfd) but were added mistakenly without any effect here (https://github.com/bitcraze/crazyflie-firmware/commit/557dc37f7c8f215e10360f8a0b53372c94d7be3b). Would it be right to assume that the small angle approximation is still being used (as the `angle` variable is unused)? I can't test physically unfortunately but would love to help in correcting this. Feel free to make any mods to the PR as well!

hmllr:
I think it is not the small angle approximation that is used (ok, it is an assumption that the angle is small, but not in the classical way where you then leave away the cosine) - it is used that the VL sensor always returns the closest point it sees, and the FoV is 27degrees for the VL53L1X (or 25degrees for the older VL53L0X - I have no idea where the 15degree in the commented out code are coming from), so the drone hardly tilts enough to not return the closest distance to the ground.
This means there are two possibilities in how to clean this:
a) assume we always tilt less than 13.5degrees and removing all this angle stuff
b) the more correct way would be to add it back in - I think the only other thing you'd have to change would be to add in a divided by this->R[2][2] in the measurement equation, as otherwise the z seems to be missing there. I haven't tested it though and am not 100% sure.
Does this make things a bit clearer?
 
I am also linking issue #474 here, very old and related.

PR #1255, opened 2023-03-24
usec timer: add support to reset time

whoenig:
* initUsecTimer -> usecTimerInit for consistency
* new (public) function to reset the timer to 0
* new parameter "usec.reset" to trigger a reset

The latter an easy way to synchronize the clocks of multiple UAVs, by
sending a broadcast parameter change.


krichardsson:
Thanks!

PR #1256, opened 2023-03-24
usddeck: fix incorrect use of MAX_USD_LOG_EVENTS

whoenig:
The previous code was incorrect and did not allow to record data if MAX_USD_LOG_EVENTS=1


krichardsson:
Thanks!

PR #1257, opened 2023-03-24
usddeck: improve error handling on write

whoenig:
Instead of hitting an assert (and crash), we report an error and stop
writing to the card. This can happen if the card is loose, even during flight.


krichardsson:
Thanks!

PR #1267, opened 2023-04-13
Fix formatting in estimator_complementary.c

matejkarasek:
Fixing small formatting issue in estimator_complementary.c


krichardsson:
Thanks!

PR #1272, opened 2023-04-19
Extended supervisor

krichardsson:
This PR improves the supervisor functionality to handle bigger, brushless platforms in a better way. It is still work in progress but the goal is to move handling of conditions such as arming, loss of connection, emegency stop and similar to the supervisor.


knmcguire:
So these are some important references to emergency stop in the **high level commander**

This is the trajectory command, which we should probably still keep in here for Crazyswarm or others, but there should be some deprecation sign here? Or we change the functionality that only the planner is stopped but it's not changing the setpoint but instead sends a stop request.
 https://github.com/bitcraze/crazyflie-firmware/blob/39cf2a2bbb80859d3cea23aacd419b21cd5d2a33/src/modules/src/crtp_commander_high_level.c#L134
When the planner is stopped, this logic will actually return a nullsetpoint (with the pose states to ModeDisable) such that all motors will stop when the stabilizer This should either be removed or changed as an emergency stop request:
https://github.com/bitcraze/crazyflie-firmware/blob/39cf2a2bbb80859d3cea23aacd419b21cd5d2a33/src/modules/src/stabilizer.c#L287
https://github.com/bitcraze/crazyflie-firmware/blob/39cf2a2bbb80859d3cea23aacd419b21cd5d2a33/src/modules/src/crtp_commander_high_level.c#L319-L323
This function belong so the COMMAND_STOP, my comments are the same as above
https://github.com/bitcraze/crazyflie-firmware/blob/39cf2a2bbb80859d3cea23aacd419b21cd5d2a33/src/modules/src/crtp_commander_high_level.c#L548-L557

I think the issues is the following, as currently there are 2 trajectory or planner states on which the crtp_commander_high_level is responding on differently
- TRAJECTORY_STATE_IDLE ([plan_stop](https://github.com/bitcraze/crazyflie-firmware/blob/39cf2a2bbb80859d3cea23aacd419b21cd5d2a33/src/modules/interface/planner.h#L86) and [plan_is_stopped](https://github.com/bitcraze/crazyflie-firmware/blob/39cf2a2bbb80859d3cea23aacd419b21cd5d2a33/src/modules/interface/planner.h#L91) is reacting on those)
- TRAJECTORY_STATE_DISABLED ( [plan_disable()](https://github.com/bitcraze/crazyflie-firmware/blob/39cf2a2bbb80859d3cea23aacd419b21cd5d2a33/src/modules/interface/planner.h#L96) and [plane_is_disabled() ](https://github.com/bitcraze/crazyflie-firmware/blob/39cf2a2bbb80859d3cea23aacd419b21cd5d2a33/src/modules/interface/planner.h#L100) )

So [crtpCommanderHighLevelGetSetpoint ](https://github.com/bitcraze/crazyflie-firmware/blob/39cf2a2bbb80859d3cea23aacd419b21cd5d2a33/src/modules/src/crtp_commander_high_level.c#L301)sets the setpoint in case of TRAJECTORY_STATE_IDLE to a nullsetpoint, while it does not do that with TRAJECTORY_STATE_DISABLED. The latter was introduced with the push-based commander with mixing of HL and normale setpoints (see [this PR](https://github.com/bitcraze/crazyflie-firmware/pull/903)) as you don't want the drone to fall out of the sky if it gets a lower level setpoint, but you do want to simple disable the planner (done with [crtpCommanderHighLevelDisabled()](https://github.com/bitcraze/crazyflie-firmware/blob/39cf2a2bbb80859d3cea23aacd419b21cd5d2a33/src/modules/src/crtp_commander_high_level.c#L295)

Anyway, we might need to take a good look at the logic of this to make sure it all goes okay.

knmcguire:
This is the emergency stop handled by **generic commander**:

There is a packet type of the [crtp_commander_generic], on which the reaction is to do nothing, assuming the setpoint is already on zero upon initialization (which might not be the case with the HL commander coming in front) (https://github.com/bitcraze/crazyflie-firmware/blob/master/src/modules/src/crtp_commander_generic.c) that also handles stopping of motors

https://github.com/bitcraze/crazyflie-firmware/blob/39cf2a2bbb80859d3cea23aacd419b21cd5d2a33/src/modules/src/crtp_commander_generic.c#L66

https://github.com/bitcraze/crazyflie-firmware/blob/39cf2a2bbb80859d3cea23aacd419b21cd5d2a33/src/modules/src/crtp_commander_generic.c#L81-L87

Here we should probably not assume that the setpoint is zero, and put out a request to stop for the supervisor instead.

knmcguire:
Now for the **localization services**

I see that this is already in the first commits, but I'll add it anyway just for the sake of completion :)

Couple of localization services message types that are related:
https://github.com/bitcraze/crazyflie-firmware/blob/39cf2a2bbb80859d3cea23aacd419b21cd5d2a33/src/modules/src/crtp_localization_service.c#L296-L301

And these are related to these functions in stabilizer loop:
https://github.com/bitcraze/crazyflie-firmware/blob/39cf2a2bbb80859d3cea23aacd419b21cd5d2a33/src/modules/src/stabilizer.c#L340-L355

I guess... the question is, what was the function of the emergencystoptimeout that I don't really understand yet

knmcguire:
And the emergency stop from the **stabilizer** itself which can be enabled from a parameter.

https://github.com/bitcraze/crazyflie-firmware/blob/39cf2a2bbb80859d3cea23aacd419b21cd5d2a33/src/modules/src/stabilizer.c#L371-L374

And the functions are the same that localization srvs are using.

I see that the first commits of this PR have already taken this into account

src/modules/src/stabilizer.c
```diff
@@ -368,10 +350,6 @@ PARAM_ADD_CORE(PARAM_UINT8, estimator, &estimatorType)
  * @brief Controller type Auto select(0), PID(1), Mellinger(2), INDI(3), Brescianini(4) (Default: 0)
  */
 PARAM_ADD_CORE(PARAM_UINT8, controller, &controllerType)
-/**
- * @brief If set to nonzero will turn off power
- */
-PARAM_ADD_CORE(PARAM_UINT8, stop, &emergencyStop)
```
knmcguire, line 374: We should probably not remove this since this is a core parameter. Let's follow the deprecation manual https://test-www.bitcraze.io/development/development-overview/#the-update-process-and-deprecations

src/modules/src/supervisor_state_machine.c
```diff
@@ -0,0 +1,226 @@
...
+  "Warning, level out",
+  "Exception, free fall",
+  "Locked",
+};
+static_assert(sizeof(stateNames) / sizeof(stateNames[0]) == supervisorState_NrOfStates);
+#endif
+
+
+// State transition definitions
+static SupervisorStateTransition_t transitionsPreFlChecksNotPassed[] = {
```
knmcguire, line 53: `PreFl ` Could it just be PreFlight ? The name is pretty long anyway

knmcguire (commented):
very cool! I like the approach. I'll do a more thorough review once you are done.

krichardsson:
I think the high level commander looks OK and that both the stop and idle states are valid. To me it does not look like they are related to emergency stop, but rather just a normal stop. 
Stop = stop the motors. I think this is a valid case, I can imagine my self wanting to stop motors without doing a land at some point
Disable = Not using the high level commander at all. Makes sense :-)

What I think is missing though, is a connection between the high level commander and the isFlying in the supervisor. Currently the isFlying is only checking if the motors are spinning. The problem is that the supervisor changes state (from flying to landed) based on isFlying and we might end up being landed in mid-air, if we temporarily stop the motors.

krichardsson:
For the generic commander I think the same reasoning holds, the stop set point is probably not an emergency stop function, it simply means "stop the motors". @ataffanel Can you shed any light on the semantics?

I don't think we need to do anything here.

krichardsson:
The functionality in localization services is the real emergency stop. This functionality has been implemented in the new supervisor and I don't think there is anything more here that must be added at this point.

The idea is that the emergency stop message stops the motors - pretty straight forward.

The timeout is related to making sure the emergency stop communication is working. An "emergency stop device" is expected to continuously send "heartbeats" to show that it is alive, if no heartbeat is received, the Crazyflie will assume the device is broken and do an emergency stop.

krichardsson:
The stop parameter in **stabilizer** has now been re-implemented in the supervisor and should behave the same way as before.

ataffanel:
> For the **generic commander** I think the same reasoning holds, the stop set point is probably not an emergency stop function, it simply means "stop the motors". @ataffanel Can you shed any light on the semantics?

This is correct. The reason we have both a stop message and an emergency stop is that the stop message is intended for normal operation to stop the motors (for example when implementing a landing sequence controlled remotely). The e-stop is designed to be independent from the setpoint so that it can be activated independently from the normal control algorithms.

tobbeanton:
canFly state becomes wrong if system is disarmed in ready to fly state.

krichardsson:
@tobbeanton It looks like it should go back to `supervisorStatePreFlChecksNotPassed` https://github.com/bitcraze/crazyflie-firmware/blob/1fd125915efaf3917fdb31d3e8b56e7db935c6a3/src/modules/src/supervisor_state_machine.c#L75-L79

This should set canFly to false https://github.com/bitcraze/crazyflie-firmware/blob/1fd125915efaf3917fdb31d3e8b56e7db935c6a3/src/modules/src/supervisor.c#L215

I assume something is broken here?

tobbeanton:
I added a canArm log variable but I think we still need to discuss how we want things to work. It wasn't as easy as anticipated :smile:

src/modules/src/stabilizer.c
knmcguire: moved to another file

PR #1273, opened 2023-04-21
Add CPX_F_APP in cpxInternalRouter

RavenLite:
CPX packet on CPX_F_APP function not handled in `cpxInternalRouterRouteIn()`.
Currently, the user can't receive packets using `cpxInternalRouterReceiveOthers()` to develop their own app.

### DIFF
--- a/src/modules/src/cpx/cpx_internal_router.c
+++ b/src/modules/src/cpx/cpx_internal_router.c
@@ -76,6 +76,7 @@ void cpxInternalRouterRouteIn(const CPXRoutablePacket_t* packet) {
       case CPX_F_SYSTEM:
       case CPX_F_CONSOLE:
       case CPX_F_WIFI_CTRL:
+      case CPX_F_APP:
       case CPX_F_BOOTLOADER:
       case CPX_F_TEST:
         xQueueSend(mixedQueue, packet, portMAX_DELAY);


krichardsson:
Thanks for the pull request!

It indeed looks like `CPX_F_APP` has not been implemented, it actually does not seems to be used anywhere.

Even though this change would post the message to the `mixedQueue` (which is a good start), I still don't think it will be possible for an app to get hold of it. What is missing is the possibility for an app to register a callback [in the main CPX handler](https://github.com/bitcraze/crazyflie-firmware/blob/9956e68e10ba52855634c08abde62fb093707814/src/modules/src/cpx/cpx.c#L96-L151), for instance.

mfxjx000205:
If STM32 receive a CPX packet which has CPX_f_APP function from GAP8, it seems no queue for this message to save in. So, it is important for us to add this message into a queue.

mfxjx000205:
> 

The CPX_F_APP in firmware need us to add a new queue(for example:AppMSGQueue)? or we can't receive the packet.

krichardsson:
You are right, this is also missing. 

I think it makes sense to put the packet in `mixedQueue`.

You need to add a case in [`cpxInternalRouterRouteIn()`](https://github.com/bitcraze/crazyflie-firmware/blob/9956e68e10ba52855634c08abde62fb093707814/src/modules/src/cpx/cpx_internal_router.c#L75-L80) for `a` to add it to the queue (any message with destination `CPX_T_STM32` will end up in `cpxInternalRouterRouteIn()`)

After that, (as mentioned earlier) [in the main CPX handler](https://github.com/bitcraze/crazyflie-firmware/blob/9956e68e10ba52855634c08abde62fb093707814/src/modules/src/cpx/cpx.c#L96-L151) add a case for `CPX_T_STM32`. You need to add a function pointer and a function where an app can register a handler function for the pointer. In the case, call the function pointer if it is not 0, otherwise discard the CPX packet. 

You can find an example of function pointers and registration functions in [mem.h](https://github.com/bitcraze/crazyflie-firmware/blob/9956e68e10ba52855634c08abde62fb093707814/src/modules/interface/mem.h) for instance, even though this is slightly more complex with a mem def.

knmcguire:
@RavenLite the release is on the near horizon. Are you able to implement this you think or will you re-evaluate your approach?

RavenLite:
> @RavenLite the release is on the near horizon. Are you able to implement this you think or will you re-evaluate your approach?

@knmcguire Sorry for the late response. We have implemented most of our thought, @mfxjx000205 will continue working on this PR as soon as possible.

PR #1275, opened 2023-04-24
Add dependency checker tool

krichardsson:
This PR adds a tool for visualizing and counting dependencies between files in the source tree.


knmcguire (approved):
looks good and indeed quite self contained.

One question though, is this something we would like to host on our website though?

PR #1303, opened 2023-07-18
Make sure we can access the hash of the ledringtimingsmem

dolfje:
It was always the intention that you can check if crazyflie still has the latest ledring timings. That is a hash that you can upload and check afterwards. But current code only allows updating the timings itself.

### DIFF
--- a/src/deck/drivers/src/ledring12.c
+++ b/src/deck/drivers/src/ledring12.c
@@ -1119,8 +1119,8 @@ static bool handleLedringmemWrite(const uint32_t memAddr, const uint8_t writeLen
 static bool handleTimingmemRead(const uint32_t memAddr, const uint8_t readLen, uint8_t* buffer) {
   bool result = false;
 
-  if (memAddr + readLen <= sizeof(ledringtimingsmem.timings)) {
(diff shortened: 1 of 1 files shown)


ataffanel:
Thanks for the PR! I agree on principle: the hash was made to be read to verify if the led timing memory needs to be re-uploaded.

Though, while changing the read part is no problem, it is not implemented in the lib, the write part would be a breaking change and is a little bit more problematic. Mostly that the spirit was to get the hash calculated by the Crazyflie and not by the client writing the data.

I have a couple of alternate solution that would work:
 1. Make reading and writing asymetrical: we write timings and read hash-then-timing. This does not feel right though since it will not emulate a memory anymore
 2. Make the hash accessible on read at address  0xFFFC. This would work fine for me.
 3. Make the hash accessible as a log variable instead. A bit odd but it would work.

In all those 3 cases, the hash is calculated in the handleWrite function.

What do you think?

knmcguire:
@dolfje any comments on this? We don't like to keep pull request hanging so if you are not able to address these we will close the pull request if that is okay?

PR #1306, opened 2023-08-03
Added setup.py for installing Python bindings

simondlevy:
This feature allows you to install the Python cffirmware lib globally.

### DIFF
--- a/docs/building-and-flashing/build.md
+++ b/docs/building-and-flashing/build.md
@@ -138,7 +138,7 @@ First make sure that you have [SWIG](https://swig.org/) installed on your system
 make cf2_defconfig
 make bindings_python
 cd build
-sudo python3 setup.py install
+python3 setup.py install --user
 ```
 
 ## Make targets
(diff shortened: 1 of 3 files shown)


knmcguire:
Looks all good to me! @whoenig any objections to this?

docs/building-and-flashing/build.md
```diff
@@ -137,9 +137,10 @@ First make sure that you have [SWIG](https://swig.org/) installed on your system
 ```bash
 make cf2_defconfig
 make bindings_python
+cd build
+sudo python3 setup.py install
```
whoenig, line 141: I would prefer `python3 setup.py install --user`, if that works?

whoenig (approved):
I like it generally, but it would be better if we could avoid the sudo somehow. I suggested a possibility inline.

simondlevy:
I agree with @whoenig about the sudo.  I am about to do another pull request that uses --user instead

simondlevy:
I may have messed this whole process up, but it should be straightforward to get the two changes in anyway.

knmcguire:
Alright! We will keep an eye out on the new PR then :)

PR #1313, opened 2023-09-11
Fix LH as ground truth for LH1

krichardsson:
When the lighthouse system is used as the ground truth, data is not pushed to the kalman estimator, instead the crossing beam method is used. 

This PR fixes a bug that pushes one of the sweeps to the estimator anyway.
It also fixes increase of stats counters that should not be increased.


ataffanel:
Looks good, merging!

PR #1325, opened 2023-10-16
Refactor python tests

krichardsson:
Extracted kalman estimator emulation and utilities into the bindings directory. The idea is that will be useful in the future for simulations and similar.

### DIFF
--- a/bindings/util/estimator_kalman_emulator.py
+++ b/bindings/util/estimator_kalman_emulator.py
@@ -1,0 +1,137 @@
+import math
+import cffirmware
+
+
+
+
+
+class EstimatorKalmanEmulator:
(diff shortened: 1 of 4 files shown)


knmcguire (approved):
it looks all good to me, perhaps @whoenig would like to take a look too?  i just had one comment for better understanding.

So, this only supports TDOA with loco right now?

krichardsson:
> So, this only supports TDOA with loco right now?

Yes, decoding of other sensors to be added (when needed). The appropriate measurement model files for the added sensors must also be added to the build

knmcguire:
for the upcoming release we should have things merged so I'll just go ahead with that.

PR #1330, opened 2023-11-06
TDoA3 hybrid mode

### LINKED ISSUE #576: Inter ranging between multiple crazyflies by using the Loco decks

krichardsson:
This is an implementation of TDoA3 hybrid mode, that is a mix of TDoA and TWR.

The functionality is experimental and is a bit basic, but should at least serve as a platform to extend functionality from. 

Closes #576


knmcguire (approved):
It seems all good on the first eye, but it is a lot of code. Since all is behind build flags and we are in the middle of wrapping up the release, so it would be nice to include this in the testing. I'll go ahead and merge.

PR #1334, opened 2023-12-19
allow to build apps with other configs

atoums:
hello, that would be nice to change this so we can chose the desired config when building out of tree.

Thank you, have a nice day !


ataffanel:
This makes sense, thanks for taking the time to propose a PR for it! I am merging it now.

PR #1335, opened 2023-12-20
Addr pca interface

atoums:
changed pca95x4 inferface to take as parameter the address written on instead of having it staticly through macro.

We currently are using a version of the multiranger we manufactured, with 9 lasers, so we couldn't use the regular pca95x4 interface that only allow reading of 8 data at a time. We changed the signature of a few of theses functions so that we have 2 pca95x4 configurations to read data from, in our mrTask.


krichardsson:
Merging. Thanks for you contribution!

PR #1338, opened 2024-01-04
moving some physical constants so they can be config dependants

atoums:
Hello,
The mass of the crazyflie and the arm length has been moved so they can easily be integrated to a configuration.
Have a nice day !

### DIFF
--- a/src/modules/src/controller/controller_brescianini.c
+++ b/src/modules/src/controller/controller_brescianini.c
@@ -38,6 +38,7 @@
 #include "num.h"
 #include "math3d.h"
 #include "physicalConstants.h"
+#include "platform_defaults.h"
 
 static struct mat33 CRAZYFLIE_INERTIA =
     {{{16.6e-6f, 0.83e-6f, 0.72e-6f},
--- a/src/modules/src/controller/controller_mellinger.c
+++ b/src/modules/src/controller/controller_mellinger.c
@@ -42,6 +42,7 @@ We added the following:
 #include "position_controller.h"
 #include "controller_mellinger.h"
 #include "physicalConstants.h"
+#include "platform_defaults.h"
 
 // Global state variable used in the
 // firmware as the only instance and in bindings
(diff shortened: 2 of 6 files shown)


knmcguire (changes requested):
Oh yes nice catch! That would be a better place for it. 

To make it even better I think that these values should actually also be  placed in platform_defaults_cf2.h, since these are platform-specific.

knmcguire:
Also, an added thought, after we merge this an issue should be made that we also add these values for the flapper/bolt, although the latter might be difficult as anyone can make 'their own' platform. 

Also the brushless version of the crazyflie that is coming out soon, is still part of the cf2 platform with some adjustments, but that will be a bit tricky too. 

All of what I mention here in this post doesn't need to be changed in this PR, but just me thinking out loud :). Just the change indicated in the previous comment should be added.

atoums:
Those changes are idndefs to setup a default value that is needed for a platform to compile; as the mass is required for all platforms in the brescianini and mellinger controllers and the length is for the tag, cf2 and bolt configs. they are overridden by those #defined in a platform config.
Another security of provided it in the def config with ifndefs is not to break the homemade configs of those running the code that haven't redefined the mass in their config.
So provinding a default value in the def config seems important.

I can redefine those a second time with the same values in the cf2 platform config file to make it explicit for this platform, but i do not see how i can be mush more specific.
Sorry if i missed the target of your comments.

knmcguire:
Yes that is true, but I think it is still good to indicate where these default values come from. The mass and arms are significant to the CF2.1 specifically, so I do think it should go in the platform file. The ifndefs are just to make sure that it still compiles for the other platforms, which are the bolt and flappers, so I don't see it as a duplication since the intention is different. Technically these ifdefs should not be necessary and just as a failsafe and in the future, if a platform file does not have these values defined, that should generate a build error (but not for now)

After this PR, I will make a ticket that these values need to be specifically defined in the platform's file for the Bolt or flapper. 

Hope that makes sense! It will make the code more understandable.

atoums:
Thank you for explaining again. I hope this will do.

knmcguire:
Thanks! I'll probably do a quick build and flight test soon before merging but it looks good so far to me :)

knmcguire:
flight tested and it looks all good. I'll merge it. Thanks!

PR #1340, opened 2024-01-10
Servo support

matejkarasek:
This PR adds support for a single extra servo (on top of the existing 4 motors/actuators used in the power distribution). 
Its position is controlled via parameter "servo.servoAngle".
It is implemented as a deck, which can be forced in the defconfig as:
```
CONFIG_DECK_FORCE="bcServo"
CONFIG_DECK_SERVO=y
```
The servo is meant to be powered from VCOM on the Crazyflie/Crazyflie Bolt (mind the current limitations) or by an external 5V supply, e.g. from an ESC when using the CrazyflieBigQuad deck.

The default signal pin is TX2, but can also be configured to RX2, IO1, IO2, IO3 or MOSI via following config:
```
CONFIG_DECK_SERVO_USE_RX2=y
```

The servo signal wire can be connected directly, or via the BigQuad deck (where servo compatible connectors are already available)


knmcguire:
Hi! That's a great feature! We will need to find some time to review this and the coming weeks are a bit busy for us. Hopefully we will get to it soon!

src/deck/drivers/src/servo.c
```diff
@@ -0,0 +1,233 @@
...
+void servoSetAngle(uint8_t angle)
+{
+  // set CCR register
+  // Duty% = CCR/ARR*100, so CCR = Duty%/100 * ARR
+
+  double pulse_length_us = (double)(angle) / servo_range * (servo_MAX_us - servo_MIN_us) + servo_MIN_us;
+  double pulse_length_s = pulse_length_us / 1000000;
+  const uint32_t ccr_val = (uint32_t)(pulse_length_s * SERVO_PWM_PERIOD * SERVO_PWM_FREQUENCY_HZ);
+  servoMap->setCompare(servoMap->tim, ccr_val);
+  DEBUG_PRINT("Set Angle: %u deg, pulse width: %f us \n", angle, pulse_length_us);
```
tobbeanton, line 167: Maybe the print should only be active when debugging or optional? Will it not generate a lot of console printing otherwise?

tobbeanton (commented):
Thanks and very good contribution. I like that it has the same structure as the motor driver and that it is separated.

src/deck/drivers/src/servo.c
matejkarasek: In our use case the servo would only be used occasionally (e.g. close/open a gripper), but indeed it could generate quite some console prints in some cases (though it is not meant to be used in power distribution...)

Let me put it behind a debug flag...

knmcguire:
hi @matejkarasek, we'd like to test this out before merging so we need to find a bit of time somewhere, hopefully in the next few weeks

matejkarasek:
Thanks, yes, that makes sense!

src/deck/drivers/src/servo.c
```diff
@@ -0,0 +1,237 @@
...
+    .usedGpio = 0,
+  #endif
+  
+  .init = servoInit,
+  .test = servoTest,
+};
+
+DECK_DRIVER(servo_deck);
+
+PARAM_GROUP_START(servo)
```
knmcguire, line 228: If you want you can add a bit more documentation per parameter through doxygen:

https://www.bitcraze.io/documentation/repository/crazyflie-firmware/master/userguides/logparam/

src/deck/drivers/src/Kconfig
```diff
@@ -390,6 +390,37 @@ config DECK_OA
     help
         Codename: Obstacle Avoidance driver, obsolete deck.
 
+config DECK_SERVO
+    bool "Support the servo deck (prototype/via BigQuad deck)"
```
knmcguire, line 394: This seem to also be able to be driven from the Crazyflie Vcom/Tx directly right?

knmcguire (commented):
I've tested it with a servo with a crazyflie+breakout deck). I'm not sure if it is going correctly, since when I set an angle higher or lower than idle, it seem to not set  to go to that position but keeps rotating. It is supposed to do that, or was it supposed to just go to an angle and stay there? Probably I'll need to adjust the other parameters but it is difficult to get the meaning of them.

src/deck/drivers/src/servo.c
```diff
@@ -0,0 +1,237 @@
...
+
+PARAM_ADD(PARAM_UINT8 | PARAM_RONLY, servoInitialized, &isInit)
```
knmcguire, line 230: Ah I think the initialization parameter needs to be in the param group 'deck' : https://www.bitcraze.io/documentation/repository/crazyflie-firmware/master/api/params/#deck

matejkarasek:
> I've tested it with a servo with a crazyflie+breakout deck). I'm not sure if it is going correctly, since when I set an angle higher or lower than idle, it seem to not set to go to that position but keeps rotating. It is supposed to do that, or was it supposed to just go to an angle and stay there? Probably I'll need to adjust the other parameters but it is difficult to get the meaning of them.

The servo is supposed to go to the angle that you specify. How did you connect it? I now have a CF2.1 with a prototyping deck, and connected + to VCOM, - to GND and signal to PA2/TX2 and this works fine. My cf2_defconfig is:

```
CONFIG_PLATFORM_CF2=y

CONFIG_DECK_FORCE="bcServo"
CONFIG_DECK_SERVO=y
```

The default settings should work for most standard servos, I will add the description for the params, sorry, forgot to do that...

src/deck/drivers/src/Kconfig
matejkarasek: Yes, the default is TX**2**, here a photo of a working setup:
image

src/deck/drivers/src/servo.c
matejkarasek: Described in the latest commit

src/deck/drivers/src/Kconfig
matejkarasek: I meant that you would typically use the prototype deck or BigQuad deck to connect the servo

knmcguire: yeah or just change it to 'via a prototype deck'. I think that should be clear too

knmcguire:
I know what went wrong. I had a servo that was configured for rate control, so that is why it didn't work as expected. I now tried a different one and it worked like a charm!

So if you could add the doc to the parameters and change the initialization parameter to the deck group then it would be ready to merge

matejkarasek:
Ah ok, well this assumes servo in position mode :)

All the requested changes should be in the latest commit

knmcguire (approved):

knmcguire:
It includes all the important changes so I'll go ahead and merge it. Thanks @matejkarasek !

knmcguire:
If you'd like, you can make a small mention in the show&tell in discussion to notify people on this feature :) https://github.com/orgs/bitcraze/discussions/categories/show-and-tell

matejkarasek:
Awesome, thanks! Will post it there

matejkarasek:
The functionality is described here: https://github.com/orgs/bitcraze/discussions/1208

PR #1341, opened 2024-01-19
radiolink: add log variables for statistics

whoenig:
* radio.numRxBc counts the number of broadcasts received
* radio.numRxUc counts the number of unicasts received


whoenig:
An example use-case is the continuous end-to-end monitoring of the ack rate for both unicast and broadcast commands, see https://github.com/IMRCLab/crazyswarm2/pull/406 as an example. Here, we log the two new variables at 1 Hz and compare the received packet count with the sent packet count from the radio/connection.

knmcguire (approved):
Seems good to me! I'll merge

PR #1345, opened 2024-02-02
fixed the way of building apps with any disired config

atoums:
Hello, 
First, I am deeply sorry as i misunderstood the way building an app works. So the way of building an app, previously proposed from a PR, wasn't able to do the job it was supposed to, but didn't compromise the way of doing so that was already in place.
Using the previous Makefile file with CONFIG="" only compiled the firmware and not the app !

The modifications here are comments explaining the difference of building the apps.
the all rule allow to build an app with any config (eg. bolt) BUT requires the config file to be already included in the app-config file.
This might or not be desired.

In case not, the any rule concats the desired config file in a tmp oot-config file (eg. bolt_defconfig) before the app-config file
THEN calls the all rule.

NB. Binary files are also present in the app filesystem after compiling (*.o and *.o.cmd) (they were already from legacy) that are not caught by the clean rule. But that's not the subject of this PR.

### DIFF
--- a/tools/make/oot.mk
+++ b/tools/make/oot.mk
@@ -9,6 +9,7 @@ OOT ?= $(PWD)
 # build.
 #
 OOT_CONFIG ?= $(OOT)/oot-config
+ANY_CONFIG := $(OOT)/build/oot-config
 
 OOT_ARGS ?= -C $(CRAZYFLIE_BASE) OOT=$(OOT) EXTRA_CFLAGS="$(EXTRA_CFLAGS)"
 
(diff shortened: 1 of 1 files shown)


atoums:
The commit of the previous PR is : df96c57f6b7945ca4050057c120c9b33c9171397

Again sorry and have a nice week-end !

knmcguire:
alright, so what I understand from your comments, your issue with the current way is that the oot/app layer can only be build with the default config (CF2) and not with bolt config? 

Also, tt is probably best to add at least the comments to the docs of the app layer instead of the .mk file for more visibility: https://github.com/bitcraze/crazyflie-firmware/edit/master/docs/userguides/app_layer.md

gemenerik:
Fixed with #1398 - correct?

PR #1348, opened 2024-02-20
Added crashed state and the ability to recover from it

evoggy:
This adds a new crashed state which is entered when flying and tumbled. Instead of going into the locked state (which we cannot recover from) the new crashed state is possible to recover from when we're not tumbled anymore.


knmcguire (approved):
It seems to work nicely! I tried it with the other PRs as well.

We did have a discussion at the office that the current supervisor states and transitions are not very logical and clear at the moment so that perhaps requires a bit of an overhaul, but is not part of this PR.

PR #1349, opened 2024-02-21
Start timeout after landing before disarming

gemenerik:
Previous behavior: The system would disarm immediately upon landing.
New behavior: Upon landing, the system will remain in a landed state, maintaining its armed status and readiness for takeoff. A configurable timeout mechanism is introduced, triggering the supervisor state reset and disarming the system after the specified duration.

The timeout duration is a persistent parameter.

### DIFF
--- a/src/modules/src/supervisor.c
+++ b/src/modules/src/supervisor.c
@@ -261,8 +261,11 @@ static void postTransitionActions(SupervisorMem_t* this, const supervisorState_t
   }
 
   if (newState == supervisorStateLanded) {
-    DEBUG_PRINT("Landed, starting disarm timer\n");
     supervisorSetLatestLandingTime(this, currentTick);
+  }
+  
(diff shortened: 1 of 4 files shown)


src/modules/src/supervisor.c
```diff
@@ -235,13 +253,18 @@ static bool checkEmergencyStopWatchdog(const uint32_t tick) {
...
-static void postTransitionActions(SupervisorMem_t* this, const supervisorState_t previousState) {
+static void postTransitionActions(SupervisorMem_t* this, const supervisorState_t previousState, const uint32_t currentTick) {
   const supervisorState_t newState = this->state;
 
   if (newState == supervisorStateReadyToFly) {
     DEBUG_PRINT("Ready to fly\n");
   }
 
+  if (newState == supervisorStateLanded) {
+    DEBUG_PRINT("Landed, starting disarm timer\n");
```
knmcguire, line 264: We should complement this with another message that the platform is disarmed again after the timer is done.

PR #1361, opened 2024-04-03
Introduce BMP390 support

gemenerik:
There is no need to discern between the BMP388 and BMP390 besides the chip identifier and the printouts.
Hence the majority of this pr concerns renaming functions and variables to the more generic 3xx.


tobbeanton:
All good!

PR #1363, opened 2024-04-09
unit tests update for newer ruby versions

ToveRumar:


evoggy (approved):
Looks and and is tested

PR #1364, opened 2024-04-11
Storage; Defragment kve storage when starting up by default

ToveRumar:
The kve will be defragmented before running the kve sanity check. Defragmenting before checking will make sure that the kve is sane after defragmentation and if not the kve will be reformatted.

This can be turned off in the Kconfig, so that users that do not want this to possibly change their startup time (if eeprom is severely fragmented) can go back to only defragmenting once kve is full

### DIFF
--- a/src/hal/src/storage.c
+++ b/src/hal/src/storage.c
@@ -54,6 +54,14 @@
 #define KVE_PARTITION_START (1024)
 #define KVE_PARTITION_LENGTH (7*1024)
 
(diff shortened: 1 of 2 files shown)


ToveRumar:
What do you think of the placement of the config item?
I placed it in Parameter Subsystem, since that was what i think it is most related to.
I gladly take input on this

ataffanel:
Looks good, merging, thanks!

PR #1366, opened 2024-04-22
Documentation: Add instructions for workaround for deadlysignal in un…

ToveRumar:
…it tests.

There is a incompatibility between address sanitizer and recent ubuntu versions that causes an eternal loop of deadly signal when running unit tests in local environment. See this issue for more info https://github.com/google/sanitizers/issues/1716


tobbeanton (approved):
It all looks good to me!

PR #1378, opened 2024-05-29
Lee controller: fix inconsistend order h/c file

whoenig:
* h file had #ifdef CONFIG_CONTROLLER_OOT after Lee; c-file had it before
* Add define to force usage of Lee controller


knmcguire (approved):
Yes good catch. Thanks!

PR #1379, opened 2024-05-29
printf: add support to print NaNs

whoenig:


ataffanel:
Looks good, thanks! Merging.

PR #1380, opened 2024-05-29
Motors: unify types for PWM to be uint16_t

whoenig:
This uses the uint16_t type more consistently. For uSD logging this will reduce the memory needed. For CRTP logging it should not have any impact, since the user-code requests the logging variables with a desired type. Thus, "legacy" code might request the logging variable as uint32_t and it will be converted accordingly.

### DIFF
--- a/src/drivers/interface/motors.h
+++ b/src/drivers/interface/motors.h
@@ -322,9 +322,9 @@ void motorsBurstDshot();
 void motorsSetRatio(uint32_t id, uint16_t ratio);
 
 /**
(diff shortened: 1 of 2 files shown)


tobbeanton:
I'm trying to figure out if they where assigned uin32 in the first place. Looks like they originate from a pull-request and perhaps there is no real reason they are not uint16. For me it makes most sense to have them as uint16 so I'm approving the pull-request. Any other opinions?

ataffanel:
We have looked at it and it is a good change, no idea why it became 32 bits and it should not break anything to change it back to 16 bit. Thanks!

PR #1383, opened 2024-06-04
Add platform configuration for CF2.1 brushless

gemenerik:
- Platform comes with modified defaults for PID tuning, idle thrust, platform mass, and battery limit.
- The brushless defconfig (`make cf21bl_defconfig`) enables manual arming and the DSHOT ESC protocol.
- Moved idle thrust configuration out of kbuild.
- Forces manual arming for the brushless in kbuild.


src/platform/interface/platform_defaults_cf21bl.h
tobbeanton, line 1: As a start we could take the PID settings from the ICRA demo as defaults?

src/modules/src/Kconfig
```diff
@@ -150,16 +150,6 @@ config MOTORS_REQUIRE_ARMING
         When enabled, system must be armed to be able to take off. Arming can
         be done in several ways, e.g. though cfclient or external transmitter.
 
-config MOTORS_DEFAULT_IDLE_THRUST
```
tobbeanton, line 153: Is this supposed to be removed?

Kconfig
```diff
@@ -69,6 +69,12 @@ config PLATFORM_CF2
     select SENSORS_BMI088_BMP3XX
     select SENSORS_MPU9250_LPS25H
 
+config PLATFORM_CF21BL
+    bool "Build for CF2.1 brushless"
+    select SENSORS_BMI088_BMP3XX
+    select SENSORS_MPU9250_LPS25H
```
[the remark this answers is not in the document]
tobbeanton: To clarify, the SENSORS_MPU9250_LPS25H

src/platform/interface/platform_defaults_cf21bl.h
gemenerik: Yes, I have used those for this file

src/modules/src/Kconfig
gemenerik: Currently, having a parameter both defined in the platform defaults file AND in kconfig is incompatible. 
```
error: "CONFIG_MOTORS_DEFAULT_IDLE_THRUST" redefined [-Werror]
```

Given [our rule of thumb](https://www.bitcraze.io/documentation/repository/crazyflie-firmware/master/userguides/platform/#persistent-paramters) I thought it was more logical to not have it as a compile time parameter. With the implementation in this PR it still defaults to 0 for platforms that do not have it defined in their platform defaults and the value can always be changed through the parameter framework. 

I think this also covers the case for a custom build with the Bolt where the response to the idle thrust may be unknown, since it defaults to 0 for that platform.

What do you think?

Kconfig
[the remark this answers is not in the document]
gemenerik: Removed and tested OK

src/modules/src/Kconfig
tobbeanton: I think that makes sense!

src/platform/interface/platform_defaults_cf21bl.h
tobbeanton, line 1: I still think we should add the default tuning parameters that we have for the ICRA demo?

tobbeanton: Ahhh, OK, they seem a bit high to me but perhaps the brushless is that good!

PR #1384, opened 2024-06-11
Toverumar/mutual excl idlethrust

ToveRumar:
Idle thrust can not be higher than 0 if autoarm is on (i.e the device is unarmed from start).

This is due to the ESCs for the brushless motors (only BL needs idle thrust)  needs to receive a 0 signal before any other pwm signal to start. 
The ESC wakeup time is not something we keep track of in the Firmware and so to get these to features to work simultaneously would require some work. We do not see it as necessary to be able to do both as of now. 

For this reason we want to limit the use of these together. 
We limit them both at run time and at build time.

### DIFF
--- a/src/modules/src/power_distribution_quadrotor.c
+++ b/src/modules/src/power_distribution_quadrotor.c
@@ -35,6 +35,9 @@
 #include "math.h"
 #include "platform_defaults.h"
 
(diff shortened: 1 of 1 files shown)


ToveRumar:
What do you think about also limiting this for anything that is a quad but is not a brushless? This is how it is now. Should we check so that it will only limit it if its a brushless?

Issue #1389, opened 2024-07-01
Bolt: one motor doesn't spin


AdrienGuenard:
Hello,
I am facing a bug with recent version of firmware built for bolt config: 1 of the 4 motors doesn't spin. Has anyone else noticed the problem? I think that this problem may come from commit 3d5129ff6afbde9bfe12cc70cb0dfaf948423316

tobbeanton:
I can confirm the [problem](https://github.com/bitcraze/crazyflie-firmware/commit/3d5129ff6afbde9bfe12cc70cb0dfaf948423316#diff-f0379ea184fe2cefaa1e47311527c3a7f32d1b863f9c0a3ff682fe6110e74b96R246), it is the Crazyflie brushless ESC reset pin that conflicts with the Bolt M4 mosfet. We will try to have a fix soon.

tobbeanton:
A fix #1391 has now been submitted. Please try with that.

AdrienGuenard:
Thank you, it seems solved!

PR #1391, opened 2024-07-03
Fixed CF-BL and Bolt M4 PC15 conflict resulting in that Bolt M4 didn't spin.

tobbeanton:
The CF-BL uses PC15 for ESC-reset and Bolt M4 uses it for it's ESC enable MOSFET. hasPC15ESCReset variable was introduced to separate the configurations during init.

Fixes issue #1389

Tested on CF2.1-BL, Bolt and CF2.1


ataffanel (approved):
Looks good to me, it solves the problem.

PR #1392, opened 2024-07-11
Fix out of tree controllers

gemenerik:
Following the example from [this blogpost](https://www.bitcraze.io/2023/02/adding-an-estimator-or-controller/), the OOT controller is not being selected and defaults to PID. This PR fixes this by checking if CONFIG_CONTROLLER_OOT is defined.

Additionally, it adds the example from the blogpost to the examples dir.


examples/app_out_of_tree_controller/src/out_of_tree_controller.c
```diff
@@ -0,0 +1,79 @@
...
+
+
+// We still need an appMain() function, but we will not really use it. Just let it quietly sleep.
+void appMain() {
+  DEBUG_PRINT("Waiting for activation ...\n");
+
+  while(1) {
+    vTaskDelay(M2T(2000));
+
+    // Remove the DEBUG_PRINT.
```
ToveRumar, line 49: Are these lines still necessary or are the left from some testing?

gemenerik: This comes straight from the blog post. I think this has been commented out from the hello world example. I'll remove it.

ToveRumar (approved):

PR #1395, opened 2024-08-06
Building and flashing guide improvements

gemenerik:
Main goal of this PR is to clarify that brushless uses a different defconfig

Major changes:
- Add brushless defconfig tab
- Remove roadrunner defconfig reference

Minor changes: 
- Consistent code block formatting
- Retrieve number of processing units rather than assuming 12


knmcguire (commented):
There is unfortunately an issue, this will break `tb docs` of the toolbelt of building the documentation locally so it always need to be build as a submodule from the website. It won't know what the meaning of the tab group is otherwise

Another thing ofcourse is that now the markdown itself is less readable by itself. So unfortunately I do think that we need to discuss this break first before merging. We need to align on what is accepted within the repo doc files or not.

gemenerik:
> There is unfortunately an issue, this will break `tb docs` of the toolbelt of building the documentation locally so it always need to be build as a submodule from the website. It won't know what the meaning of the tab group is otherwise
> 
> Another thing ofcourse is that now the markdown itself is less readable by itself. So unfortunately I do think that we need to discuss this break first before merging. We need to align on what is accepted within the repo doc files or not.

The main point of this PR was to make it clear that for brushless people will need to use another defconfig. I removed the Jekyll liquid tags, let's discuss those another time!

PR #1397, opened 2024-08-08
Disable controller selection if OOT controller is selected

gemenerik:
Currently, kconfig allows both one of the implemented controllers + the OOT controller to be selected. Since only one can be used at a time this resulted in the implemented controller being used over the OOT controller (due to the order of how they are checked in `controller.c`. This PR keeps the config/menu structure the same, but when OOT controller is selected, this automatically disables the choice of regular controller.

Visualized in menuconfig:
Screenshot from 2024-09-09 13-12-08

Screenshot from 2024-09-09 13-12-02

### DIFF
--- a/src/modules/src/Kconfig
+++ b/src/modules/src/Kconfig
@@ -3,6 +3,7 @@ menu "Controllers and Estimators"
 choice
     prompt "Default controller"
     default CONFIG_CONTROLLER_AUTO_SELECT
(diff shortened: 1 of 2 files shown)


gemenerik:
Perhaps this is why it was named "default" controller? Can we remove the "default" now?

tobbeanton (approved):
Nice fix!

PR #1337, opened 2023-12-21
PID avoid derivative kick

gelanchez:
Addresses #1336

PR #1403, opened 2024-08-23
Fix PID derivative kickoff and reset

matejkarasek:
This PR fixes some unwanted consequences of #1337 

Since the D term now uses the difference of current and previous **_measurements_** (instead of **_errors_**) this introduced two issues:

1. for yaw, flying near 180deg yaw results in D spikes (measurements jumping between +180 and -180)

2. the pid internal measurement state is being reset to zero, which is not true for x, y, z, and attitude angles. This resulted in a D spike at the beginning of every flight

Both of these issues get especially noticable when LP filters are being used in the PID, as the spikes will propagate in time.

For 1), I removed the `pidSetError` functionality, as it was only used to deal with yaw discontinuity around +180/-180. 
Instead, an additional parameter in `pidUpdate` is now used to enable extra code inside the pid that deals with the discontinuity.

For 2), the reset now takes into account current state values (x, y, z, roll, pitch, yaw).

TODO: The reset in src/modules/src/crtp_commander_rpyt.c is not yet fixed, as I did not find an elegant way to read out the x,y,z states.

Still, even without the last TODO item, this should fix the majority of use cases...

### DIFF
--- a/src/modules/src/controller/controller_pid.c
+++ b/src/modules/src/controller/controller_pid.c
@@ -113,11 +113,11 @@ void controllerPid(control_t *control, const setpoint_t *setpoint,
     // behavior if level mode is engaged later
     if (setpoint->mode.roll == modeVelocity) {
       rateDesired.roll = setpoint->attitudeRate.roll;
-      attitudeControllerResetRollAttitudePID();
+      attitudeControllerResetRollAttitudePID(state->attitude.roll);
     }
     if (setpoint->mode.pitch == modeVelocity) {
       rateDesired.pitch = setpoint->attitudeRate.pitch;
-      attitudeControllerResetPitchAttitudePID();
+      attitudeControllerResetPitchAttitudePID(state->attitude.pitch);
     }
 
     // TODO: Investigate possibility to subtract gyro drift.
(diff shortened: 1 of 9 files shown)


src/utils/src/pid.c
```diff
@@ -73,15 +78,23 @@ float pidUpdate(PidObject* pid, const float measured, const bool updateError)
   * in the setpoint. By using the process variable for the derivative calculation, we achieve
   * smoother and more stable control during setpoint changes.
   */
-  float deriv = -(measured - pid->prevMeasured) / pid->dt;
+  float delta = -(measured - pid->prevMeasured);
+
+  if (isYawAngle){
```
gemenerik, line 83: Consider adding a comment to explain that this code prevents D spikes.  It’s amusing that the workaround for derivative kicks ended up introducing kicks in the derivative itself.

gemenerik: Perhaps include this in the note above

src/modules/src/controller/controller_pid.c
```diff
@@ -154,8 +154,8 @@ void controllerPid(control_t *control, const setpoint_t *setpoint,
     cmd_pitch = control->pitch;
     cmd_yaw = control->yaw;
 
-    attitudeControllerResetAllPID();
-    positionControllerResetAllPID();
+    attitudeControllerResetAllPID(state->attitude.roll, state->attitude.pitch, state->attitude.yaw);
+		positionControllerResetAllPID(state->position.x, state->position.y, state->position.z);
```
gemenerik, line 158: Formatting

gemenerik (changes requested):
I haven't observed the spikes in practice, though they seem logically possible. Thanks for your patience and for the PR.

src/modules/src/controller/controller_pid.c
matejkarasek: fixed

src/utils/src/pid.c
matejkarasek: I added a comment about the yaw spike fix

matejkarasek: > It’s amusing that the workaround for derivative kicks ended up introducing kicks in the derivative itself.

Thanks for the review, yes, indeed :D

PR #1408, opened 2024-09-09
Tumble check: platform-specific settings

matejkarasek:
Tumble-check settings (defining how sensitive the check is) were hard coded in `supervisor.c`.
This PR allows to make these settings platform-specific and override the default values in `platform_defualts_[platform-name].h` of each platform.

Further, this PR introduces a persistent parameter, with which the tumble check can be disabled.


src/modules/src/supervisor.c
```diff
@@ -297,6 +297,8 @@ static void postTransitionActions(SupervisorMem_t* this, const supervisorState_t
   }
 }
 
+uint8_t tumbleCheckEnabled = 1;
```
gemenerik, line 300: I would prefer to keep the `SUPERVISOR_TUMBLE_CHECK_ENABLE` in the platform defaults. We generally aim to keep anything configurable out of the source files (other than platform defaults), and this option can be valuable for those flying rate controllers (doing flips) or making extreme hardware mods.

src/modules/src/supervisor_state_machine.c
```diff
@@ -63,12 +63,6 @@ static const char* const conditionNames[] = {
 };
 static_assert(sizeof(conditionNames) / sizeof(conditionNames[0]) == supervisorCondition_NrOfConditions);
 
-#if SUPERVISOR_TUMBLE_CHECK_ENABLE
```
gemenerik, line 66: While the solution with the callbacks isn’t the most elegant, it serves its purpose for now.

gemenerik (changes requested):
Thanks for the PR. Nice addition.  

It would be nice if users can still disable the tumble check altogether in the platform defaults.

src/modules/src/supervisor.c
matejkarasek: No problem, putting it back

src/modules/src/supervisor_state_machine.c
matejkarasek: I returned SUPERVISOR_TUMBLE_CHECK_ENABLE to platform_defaults.
Since I would also like to keep the option to enable/disable the detection with a parameter (no need to build and flash the firmware), it is now handled in supervisor.c

matejkarasek (commented):
All should be resolved

gemenerik (approved):

PR #1415, opened 2024-10-08
Stop motors when stabilizer task is blocking

### LINKED ISSUE #1406: Stabilizer task blocks due to gyro failure, freezing flight control
When the gyro sensor stops sending data unexpectedly, the stabilizer task can become blocked [here](https://github.com/bitcraze/crazyflie-firmware/blob/master/src/modules/src/stabilizer.c#L286). This causes all processes dependent on the stabilizer to freeze. The rate supervisor task does not appear to be designed to detect a complete block in the stabilizer task, nor does it take any corrective action beyond printing a warning.

This issue was observed on a Crazyflie at the office, where the gyro sensor tends to fail mid-flight, particularly under high thrust conditions. When this occurs, the Crazyflie gets stuck at high thrust. @tobbeanton confirmed the gyro failure by observing a complete absence of data interrupts using an oscilloscope.

Other parts of the system, like USB connectivity and cfclient plotting, continue to function normally. 

Steps to Reproduce:
- Use a Crazyflie equipped with a gyro sensor that has a history of failing.
- Perform a flight with high thrust maneuvers.
- Observe if stabilizer-dependent tasks freeze when the gyro stops sending data.

Expected Behavior: The system should handle gyro sensor failures gracefully, preventing the stabilizer task from blocking and allowing dependent processes to cut the motors.

Actual Behavior: The stabilizer task blocks, causing all dependent processes to freeze, which can result in the Crazyflie being stuck at high thrust :see_no_evil:

gemenerik:
This PR introduces a new task, `rateSupervisorTask`, to monitor the `stabilizerTask` and detect if it is blocking. The `rateSupervisorTask` uses a semaphore to synchronize with the `stabilizerTask` and validate the loop rate. If the `stabilizerTask` is blocking, the `rateSupervisorTask` will stop the motors. This supervisor task also takes over the existing rate supervising that was previously done within the `stabilizerTask`. Fixes #1406.

Tested by sleeping in the supervisor task. 

We should discuss after how long this should step in. Currently at 2000 ms, but I think we could intervene more quickly.

### DIFF
--- a/src/config/config.h
+++ b/src/config/config.h
@@ -61,45 +61,46 @@
 
 
 // Task priorities. Higher number higher priority
-#define PASSTHROUGH_TASK_PRI    5
-#define STABILIZER_TASK_PRI     5
-#define SENSORS_TASK_PRI        4
-#define ADC_TASK_PRI            3
-#define FLOW_TASK_PRI           3
-#define MULTIRANGER_TASK_PRI    3
-#define SYSTEM_TASK_PRI         2
-#define CRTP_TX_TASK_PRI        2
-#define CRTP_RX_TASK_PRI        2
-#define EXTRX_TASK_PRI          2
-#define ZRANGER_TASK_PRI        2
-#define ZRANGER2_TASK_PRI       2
-#define LOG_TASK_PRI            1
-#define MEM_TASK_PRI            1
-#define PARAM_TASK_PRI          1
-#define PROXIMITY_TASK_PRI      0
-#define PM_TASK_PRI             0
-#define USDLOG_TASK_PRI         1
-#define USDWRITE_TASK_PRI       0
-#define PCA9685_TASK_PRI        2
-#define CMD_HIGH_LEVEL_TASK_PRI 2
-#define BQ_OSD_TASK_PRI         1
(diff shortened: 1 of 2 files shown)


ToveRumar (commented):
It looks like for example config.h has things that are not really changes checked in. Might be line ending differences?

ToveRumar:
> Perhaps it would be even better to just ASSERT when this happens?

Agree. I think asserting is a good idea in this case.

gemenerik:
> It looks like for example config.h has things that are not really changes checked in. Might be line ending differences?

I added `RATE_SUPERVISOR_TASK_PRI`, `RATE_SUPERVISOR_TASK_NAME`, `RATE_SUPERVISOR_TASK_STACKSIZE` and adjusted indentation of others to match the longer name. Diff doesn't seem to pick up on it nicely.

gemenerik:
> > Perhaps it would be even better to just ASSERT when this happens?
> 
> Agree. I think asserting is a good idea in this case.

With latest commit:

```
SYS: The system resumed after a failed assert [WARNING]
SYS: Assert failed at ../src/modules/src/stabilizer.c:274
```

Issue #1425, opened 2024-10-16
Manual disarming can trigger locked state


gemenerik:
Sending a disarm request before the CF has registered a landing triggers locked state.

PR #1426, opened 2024-10-17
Disarming improvements

gemenerik:
Previously, sending a "disarm" request while the drone was in an unexpected state—such as still being in flight mode when the user believed it had landed—could unintentionally trigger an emergency stop or crash state. This issue arises because the drone’s internal supervisor state is largely hidden from the user, leading to confusion when attempting to disarm in states that instead trigger an emergency stop.

This PR fixes the issue by ensuring that disarm requests no longer trigger an emergency stop or crash state. If a disarm command is sent before the system confirms the landing, it will not result in an emergency stop or system lock-up. However, users can still trigger an emergency stop through the correct command as intended.

In the event that users have been deliberately using the disarm command to trigger an emergency stop, they must now switch to using the dedicated emergency stop command.

This fix addresses issue #1425.

---

Furthermore, sending a disarm request while in landed state now allows you to skip waiting for the disarming timeout and disarm early.

### DIFF
--- a/src/modules/src/supervisor.c
+++ b/src/modules/src/supervisor.c
@@ -264,7 +264,7 @@ static void postTransitionActions(SupervisorMem_t* this, const supervisorState_t
     supervisorSetLatestLandingTime(this, currentTick);
   }
 
-  if ((previousState == supervisorStateLanded) && (newState == supervisorStateReset)) {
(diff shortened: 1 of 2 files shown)


evoggy (approved):

tobbeanton (approved):
All good now!

Issue #1354, opened 2024-03-22
Implement and document linking a lib with the client


ataffanel:
Before k-config it used to be possible to easily link add a static lib to be linked in the Crazfile firmware. This allowed, among other thing, to expand the firmware with Rust code.

It does not seems easy to properly add static libs to kconfig. This needs to be investigated and documented.

Related to https://github.com/orgs/bitcraze/discussions/1236

ataffanel:
#1429 gave a good practical example for it.

Issue #1404, opened 2024-08-27
“crazyflie-firmware” compilation chain


CasperLiZ:
Hello, I am working on a project based on CF2.1. I need to add a static library file "xxxx.lib". Based on the current compilation chain of crazyflie-firmware, how should I do it? Thank you very much.

whoenig:
Also interested in this, see https://github.com/orgs/bitcraze/discussions/1236.

ataffanel:
Hi, this is something we need to look at. I have scheduled it for triage so we can organize to look at it.

ataffanel:
We are planning to look at this but not for the next upcoming release. Stay tuned :smile:

ataffanel:
Fixed by pr #1429

PR #1429, opened 2024-10-29
Add POC Rust Crazyflie app example

ataffanel:
This PR adds a Proof Of Concept Rust Crazyflie hello world app.

Some creativity was needed to link the Rust-generated static lib in the Crazyflie firmware. This implementation could be used as a base to fix issues #1354 and #1404.


ataffanel:
I unfortunatly do not have access to a Crazyflie at the moment so all I could test is that it compiled and linked. This is the same code as in the previous example (at https://github.com/ataffanel/crazyflie-app-hello-rs), so the rust part should be OK. Though, it definitly needs to be tested on a real CF before merging!

whoenig (approved):
I tested this PR. Both compilation (Ubuntu 24.04, rust 1.82.0) and test (connecting via CFClient to a Crazyflie 2.1 with the built firmware) work fine and the text shows up on the console.

Code looks good to me as well.

ataffanel:
Thanks for testing! I am merging it.

PR #1433, opened 2024-11-01
Remove the RPM test from the exp  deck driver completely

ToveRumar:
Let the driver initialize the RPM deck so it can be used and then just check pins


tobbeanton (approved):
looks good!

PR #1441, opened 2024-12-05
Automatically disarm after N seconds without flying

gemenerik:
This PR introduces automatic disarming of the Crazyflie if no flight is started within 30 seconds after arming. This time-out is configurable. Implementation largely identical to post-flight auto-disarming. 

Since we currently print to console when ready to fly and when disarming, this introduces "spam" to the console when using auto-arming. Considering removing these printouts.

Possible discussion points:
- Trigger disarm after how many seconds?
- Removing disarming and ready to fly debug prints.


ataffanel (changes requested):
The change looks good to me. 30s might be a good time to start with.

About the spamming: what about not printing arming states only when auto-arming is enabled. We can just write in the console that auto-arming is enabled and nothing else.

Otherwise this looks good to merge to me!

PR #1444, opened 2024-12-12
Added ESC bootloader wakeup signal.

### LINKED ISSUE #1443: ESC-configurator does not find M4 after flashing BLHeli_M
When following [this](https://www.bitcraze.io/documentation/repository/crazyflie-firmware/master/userguides/advanced-configuration/ledring_on_cf21bl/) guide the ESC-configurator does not find M4 after flashing BLHeli_M. It works to flash back the original ESC FW and then M4 appears again.

The couse is a timeout in the ESC to put them to sleep and for some reason BLHeli_M is slower at responding so M4 goes to sleep before it has been addressed.

tobbeanton:
This PR will send a low pulse for 1 ms after the ESC has entered the bootloader to keep the bootloader alive and not go to deep sleep. This only happes then 4way interface has been enabled though a client such as the ESC-configurator and will not prevent the ESC from sleeping under normal operation.

Fixes #1443

### DIFF
--- a/src/modules/src/serial_4way.c
+++ b/src/modules/src/serial_4way.c
@@ -33,6 +33,10 @@
 #define USE_SERIAL_4WAY_BLHELI_BOOTLOADER
 #ifdef  USE_SERIAL_4WAY_BLHELI_INTERFACE
 
+//FreeRTOS includes
+#include "FreeRTOS.h"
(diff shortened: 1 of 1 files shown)


ataffanel:
LGTM, merging!

PR #1446, opened 2024-12-30
Autofly

shang-huan:


ToveRumar:
Hi @shang-huan !

I just wanted to check if this merge request is something you would like to merge into the main trunk of the crazyflie-firmware.
If so we would need a clearer explanation of what the MR is intended to do.

shang-huan:
I'm sorry, it should have been submitted incorrectly.

PR #1456, opened 2025-02-11
Revert " #338: Fix yawrate sign in commander_generic packets"

gemenerik:
This reverts commit f6c0e737cfd061b72efcb125af7d545e06597807.

Requires using correct yaw convention in cflib: https://github.com/bitcraze/crazyflie-lib-python/pull/516

### DIFF
--- a/src/modules/interface/crtp.h
+++ b/src/modules/interface/crtp.h
@@ -30,7 +30,7 @@
 #include <stdint.h>
 #include <stdbool.h>
 
-#define CRTP_PROTOCOL_VERSION 8
+#define CRTP_PROTOCOL_VERSION 9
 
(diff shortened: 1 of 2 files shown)


gemenerik:
Latest commits to this and https://github.com/bitcraze/crazyflie-lib-python/pull/516 ensure compatibility when updating either the Crazyflie or lib independently. The lib now checks the Crazyflie's CRTP version and sends either the legacy packet (with the original negated yaw) or the new packet type (with the corrected yaw sign). This allows for a smooth transition without breaking existing functionality. The legacy types will be deprecated in the future.

ToveRumar (approved):
Looks good to me. 

Tested it with and without cflib change. Tested by checking the yaw in cfclient.

Issue #1445, opened 2024-12-20
Appchannel MTU inconsistent with CRTP max packet size 


TheCodeinator:
In https://github.com/bitcraze/crazyflie-firmware/blob/master/src/modules/interface/app_channel.h the maximum MTU of an Appchannel packet is defined as 31 bytes. However both in the cflib client library as well as the crtp firmware implementation the maximum crtp MTU is defined as 30 bytes. Therefore for example using the appchannel API in the cflib to send a packet with exactly 31 bytes of payload size will give an error. In my opinion this is a bug  and the appchannel MTU should be reduced to 30 bytes in both firmware and documentation. Am I missing something or is this indeed inconsistent?

Have a nice holiday season

ataffanel:
This should indeed be consistent! We are going to have a look.

PR #1460, opened 2025-03-10
The max size of the data in crtp is 30 not 31.

ToveRumar:
Fix this so that documentation and code matches.
In the lib the max size is set to 30.

Fixes https://github.com/bitcraze/crazyflie-firmware/issues/1445


ToveRumar:
@ataffanel has looked at this PR and approved it. Will merge it

PR #1462, opened 2025-03-11
Add docs that you could disable the bat_compensation 

ToveRumar:
If you use 2.1+ or thrust upgrade you should probably disable it


ArisMorgens:
Looks good!

PR #1371, opened 2024-04-24
autogen log and param fixes

knmcguire:
Fixes some of the missed log and param descriptions that were ignored by doxygen due to` #if `statements.

I chose to just remove all of the ifstatements related to the maximum number of TWR as that define default has remained at 8 for years, and no one complained for the last 2 years since it wasn't converted to kbuild properly, so let's just assume that this is okay. 

fixes #1323

PR #1464, opened 2025-03-21
Make number of loco nodes configurable again, and overhaul Doxygen handling

### LINKED ISSUE #1373: unhardcode loco pos anchors TWR
As of this PR, the old config of setting a maximum amount of loco anchors for TWR is now no longer configurable but hardcoded. This might cause bugs and problems in the future and we should fix this to `CONFIG_DECK_LOCO_NR_OF_ANCHORS `

Mind that we also have to make sure that the logs depended on this are still readable for doxygen to generate autodoc. Check this PR to see what fixed that #1371

gemenerik:
This PR does two major things.

First, it restores the configurability of the number of loco nodes through kbuild. This functionality had been broken, and this change brings it back, allowing the number of nodes to be adjusted at build time as originally intended. This is essentially a revert of #1371, or at least largely brings things back to how they were before that change.

Second, and more extensively, it fixes how Doxygen handles kbuild configuration. Previously, Doxygen lacked access to the autoconf.h definitions generated by kbuild, which meant it couldn't evaluate #ifdef blocks set by kconfig correctly. As a result, any code or documentation behind those conditionals was silently skipped. To solve this, the Doxygen configuration now mirrors the relevant include paths from the Makefile, allowing it to see the configured values. This makes the generated documentation much more accurate and complete. It is dependent on the configuration however

One downside of this approach is that the documentation output becomes configuration-dependent. However, this is still a big step forward—before this change, those parts simply weren’t documented at all.

While working on this, I ran into an overwhelming number of Doxygen warnings—likely thousands—which made it clear that much of the documentation was broken or missing. This PR includes a large cleanup pass fixing many of those warnings. To prevent this from happening again, Doxygen is now configured to treat warnings as errors.

Fixes #1373


gemenerik:
Perhaps this is also a good time to check it is clear that this limit does not apply to TDOA3

ToveRumar (approved):
Looks good to me

PR #1465, opened 2025-03-21
New method for battery compensation

Rather1337:
Changed the battery compensation to go from V_bat to thrust with a third order polynomial. Adjusted the system id scripts accordingly.

Added functionality to change the battery compensation at compile time (type is set in defconfigs).

Open questions:
- Sysid readme mentions some emergency stop branch in cflib. Seems to not exist anymore
- In motors.c: motorsCompensateBatteryVoltage(), it might make sense to either make batCompEnable as a parameters s.t. it can be changed on the fly. In that case, remove the battery compensation setting in the defconfig.
- In motors.c: motorsCompensateBatteryVoltage(), removing the batVoltage < 2 check might make sense. Other code should catch a case like that and the battery low indication is also in place.


ToveRumar:
Hi!

Thank you so much for your contribution and for pushing this PR so quickly!

As we discussed, we will need to decide how the battery compensation method should be set—whether via Kconfig or parameters. We still need to discuss this internally and will get back to you on it.

In the meantime, it would be great if you could clean up this PR so that it only includes the intended changes. It looks like some Python files were removed—was this intentional? Additionally, there are some formatting inconsistencies that should be addressed. For example, in motors.c, the indentation level differs from the original file, making it appear as if more changes were made than actually were. There are also general formatting adjustments needed, such as pointer asterisk placement and spacing after // comments.

Cleaning this up would really help during the review process and make collaboration on the PR smoother.

Thanks again for your hard work—it’s much appreciated!

Rather1337:
Hi,

I've reverted all the formatting changes done by VSCode. Now only the actual changes are part of the PR. If there is anything else regarding the formatting, please let me know.

Regarding the deleted files: Yes, I've combined some code into a single file, since the base communicator is always the same for all data collection. I believe that makes it easier to understand. I hope that's okay.

There are some more things I would like to discuss here before we finish the PR:
- As mentioned in the meeting, the battery measurement is low pass filtered before it's used for the compensation. The core idea is to avoid oscillations. However, in my experience, the time constant had to be set quite low for the setup to work properly. This probably slows down the motor speed convergence a bit. 
- Just in the couple days I've realized the following. Right now, the motor forces dynamics $\dot{f}=\frac{1}{\tau}(f_{cmd}-f)$ are directly fit instead of using the propeller speed dynamic $\dot{\omega}=\frac{1}{\tau}(\omega_{cmd}-\omega)$. The latter should be the proper approach as far as I understand now, but I'm not entirely sure. The data fits the force dynamic better though. Should I add a note regarding that? Do we want to change it to propeller speeds?

BR

Rather1337:
Hi @ToveRumar,
I've just seen that in the `powerDistributionForceTorque` in `power_distribution_quadrotor.c`, the quadratic fit is still used. With the new setup, that should also be changed to:
`motorThrustUncapped = motor_force / thrust_max * UINT16_MAX;`
In that case, it would also make sense to move the constants out of `motors.c` into some config file, like `platform_defaults_cf2.h`, s.t. there is a proper global definition.
Should I add those changes to the PR?

tobbeanton:
> Should I add those changes to the PR?
Yes, they all sounds like good changes

spencerfolk:
Hey @Rather1337 thanks for contributing. In a separate [discussion post](https://github.com/orgs/bitcraze/discussions/1879#discussioncomment-12801437) I was asking about ways to get an accurate measurement of thrust from the motors for state estimation purposes. It seems the best I can do right now is to infer the thrust from the 16-bit commanded pwm signal to each motor as well as the supply voltage. 

@tobbeanton pointed me to this PR. It looks like you might have these mappings [here](https://github.com/bitcraze/crazyflie-firmware/blob/a73d34dabc2cb14be2a109c7674452efcd2003ef/src/drivers/src/motors.c#L55-L69). Although I looked at the files and it looks like you only have data for the brushed CF. Do you happen to have data for the brushless Crazyflie? I previously conducted thrust stand tests for the isolated brushless CF motor/prop, but I wasn't using the same ESC and I'm concerned that it won't be an accurate fit since each ESC is different.

Rather1337:
Hey @spencerfolk,
unfortunately, we currently don't have access to the brushless version, but we would be happy to take the measurements with our setup. @tobbeanton does the sysid deck also work with the brushless cf?

tobbeanton:
@Rather1337 The system-id deck works with the brushless as well. If you can not get hold of a brushless let us arrange one for you!

Rather1337:
@tobbeanton that would be great! Good to hear that the deck works too.

Rather1337:
I've now also updated the power [distributionForceTorque](https://github.com/bitcraze/crazyflie-firmware/pull/1465/files#diff-f6eaf525d7afa3ea43692bb0c53829cd93cdcd5d163831e889779e6c473807bbL103). If you are happy with the way I've structured the constants, the PR is almost ready from my side. All the tests are passing now. One thing that is still a to do is the "Emergency Stop branch" mentioned in the old [readme](https://github.com/bitcraze/crazyflie-firmware/pull/1465/files#diff-ce566808522910df79cee8d777f9d816a59cde4a5a220685ea952b755342b41fR27). This would be nice, since right now pressing CTRL-C makes the drone getting stuck in the current state. Also, as soon as we get hands on a brushless drone and our 6 axis load cell arrives, we can identify the missing parameters (mostly thrust->torque). 

To make sure I understand this correctly: Since for brushless motors, the speed does not depent on the battery voltage, but only the commanded PWM, we only need to identify the PWM-> Thrust curve instead of the v_motor->thrust. So I assume that will be a different PR then?

ArisMorgens:
Hello @Rather1337!

We finally managed to test your PR with our Crazyflies and the results seem pretty good. Here is what we did:

### Test 1: Hover with the Lighthouse deck
We used 4 Crazyflies:
- 1 with old propellers (2.1) and the PR
- 1 with new propellers (2.1+) and the PR
- 1 with the thrust upgrade kit and the PR
- 1 with the thrust upgrade kit without the PR (2025.02 release fw)
All of them behaved normally and started drifting only when their battery was low.

### Test 2: Hover with the Flow deck v2
With the same setup, their behavior was pretty normal.

### Test 3: Lifting weights
We used 2 Crazyflies:
- 1 with the thrust upgrade kit and the PR
- 1 with the thrust upgrade kit without the PR (2025.02 release fw)
The Crazyflie with the PR managed to carry a weight that the Crazyflie with the release fw couldn't. The difference was not that big but still noticeable.

To sum up, the PR seems good in action. In the following days, we will take a look at the structure and might propose some necessary changes.

ArisMorgens:
That's a nice catch @jglane! `0.027f` is not correct either since Crazyflie 2.1 and 2.1+ weight **29g**.
`CF_MASS` is currently used only by these three controllers: Mellinger, Brescianini and Lee. This means that we couldn't notice the difference using the default PID controller. However, it would be helpful to have a precise mass parameter for the different stock Crazyflies (**29g** for 2.1 and 2.1+, and **32.5g** for the thrust upgrade kit).

@Rather1337 Could you explain how you measured the `CF_MASS`? I assume that `0.03454f` results from a Crazyflie with the thrust upgrade kit and maybe a lighthouse deck?
This value is set in the default parameters and is now used by every Crazyflie 2.x drone. I think we should change that since it's not true. A possible solution would be to have the `CF_MASS` equal to `0.029f` as the default value and define it equal to `0.0325f` (or add the difference) when setting the `THRUST_BAT_COMPENSATION_TYPE`  to `2` (meaning that thrust upgrade kit is used). What are your thoughts on that?

Rather1337:
Thanks for bringing that up @jglane! I've double checked and the stock Crazyflie with battery weighs indeed only 30g (on our imprecise scale). So 29g sounds good. The 34.54g is some number that has been floating around in the lab for a stock cf2 with marker deck and markers. I've never checked the value exactly, since we currently don't have a high precision scale, but our scale displays a value around 34g. I thought the 27g was the weight of the drone without the battery - which is roughly true if you have the marker deck on, see [here](https://store.bitcraze.io/products/250mah-lipo-battery?variant=39540518748247) - and didn't think the marker deck would make such a large difference, but it does apparently. Sorry for letting that slip!

@ArisMorgens I agree that we should change the `CF_MASS` in accordance to the battery compensation type and change it to **29g** for the stock cf2. Maybe some more precise measurement is possible there? 

I'm wondering how one is intended to change the weight of the drone depending on the used decks? It would be nice if the mass was a parameter you can change online. I guess the alternative would be to change the mass parameter in the `platform_defaults` before flashing or somehow in the config file (maybe setting the additional mass, i.e. payload, added to the drone).

jglane:
Glad I could help! What you said makes sense. In my humble opinion I don't think it's necessary to change the mass parameter online because all the controllers should have some sort of disturbance rejection like an integral term implemented. So the mass value just needs to be close to what it actually is and the controller will compensate for the rest.

If it helps, my lab has a precise scale and I measured the stock Crazyflie 2.1+ at 28.3g and with a passive marker deck with 4 markers at 31.1g.

ArisMorgens:
Hello everyone! Depending on the components used (e.g. batteries [1](https://store.bitcraze.io/collections/accessories/products/250mah-lipo-battery?variant=39540518748247), [2](https://store.bitcraze.io/collections/accessories/products/350mah-lipo-battery), or deck connectors [1](https://store.bitcraze.io/collections/spare-parts-crazyflie-2-0/products/male-deck-connector), [2](https://store.bitcraze.io/collections/spare-parts-crazyflie-2-0/products/male-long-deck-connector)) the weight of the Crazyflie changes significantly. Our measurements for the stock Crazyflie 2.x were between **28.1g** and **29.9g**, so I think a mean value of **29g** is good for now. Do you agree with that @jglane?

@Rather1337 Changing the weight depending on the used decks is a great idea that we've discussed in the company. Our long distance goal is to automatically modify the mass value depending on the detected decks and that's why we prefer the `CF_MASS` to refer to the stock Crazyflie. I think a way forward is to set `CF_MASS` to `0.029f` and increase it to `0.0325f` when using the thrust upgrade kit. Do you agree with the idea and would you like to do that?

jglane:
Hi @ArisMorgens, yes that sounds good to me.

Rather1337:
@ArisMorgens that works for active decks only, so not for the passive marker deck. Do you have a suggestion on how to include that? To be frank, it doesn't really affect us that much, since we are only using attitude controllers for high performance tasks. However, it would be a nice improvement for the onboard position controllers, which we use for swarm applications.

I can create a new PR with the fix for the mass!

ArisMorgens:
@Rather1337 All the decks (including the passive marker deck) have a 1-wire memory in order to be recognized by the Crazyflie. The battery holder and the breakout deck are the only exceptions. The prototyping deck has also a 1-wire memory that the user can write. Making the mass a parameter sounds like a good idea to me. Since this is an important change, we will discuss it internally next week and we will get back to you as soon as possible.

ArisMorgens:
Hello @Rather1337!
I opened [this issue](https://github.com/bitcraze/crazyflie-firmware/issues/1503) regarding the `CF_MASS` parameter. Let us know if you would like to change something on the proposed solution and if you're interested in working at it.

Issue #1388, opened 2024-06-28
Add rpyt setpoint handling to generic commander


knmcguire:
We'd like to add some rpyt manual setpoint handling to the generic commander for better compatibility with the other setpoints.

This is connected to this cflib issue: https://github.com/bitcraze/crazyflie-lib-python/issues/461

PR #1470, opened 2025-04-01
Added new packet type to handle manual flight setpoints

ArisMorgens:
Currently, there are no generic commander setpoints, meant for manual flight.
Also, the current way of choosing between Angle and Rate mode for roll and pitch is a bit legacy. More specifically, you have to change the parameters in [crtp_commander_rpyt.py](https://github.com/bitcraze/crazyflie-firmware/blob/856d10ce4b9842cc7bdc7ed8bfdc7734991b1a49/src/modules/src/crtp_commander_rpyt.c#L71C1-L73C100) and re-flash the Crazyflie.

I added a new type of packet in the generic crtp commander, called manualType.
Depending on the `rate` variable, the manualType packet can switch the roll and pitch mode from Angle to Rate, regardless of the parameters set in [crtp_commander_rpyt.py](https://github.com/bitcraze/crazyflie-firmware/blob/856d10ce4b9842cc7bdc7ed8bfdc7734991b1a49/src/modules/src/crtp_commander_rpyt.c#L71C1-L73C100).

This is based on Library issue [#461](https://github.com/bitcraze/crazyflie-lib-python/issues/461) and Firmware issue #1388

### DIFF
--- a/src/modules/src/crtp_commander_generic.c
+++ b/src/modules/src/crtp_commander_generic.c
@@ -74,6 +74,7 @@ enum packet_type {
   velocityWorldType       = 8,
   zDistanceType           = 9,
   hoverType               = 10,
+  manualType              = 11,
 };
 
 /* ---===== 2 - Decoding functions =====--- */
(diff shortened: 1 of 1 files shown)


ataffanel (approved):
LGTM, good job!

PR #1471, opened 2025-04-02
Updated Propeller Test in cfclient

ArisMorgens:
This is based on [this](https://github.com/bitcraze/crazyflie-clients-python/issues/744) client issue.

I added a float persisted parameter called "`propTestThreshold`" that is 0 by default.
When running the propeller test in cfclient, there are 2 options:
1. If `propTestThreshold == 0`, the console prints the acceleration variance of each motor.
2. If `propTestThreshold != 0`, the console also prints `[PASS]` if the measured value is within threshold, and `[FAIL]` if the measured value is outside the threshold.

### DIFF
--- a/src/config/config.h
+++ b/src/config/config.h
@@ -218,7 +218,7 @@
  * This is the threshold for a propeller/motor to pass. It calculates the variance of the accelerometer X+Y
  * when the propeller is spinning.
  */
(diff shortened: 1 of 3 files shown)


tobbeanton:
Looking good! One addition though, I suggest we also add it as a default platform parameter so we could have different defaults for different platforms if one likes.

tobbeanton (approved):

PR #1475, opened 2025-04-14
Fixed enabling-the-BigQuad issues

### LINKED ISSUE #1467: Flying the BigQuad with a controller
This issue resulted from [this](https://github.com/orgs/bitcraze/discussions/1829) discussion.
We tried flying the Crazyflie with the BigQuad and the lighthouse deck attached, using the **flight control** tab on the cfclient and a controller. We tested this on Ubuntu 24.04.2 with a cfclient both from the latest release ([2025.2](https://github.com/bitcraze/crazyflie-clients-python/releases/tag/2025.2)) and from [source](https://github.com/bitcraze/crazyflie-clients-python/tree/96e098d0ece64bdd141ec475afaa59f56c397dc0).

We observed the following behaviour:
1. Flying with a controller works perfectly in any case.
2. While there is a controller connected, we can Arm the BigQuad through the client but can't press the **flight control** buttons.
3. After flying with a controller, we can press the **flight control** buttons but the BigQuad won't take off. Also, to do that, the cfclient needs to reboot. This is the same behaviour observed in [this](https://github.com/bitcraze/crazyflie-clients-python/issues/697) cfclient issue.
4. If we want the BigQuad to fly with the **flight control** buttons after flying with a controller, it needs to reboot.

This seems like a priority issue between the controller and the flight control tab.
Also, the BigQuad should be able to fly in any case without rebooting.

ArisMorgens:
It was observed in [this discussion](https://github.com/orgs/bitcraze/discussions/1829) and in [this forum](https://forum.bitcraze.io/viewtopic.php?t=3931) that the external CPPM was taking priority over the other inputs, meaning that if CPPM wasn't used, the BigQuad would not take off.
This PR fixes this issue, adding a CPPM option in Kconfig.

Also, the idleThrust is now set to 7000 when the BigQuad is attached, as it might be dangerous not to have an indication when the platform is armed or not. The user should now be encouraged to use arming.

This fixes #1467.

### DIFF
--- a/src/deck/drivers/src/Kconfig
+++ b/src/deck/drivers/src/Kconfig
@@ -85,6 +85,13 @@ config DECK_BIGQUAD_ENABLE_PM
     bool "Enable Bigquad deck PM"
     default n
 
+config DECK_BIGQUAD_ENABLE_CPPM
+    depends on DECK_BIGQUAD
+    bool "Enable Bigquad deck CPPM"
+    default n
+    help
+      This enables external CPPM input.
+
 config DECK_BUZZ
     bool "Support the Buzzer deck"
     default y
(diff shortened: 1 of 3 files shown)


tobbeanton (approved):

tobbeanton:
Nice to have the BQ-deck issue fixed!

PR #1483, opened 2025-06-05
Enable kalman estimator attitude reversion based on deck requirements

krichardsson:
This PR adds functionality to enable/disable the attiude reversion (aka roll/pitch zero reversion, aka bleed to zero) based on the requirements of the installed decks. This functionality is mainly interesting for the Loco and flow decks (as far as I know) and is more of a problem for other decks. 
The functionality is only lightly tested (meaning you better test before merging) as I don't have the required hardware.

### DIFF
--- a/src/deck/drivers/src/activeMarkerDeck.c
+++ b/src/deck/drivers/src/activeMarkerDeck.c
@@ -204,6 +204,7 @@ static const DeckDriver deck_info = {
   .vid = 0xBC,
   .pid = 0x11,
   .name = "bcActiveM",
+  .requiredKalmanEstimatorAttitudeReversionOff = true,
 
   .init = activeMarkerDeckInit,
   .test = activeMarkerDeckTest,


gemenerik:
Would it maybe make sense to flip the logic here? That is, disable zero reversion only when a deck provides full pose (motion capture or Lighthouse), instead of enabling it based on Loco or Flow being present.

This approach removes reversion even for deck-less setups, but shouldn’t we still want it in that case? Assuming the drone starts level seems fair, and without reversion, gyro drift on pitch and roll might become significant over time.

I get that combining Loco or Flow with full-pose systems is an edge case, but people do use such setups in occlusion-heavy environments or fallback scenarios. That seems worth considering.

I haven’t had a chance to test this yet, but that aside, I think this is a solid architectural improvement regardless of how we resolve the points above. It’s definitely a welcome addition.

krichardsson:
@gemenerik Good point! It would make the behavior more similar to the current system for most use cases.

One issue is perhaps that the CF often is used without a deck when positioning with mocap, I usually just glue markers on the motor mounts. This means that there is no deck mounted that will trigger the disable of the reversion. On the other hand, active attitude reversion has not been a huge problem so far.

Let me know if you want me to reverse it.

gemenerik:
> I reversed the logic, but did not add a config for it.

Can you push the commit(s)?

gemenerik (approved):
Thanks, nice work.

PR #1484, opened 2025-06-05
Add general purpose mode to kbuild for the kalman estimator

krichardsson:
In the general purpose mode the estimator is always assuming that it is NOT flying, that is it does not use any drone specific behaviors. The process noise is also set to the same value for x, y and z as we don't expect any asymmetry due to motor noise.


krichardsson (commented):
Updated docs as discussed

PR #1491, opened 2025-06-23
Add streaming service for matched LH samples

krichardsson:
This PR adds a new streaming service that is similar to the old LH_ANGLE_STREAM but instead of simply sending all angles over the radio samples are grouped based on closness in time. The API also supports requesting a specific number of samples instead of continous sending. 
Hopefully this will solve a couple of problems:
1. Not over flowing the radio when many base stations are visible 
2. Better mathing in time compared to the current solution where grouping is done on the client side with limited timing information. 

I using a prameter in a slightly dodgy way by using it as a decrementing counter. This usecase is not supported by the parameter framwork but is similar to the self resetting triggers that are used elsewhere. This use case could be seen as input for a future improved param/RPC. framework

### DIFF
--- a/src/modules/src/crtp_localization_service.c
+++ b/src/modules/src/crtp_localization_service.c
@@ -88,6 +88,19 @@ typedef struct {
   } __attribute__((packed)) sweeps [NBR_OF_SWEEPS_IN_PACKET];
 } __attribute__((packed)) anglePacket;
 
+typedef struct {
+  uint8_t type;
+  uint8_t baseStation;
+  struct {
+  float sweep;
+    struct {
+      uint16_t angleDiff;
+    } __attribute__((packed)) angleDiffs [NBR_OF_SENSOR_DIFFS_IN_PACKET];
+  } __attribute__((packed)) sweeps [NBR_OF_SWEEPS_IN_PACKET];
+
+  uint8_t group_id_and_bs_count; // 4 bits group id, 4 bits base station count
+} __attribute__((packed)) matchedAnglePacket;
+
 // up to 4 items per CRTP packet
 typedef struct {
   uint8_t id; // last 8 bit of the Crazyflie address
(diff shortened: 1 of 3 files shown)


gemenerik (dismissed):
Looks good and does not impact current functionality. I don't have strong opinions on usage of parameters, let's see if @ataffanel has any thoughts.

src/modules/src/crtp_localization_service.c
```diff
@@ -363,31 +390,107 @@ void locSrvSendRangeFloat(uint8_t id, float range)
...
+        for (int i = 0; i < lhMatchedStreamBsCount; i++) {
+          matchedAnglePacket *ap = (matchedAnglePacket *)lhMatchedAnglePackets[i].data;
+          ap->group_id_and_bs_count = (lhMatchedStreamGroupId << 4) | (lhMatchedStreamBsCount & 0x0F);
+          // This is best effort, i.e. the blocking version is not needed
+          crtpSendPacket(&lhMatchedAnglePackets[i]);
+        }
+
+        lhMatchedStreamGroupId++;
+        if (lhMatchedStreamGroupCount > 0 && lhMatchedStreamGroupCount != 255) {
+          lhMatchedStreamGroupCount--;
```
ataffanel, line 465: This could be implemented using the `paramSetInt()` function : https://github.com/bitcraze/crazyflie-firmware/blob/03adf38b3bc1cae7f22e0cb7c62320f0f43b014a/src/modules/interface/param_logic.h#L117-L124.

This would send a packet to the ground each time the value is modified. If that is not detrimental to the usecase it would be preferable. If so many downlink packets is a problem though, I would suggest at least sending a notification when the value reaches 0.

krichardsson: The main goal of this new functionality was to reduce the number of packets that are sent so I was a bit reluctant to update the parameter. Doing it once when reaching the end seems as a good compromise though, I'll do that.

ataffanel (approved):
Thanks! That makes it much more observable from the ground :).

PR #1494, opened 2025-06-25
Updates to the Lee Controller

jglane:
I would like to propose some major changes to the Lee controller currently implemented:

The current implementation is a simplified version of what is in the actual paper which break the theoretical guarantees which makes this controller useful for researchers. The original 2010 paper [1] (cited in controllerLee.c) assumes a perfect model and if implemented on a real system, there will always be some steady state error. Currently, there is support for an integral term which reduces this error, but it is simply added to the controller without formal proof. Luckily, there is another 2013 paper [2] which extends the previous work to a full PID case and shows exponential stability in the presence of disturbances in the dynamics. I this pr I have implemented the controller exactly as written in [2] which preserves the theoretical guranatees. Names of constants and variables are consistent with variables in the papers.

The current implementation also calculates the desired attitude rate based on the jerk setpoint, which as far as I know is always set to zero. Furthermore, the derivative of the attitude rate which is also used in the controller is set to zero. This greatly simplifies the controller equation, but at the cost of worse tracking of time-varying attitude commands. In this pr, I have implemented first-order estimations of the first and second derivatives of the desired attitude. The values are quite noisy, so I added an averaging filter to smooth out the data. This change makes the Crazyflie much more twitchy and responsive to sudden attitude changes and as a controller that claims to track aggressive maneuvers, I believe this is important. There is also the option to disable these calculations for more stable flight using a non-aggressive trajectory like hovering. All gains have been tuned to the base Crazyflie 2.1+ without any modifications.

The controller lacks support for various flight modes. Currently, there is full support for manual flight and position setpoints, but no explicit calculations for other modes based on the theory. Other than the well-known position controller from [1], an altitude hold mode and velocity control mode are detailed, which I implemented in this pr. There is a comment for each mode with a reference to the appropriate section where the controller can be found. Manual mode and position setpoint mode have been tested with the phone app, cfclient, crazyswarm2, and cflib.

I know there is an excess of controllers in the firmware already, but I believe these changes could help researchers who want to claim some guarantees based on the low-level controller. I am open to any feedback and suggestions, thanks!

[1] Taeyoung Lee, Melvin Leok, and N. Harris McClamroch
Control of Complex Maneuvers for a Quadrotor UAV using Geometric Methods on SE(3)
CDC 2010, updated on arXiv 2011
https://arxiv.org/pdf/1003.2005

[2] Farhad Goodarzi, Daewon Lee, Taeyoung Lee
Geometric Nonlinear PID Control of a Quadrotor UAV on SE(3)
ECC 2013
https://ieeexplore.ieee.org/stamp/stamp.jsp?arnumber=6669644

### DIFF
--- a/src/modules/src/controller/controller_lee.c
+++ b/src/modules/src/controller/controller_lee.c
@@ -186,6 +186,9 @@ void controllerLee(controllerLee_t* self, control_t *control, const setpoint_t *
   if (setpoint->mode.x == modeAbs && setpoint->mode.y == modeAbs && setpoint->mode.z == modeAbs) {
     struct vec x_d = mkvec(setpoint->position.x, setpoint->position.y, setpoint->position.z);
     struct vec v_d = mkvec(setpoint->velocity.x, setpoint->velocity.y, setpoint->velocity.z);
+    if (setpoint->velocity_body) {
+      v_d = mvmul(R, v_d); // Convert body frame velocity to world frame
+    }
     struct vec a_d = mkvec(setpoint->acceleration.x, setpoint->acceleration.y, setpoint->acceleration.z);
     struct vec b1_d = mkvec(cosf(desiredYaw), sinf(desiredYaw), 0);
 
@@ -211,8 +214,11 @@ void controllerLee(controllerLee_t* self, control_t *control, const setpoint_t *
     R_d = mcolumns(vcross(b2_d, b3_d), b2_d, b3_d);
 
   // Velocity setpoint ([1] Sec. VI)
-  } else if (setpoint->mode.x == modeVelocity && setpoint->mode.y == modeVelocity && setpoint->mode.z == modeVelocity) {
+  } else if (setpoint->mode.x == modeVelocity && setpoint->mode.y == modeVelocity) {
     struct vec v_d = mkvec(setpoint->velocity.x, setpoint->velocity.y, setpoint->velocity.z);
+    if (setpoint->velocity_body) {
+      v_d = mvmul(R, v_d); // Convert body frame velocity to world frame
+    }
     struct vec a_d = mkvec(setpoint->acceleration.x, setpoint->acceleration.y, setpoint->acceleration.z);
     struct vec b1_d = mkvec(cosf(desiredYaw), sinf(desiredYaw), 0);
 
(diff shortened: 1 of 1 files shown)


jglane:
Hi @ArisMorgens thank you so much for testing my code and videos! I have addressed your concerns below:

I should have mentioned all my tests with offboard positioning were with Vicon motion capture with an update rate of 150 Hz. So quite an ideal setup. Unfortunately I do not have any additional positioning methods to test with.

I added a uint8 parameter to enable/disable attitude rate tracking. Set ctrlLee.track_att_rate to 0 to disable for simple hovering and gentle flights and 1 to enable for aggressive maneuvers. Can you test with the flow deck again with this parameter set to 0? I'm not sure which one to set as default though, 0 makes it much easier to fly out of the box, but you can't claim it to be the "true" Lee controller because the theoretical guarantees are broken.

I realize now I never actually tested the gamepad with attitude rate tracking on, and now I see the same issue of having no yaw control, thanks for bringing this to my attention. The standard gamepad mode uses a yaw rate setpoint, which the controller interprets as a desired yaw position ahead of the current yaw in the direction of the command like a "carrot on a stick". I added an override to the calculations of the desired attitude rate which uses the setpoint values if provided. This fixed the uncontrollable spinning.

Thanks for the feedback and let me know how it goes if you do more testing!

src/modules/interface/controller/controller_lee.h
```diff
@@ -26,34 +26,50 @@ SOFTWARE.
...
-    struct vec i_error_att;
-    // Logging variables
-    struct vec rpy;
-    struct vec rpy_des;
-    struct mat33 R_des;
-    struct vec omega;
-    struct vec omega_r;
-    struct vec u;
+    // Gains
+    float kx;
```
ToveRumar, line 38: I would like the variables to be more spelled out (readable) even when part of mathematical formulas. 
This will be quite hard to follow in the code. 
Is it possible to re-write these to something more in the line of the previous naming?

whoenig:
While the LEE_CONTROLLER_V2 would certainly ease friction with respect to compatibility, I do find this to be a fairly quick/hacky solution that will put users at a disadvantage. From a technical point of view this also doesn't seem to be justified as both "versions" essentially implement the same paper with some slight variation.
My main concern is that I can't tell from the diff what is even different and if the updated version indeed fixes bugs of the old implementation or adds additional terms as outlined in the newer paper. I did also discuss this with @KhWahba, who wrote the "v1" code and he also had trouble understanding what the actual semantic differences are.
Hence, I am proposing that the PR should be split up in smaller pieces, separating refactoring/renaming and semantic changes. I know that this is (much) more work, but at the end all users will benefit because there will be one trustworthy implementation of the Lee controller.

jglane:
@whoenig thanks for the feedback, and I understand it is important to review each change separately. As for the API, is it possible to change it? The main thing that would need to be changed are the parameters for the gains are three floats each which reference a vector, while my new implementation is a single float. Other than that, the old API can be used, while adding some new variables to support the full implementation.

Another thing I don't understand about the API is the gains are both parameters and log variables. It doesn't seem to be necessary to log them if they are only changed when the parameters are set. Would it be ok to remove these?

I can go ahead and start making more minor changes directly to the current implementation without a full rewrite if that would be best.

PR #1500, opened 2025-07-02
Fixes missing restart of lower prio LED sequence at stop

krichardsson:
This PR fixes a bug in the LED sequencer.

LED sequences have priorities and if a high prio sequence is started when a low prio sequence is runnning, the higher prio sequence will run instead. When the high prio sequence reaches the end (or is stopped) the lower prio sequence should continue where it was halted.

The bug was that the lower prio sequence did not continue as expected.


gemenerik (approved):
Reproduced by putting `seq_testPassed` on `SYS_LED` and observed how it affects `seq_alive` (heartbeat). This PR fixes it.

PR #1501, opened 2025-07-02
Updates to the Lee Controller - Part 1

jglane:
This is part 1 in my attempt to improve the Lee controller. These changes correspond to the first paragraph of my larger PR #1494 and address the following:

1. Implemented the true integral errors as seen in [2].
2. Fixed an issue in the torque calculation, now the implementation is exactly like [2] if the desired angular velocity was set to zero.
3. Some refactoring to hopefully make it more readable and easy to follow from the papers.

Both position and manual control were tested after each commit to make sure they still worked and had the intended effect. Next steps are to add approximation calculations as described in #1494 to estimate the desired angular velocity on the fly.

@whoenig @ToveRumar @KhWahba 

[1] Taeyoung Lee, Melvin Leok, and N. Harris McClamroch
Control of Complex Maneuvers for a Quadrotor UAV using Geometric Methods on SE(3)
CDC 2010, updated on arXiv 2011
https://arxiv.org/pdf/1003.2005

[2] Farhad Goodarzi, Daewon Lee, Taeyoung Lee
Geometric Nonlinear PID Control of a Quadrotor UAV on SE(3)
ECC 2013
https://ieeexplore.ieee.org/stamp/stamp.jsp?arnumber=6669644

### DIFF
--- a/src/modules/src/controller/controller_lee.c
+++ b/src/modules/src/controller/controller_lee.c
@@ -208,8 +208,15 @@ void controllerLee(controllerLee_t* self, control_t *control, const setpoint_t *
     mmul(mtranspose(self->R_des), R),
     mmul(mtranspose(R), self->R_des))));
 
-  // Compute desired omega
+  // Compute desired omega (TODO, zero for now)
   struct vec omega_des = vzero();
+
+  if (setpoint->mode.roll == modeVelocity) {
+    omega_des.x = radians(setpoint->attitudeRate.roll);
+  }
+  if (setpoint->mode.pitch == modeVelocity) {
+    omega_des.y = radians(setpoint->attitudeRate.pitch);
+  }
   if (setpoint->mode.yaw == modeVelocity) {
     omega_des.z = radians(setpoint->attitudeRate.yaw);
   }


stefanthorstenson:
HI @jglane , @whoenig , @KhWahba 

This has not been moved forward for a while now. What is the interest in this PR, and the motivation to work on it?

whoenig (commented):
I think there are two many changes in this PR at once. It mixes semantic changes with small structural changes. Examples:

- Removal of LOG variables is fine (those shouldn't have been there)
- Changing of variable names may be fine, but should be separately discussed
- Adding the integral logic of the ECC paper is in principal also fine (one could get the old behavior by setting those new scaling factors to 1.0)
- The changes in F_d are strange and need discussion (before: not scaled; now: scaled by mass?)
- Setting omega_des to zero is very strange. I am aware that the current computation in the firmware is not correct (for yaw != 0). But the proper way would be to fix that (we have a fix in https://github.com/IMRCLab/crazyflie-firmware/tree/lee_indi2, but didn't take the time to create a clean PR)

Overall, I personally think this PR should be split up even more, so that discussions are easier. Whenever there is a semantic change, flight tests should show that this is an improvement in terms of tracking error, for example.

PR #1526, opened 2025-09-16
Add battery compensation and thrust curves for brushless

ratheron:
Since the thrust of the brushless motors is also dependent on the battery voltage, I implemented the same battery compensation with the thrust curves as in #1465. The verification plot looks as follows:

CF21BL_thrust-verification_B500_2D
CF21BL_thrust-verification_B500_3D

However, there is a problem regarding the maximum thrust. The above curves were collected with an unofficial 500mAh battery. However, paired with the official 350mAh batteries, the (consistently) maximum reachable thrust is way lower, as can be seen in the plot below (the limit indicates a 18g thrust limit per motor, which can be reached even when the battery is low). The 350mAh battery just cant provide enough power.  Currently, the value for the max thrust is identical to the one from the thrust-upgrade-kit (#1521) and way below the advertised possible 40g payload. Even the values with the larger battery can only reach 4*18g=72g of thrust, which results in 38g of excess thrust (assuming 34g weight, as given on the product website), so the payload should not exceed ~33g. In our setup, we fly with guards and the larger battery. The drone weighs 43g in that case, so we only have 29g of additional thrust and a thrust-to-weight ratio of only 1.67.

CF21BL_thrust-curve_B500

Up for discussion is now the following: Should we proceed with the 500mAh battery thrust curves such that the max thrust is not limited that much, but the 350mAh batteries cannot reach, or rather find a suitable value for the 350mAh batteries, which should be around the current 65g of thrust? Just want to remind that the brushless with guards weighs 39g according to the code.

P.S. The other files and the drag coefficients are still worked on


tobbeanton:
Nice work! It does seem low and I'm thinking we (or you) might have an error in the test rig setup. I will try to run some more tests here to see what I get.

tobbeanton:
It turns out it is harder to measure then I thought since the system-id deck itself creates a voltage drop. My fist quick test with the 350 battery indicates 89g trust at a PWM of 58985 and a vBat measurement of 3.425V. I had to bypass the ASC37800 power measurement chip, since this and the extra two connectors otherwise cause a voltage drop of ~0.4V. Does this somewhat match your values?

Rather1337:
Hi Alex, have you been running individual motors, or all at once? With individual motors, I can also reach more thrust (even sligthly more than 0.25N), but I would limit it in a way such that all motors can achieve that thrust simultaneously.

AlexGraefe:
Yes, only with one motor.

tobbeanton:
In the picture there is another battery. Is this the 500mAh one you are using? Do you have the brushless stock 350mAh battery to test with? The thing is that this connector is gold-plated, reducing the contact resistance, which might improve the situation. On the other hand a 500mAh battery might be able to deliver higher currents with less voltage drop, depending on the C rating. However the above should not really affect the seen motor voltage (vBat).

Another thing I can think of is the propeller guards, perhaps they create that much drag, even though I have a hard time beveling that.

> What exactly do you mean by that? What would you increase?

Maybe I understood it wrong but isn't the "0.8N" number setting the maximum thrust limit? According to my measurement I get e.g. @3.2V vBat ->  65535/53000*73g = 90g -> 0.88N. That is if the flight is to be considered "done" with a voltage of 3.2V.

AlexGraefe:
@Rather1337 My setup broke, so I had to use a different scale with lower resolution, which made my measurements a bit noisy. Here are my results:
Figure_1

Our thrust values at 3.5V, 3V, etc., are very close. Since I'm only using one motor (as you suspected), I can get a higher motor voltage and a thrust of up to 0.3N. I measured a voltage drop of about 0.12V from no PWM to full PWM due to internal resistance. Scaling this up, using all four motors would cause a voltage drop of around 0.5V. This matches your maximum motor voltage of 3.5V, assuming your battery started at around 4V.

Rather1337:
Good idea! The cutoff, i.e. THRUST_MAX, is already a platform constant and can easily be changed before flashing. Since the curve is the same, that should work great. Also, in our case, this also works well, since we will most likely change the mass anyway, since we have a heavier deck and the larger battery. 

Then I will set **0.8N** as the default value and write an extensive comment on what a change of that would imply. Maybe I should first write a blog article that I can reference in the code? Also, I would add the torque coefficient, since my results differ from the ones in #1521. With the gathered insight, I will also change the torque coefficients and cutoff points for all the other platforms available to me, i.e. legacy, plus, and thrust upgrade. (More thrust, yay :D)

One question regarding the torque stuff: The sensor is quite expensive and not as straight forward to use, but I would still add it to the readme and keep it in the code?

src/drivers/src/motors.c
```diff
@@ -149,41 +149,38 @@ float motorsCompensateBatteryVoltage(uint32_t id, float iThrust, float supplyVol
...
   {
-    /*
-    * A LiPo battery is supposed to be 4.2V charged, 3.7V mid-charge and 3V
-    * discharged.
-    *
-    * A suitable sanity check for disabling the voltage compensation would be
-    * under 2V. That would suggest a damaged battery. This protects against
-    * rushing the motors on bugs and invalid voltage levels.
-    */
-    if (supplyVoltage < 2.0f)
```
tobbeanton, line 162: Is the sanity check not needed any more?

ratheron: It is still in. Git just displays it weird, since I activated the battery compensation for all platforms.

src/platform/interface/platform_defaults_cf21bl.h
```diff
@@ -47,10 +47,19 @@
...
-#define THRUST_MIN      0.03f // TODO, value is for the thrust upgrade kit
-#define THRUST_MAX      0.1625f // TODO, value is for the thrust upgrade kit
-#define THRUST2TORQUE   0.004899994f
+// Minimum and maximal thrust per motor
+// Note: The maximum thrust is a trade-off between consistency of thrust over all battery levels
+// and maximum performance with a full battery. Increase this value at your own risk. More info
+// in #1526 or this blog post: 
+// https://www.bitcraze.io/2025/10/keeping-thrust-consistent-as-the-battery-drains/
+#define THRUST_MIN      0.02136263065537499f  // N
+#define THRUST_MAX      0.2f  // N
```
ArisMorgens: It would be nice to add a comment on what does` THRUST_MAX` describe. Is it max thrust per motor?

Also, we would like it to make it configurable in **menuconfig**, so the user can modify it in compile-time instead of hardcoding it. There, we can mention the trade-off and propose some values with **0.2 N** being the default one. What do you think the range should be?

ratheron: Added a comment

Regarding the max possible value: Looking at the data from above, the abs max should maybe be 0.25N. The lower bound could be something like 0.15N for super good performance over the whole battery range even with damaged batteries.

ArisMorgens:
Hello @Rather1337!

As you can see we've started reviewing your PR. I think it looks great! I have some comments mostly on when and how to enable the battery compensation feature.

1. The battery compensation should be configurable in compile time, so the user can enable it or not in **menuconfig**. It should be enabled by default. Later, we'll have a discussion on weather it should be configurable also in run-time by having a parameter that can be modified in the **cfclient**.
2. The `THRUST_MAX` should be configurable as well with a description on the trade-off between consistency of thrust over all battery levels and maximum performance with a full battery.
3. There should be documentation on how and why to enable the battery compensation feature for all the different platforms. I don't think the documentation will be affected by the changes in the `thrust2torque` and `max thrust` values for the Crazyflie 2.X platforms.

Since these changes are mostly on the user's side, I think we can take over this PR. If that's ok with you, I can start pushing some commits that fix the above comments and make this feature more compatible with our current ecosystem.

src/platform/interface/platform_defaults_cf21bl.h
ArisMorgens: Great! Thanks for the insight.

src/platform/interface/platform_defaults_cf21bl.h
```diff
@@ -47,10 +47,19 @@
...
-// Thrust coefficients
-#define THRUST_MIN      0.03f // TODO, value is for the thrust upgrade kit
-#define THRUST_MAX      0.1625f // TODO, value is for the thrust upgrade kit
-#define THRUST2TORQUE   0.004899994f
+// Minimum and maximum thrust per motor
+// Note: The maximum thrust is a trade-off between consistency of thrust over all battery levels
+// and maximum performance with a full battery. Increase this value at your own risk. More info
+// in #1526 or this blog post: 
+// https://www.bitcraze.io/2025/10/keeping-thrust-consistent-as-the-battery-drains/
+#define THRUST_MIN      0.02136263065537499f  // N (per motor)
```
ArisMorgens: Do [these values](https://github.com/bitcraze/crazyflie-firmware/pull/1526/files#diff-8a7e5df879426c2dd40fa2fcff4ef9400f9b0f57c245bf803993144f355ba13bR56-R62) result from calculations that take into account the mass of the platform? If so, would it be feasible to include the complete functions? 
The idea behind it is for the user to get a thrust compensation based on their specific setup. Note that the mass configuration will happen after selecting their "original" platform (CF2.1+, ThrustUpgrade, CFBL etc.).

Rather1337: Nope, those are propeller/motor specific values and independent from the rest.

ratheron: No, the thrust curves are only motor/propeller dependent, so the mass is not used for the calculation there.

ArisMorgens: Ok, good to know!

ArisMorgens:
Hello @Rather1337!

I have no further questions. As a final testing step, we used your battery compensation feature on our [decentralized brushless swarm demo](https://www.bitcraze.io/2025/03/decentralized-brushless-swarm-demo/) and it worked really well, meaning that we couldn't observe any difference with the default settings. So, I'll go ahead and merge this PR.

Regarding the bracket files, it would be a nice addition for anyone that wants to do similar experiments with you. Feel free to push the files and open PRs for the links in the documentation.

Thank you again for the contribution!

ArisMorgens (approved):

Issue #1508, opened 2025-07-07
Integration and design questions around new CF2 defconfigs (cf2p, cf2t)


gemenerik:
#1465 added two new defconfigs: `cf2p_defconfig` (for upgraded propellers) and `cf2t_defconfig` (for the thrust upgrade kit).

At the moment, they don’t seem to be fully integrated - they aren’t documented, included in build_targets.json, or part of CI or automated testing.

Beyond that, a few broader questions came up while looking at how these are being used:

* Both defconfigs still set `CONFIG_PLATFORM_CF2=y` and only differ from the base CF2 by the battery compensation setting.
That raises the question: is this kind of variation best handled through separate defconfigs, or should it be exposed via Kconfig?
* There’s also some blurring between platform identity and configuration. Even though these are clearly still CF2 in software, treating them as separate defconfigs invites interpretation that they represent distinct platforms.
* More generally, it’s not always clear where a particular setting is configured between Kconfig, defconfigs, and platform defaults, it can take effort to track down what’s actually in effect.

Opening this up for discussion: what’s the intended role of these defconfigs, and how should we approach this kind of variation going forward?

ataffanel:
Result of triage discussion, we should:
- [ ] Delete the defconfig
- [ ] Document how to compile the firmware using KConfig to enable the right battery compensation
- [ ] Open a new Client/Firmware issue about allowing to configure the hardware setup of a Crazyflie.

gemenerik:
I propose two core principles for defconfigs:
  1. Every defconfig must represent a distinct platform - not just a configuration variant or use case
  2. Every defconfig must correspond to a precompiled binary that users can flash via cfclient

  These principles create clear accountability and prevent architectural confusion. If we're not willing to maintain and release a platform binary, or if it's not actually a separate platform, it shouldn't be a defconfig.

The recently added `cf2p_defconfig` and `cf2t_defconfig` violate both principles. They exist in the configs directory but aren't built in CI or released as binaries, creating a disconnect between what defconfigs imply and what they deliver. More fundamentally, they're not separate platforms - we've never considered the CF2 with thrust upgrade kit or upgraded propellers as distinct platforms (not internally, not in documentation, and not in store listings). Even their defconfig files acknowledge this with CONFIG_PLATFORM_CF2=y. 

[...]

PR #1527, opened 2025-09-16
Clean up defconfig architecture

gemenerik:
Remove cf2p/cf2t defconfigs (should be runtime config) and convert sysid/wearable to platform-agnostic .conf overlays.


ArisMorgens (approved):
This seems like a good step forward, as you explained in #1508. Nice work!

gemenerik:
> This seems like a good step forward, as you explained in #1508. Nice work!

Oh yeah, thanks for tagging that one. Good for future reference.

PR #1534, opened 2025-10-01
Fix USD deck graceful shutdown file corruption

gemenerik:
The shutdown callback was using a mutex (starts available) instead of a binary semaphore (starts empty), causing it to return immediately without waiting for the write task to finish flushing data and closing the file. This resulted in file corruption when powering off during active logging. 

Changes: 
- Use binary semaphore for shutdown synchronization
- Set enableLogging=false to exit logging loop on shutdown 
- Increase timeout from 15ms to 100ms for safety

Thanks to @krombel for reporting and investigating this issue in #1532


tobbeanton (approved):
Nice!

PR #1535, opened 2025-10-03
Updates in arming and idle thrust

ArisMorgens:
This PR adds `MOTORS_DEFAULT_IDLE_THRUST`  in Kconfig, as specified [here](https://github.com/bitcraze/crazyflie-firmware/blob/46a90870d4cbe22f6e397f4f1ba8a1d9ff2e02e9/docs/functional-areas/supervisor/arming.md?plain=1#L39C1-L39C109). 
- It is enabled if `MOTORS_REQUIRE_ARMING` is also enabled. 
- Its default value is `7000`, so it's visible to the user when the platform is ready to fly.
- Instead of defining it for each platform separately, it is now configurable in menuconfig.

This PR also makes arming required if the BigQuad deck is enabled. This check is moved in `platform_defaults.h` instead of `platform_defaults_cf2.h` just in case someone is using the BigQuad with a platform other than the Crazyflie 2.x.


src/modules/src/Kconfig
```diff
@@ -165,6 +165,15 @@ config MOTORS_REQUIRE_ARMING
         When enabled, system must be armed to be able to take off. Arming can
         be done in several ways, e.g. though cfclient or external transmitter.
 
+config MOTORS_DEFAULT_IDLE_THRUST
+    int "Idle thrust for the armed motors"
+    depends on MOTORS_REQUIRE_ARMING
+    range 0 65535
```
tobbeanton: Maybe we should limit the range to say 15000 as more then this should "never" be needed?

tobbeanton (commented):
Nice improvement and perhaps adding the little additional safety.

src/modules/src/Kconfig
ArisMorgens: Good point. Fixed in [this commit](https://github.com/bitcraze/crazyflie-firmware/pull/1535/commits/dfece58308c78bd6fd28ce3398679021351f6d56).

PR #1537, opened 2025-10-06
Updated the app layer documentation to include macOS GNU error fix

enyanil:
Updated the app layer documentation to include macOS GNU error fix when builing the app according to [discussion #1751](https://github.com/orgs/bitcraze/discussions/1751).


ArisMorgens (approved):
Looks good!

Issue #1542, opened 2025-10-13
Implementation of `powerDistributionForce` for individual motor control during flight


stmio:
We are trying to implement a way to control and address the four motors on the Crazyflie 2.1 Brushless separately, for use with our Reinforcement Learning model. We found this section of code which seemed relevant but is currently unimplemented:

https://github.com/bitcraze/crazyflie-firmware/blob/6532d3876741c3e14fb3fd2626d506d06bb1d147/src/modules/src/power_distribution_quadrotor.c#L119-L121

We have since implemented this functionality in our copy of the firmware, would you be happy to accept this as a PR? What (if any) were the original blockers for this (i.e. safety concerns)?

Thank you! :)

cc @finlaysanders

stmio:
Fixed by #1543

PR #1543, opened 2025-10-15
Implements powerDistributionForce for force control mode

FinlaySanders:
See #1542.


FinlaySanders:
Should be good now @tobbeanton :). Let us know if there's anything else we can improve.

Issue #981, opened 2022-03-07
Make initial yaw variance in Kalman filter configurable


ntamas:
It looks like the Kalman core assigns a [fairly low variance](https://github.com/bitcraze/crazyflie-firmware/blob/master/src/modules/src/kalman_core/kalman_core.c#L132) to the initial yaw. In our experience it can be a cause of unexpected flips after takeoff when using the Loco positioning system and the drone is not placed facing the X axis very accurately. We raised this to 0.2 (i.e. 11.5 degrees) in our fork to reduce the chance of flips, and it did not seem to have a negative effect on overall flight performance so I was wondering whether it would be possible to make this option configurable, either with a persistent parameter or a firmware configuration switch.

knmcguire:
Ah yes great find! And yes I would agree that this should be configurable and the variance should probably be higher. To take it further, probably all those std's on those page should be configurable...

The thing is that the gyro seems to be integrated in the covariance update of the EKF, instead of added as a measurement with a standard deviation like in regular kalman filters... so I'm not sure what the full effect would be. 

Probably it would be best to just go ahead and testt this for three cases:
- EKF with manual flight
- EKF with Loco or flow deck (no extra yaw measurements)
- EKF with lighthouse or mocap (with extra yaw measurement)

knmcguire:
It is related to this ticket I opened, to somewhere document the policy #948 

But to come back to this discussion, this is an initial standard deviation that shouldn't change during runtime. I would make a strong case that this should be a define, and configurable in kbuild.

jpreiss:
Tangentially, it would be nice if controllers could be aware of the current EKF belief variance for yaw and use it to modify their gain. Controllers should not be trying to correct yaw error with high gain unless the variance is small.

gemenerik:
I believe the observation in the original post makes logical sense. I am working on yaw estimation in the Kalman filter for the Lighthouse system and have noticed that the default initial standard deviation causes very slow convergence to any initial yaw other than zero. While the Loco Positioning System (LPS) differs from Lighthouse in that yaw is not directly observable, movement in the x-y plane makes that orientation observable again for the Kalman filter (similar for other orientation axes). Colleagues have described damped spiraling behavior during which the Kalman filter gradually corrects the yaw estimate error, demonstrating this. Increasing the initial standard deviation for yaw should allow the filter to correct itself more quickly.

[...]

PR #1546, opened 2025-10-27
Configurable Initial Yaw STD

ArisMorgens:
The initial yaw std describes how uncertain the Kalman estimator assumes the yaw angle is at startup. Its optimal value varies from case to case. For example, with the Lighthouse positioning system, you want the estimator to be certain about the initial yaw angle, while with the Loco positioning system, you want the opposite.

This PR makes the initial yaw std in the Kalman filter configurable in Kbuild.

This solves #981 for now.

### DIFF
--- a/src/modules/interface/kalman_core/kalman_core_params_defaults.h
+++ b/src/modules/interface/kalman_core/kalman_core_params_defaults.h
@@ -46,6 +46,12 @@
   .procNoiseAcc_z = 1.0f
 #endif
 
+#ifdef CONFIG_STIMATOR_KALMAN_INITIAL_YAW_STD
+#define KALMAN_INITIAL_YAW_STD (CONFIG_STIMATOR_KALMAN_INITIAL_YAW_STD / 1000.0f)
+#else
+#define KALMAN_INITIAL_YAW_STD 0.01f
+#endif
+
 /**
  * @brief Macro containing default values for kalmanCoreParams_t
  *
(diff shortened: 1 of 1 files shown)


src/modules/src/Kconfig
```diff
@@ -76,6 +76,14 @@ config ESTIMATOR_KALMAN_GENERAL_PURPOSE
         use any drone specific information for state estimation, this is useful when using the Crazyflie as a general
         positioning tag (tracker).
 
+config ESTIMATOR_KALMAN_INITIAL_YAW_STD_X1000
```
gemenerik: How did we handle the naming of other "should-be-float" config options? Perhaps it is not necessary to put x1000 in the name, and description is enough. Not sure though.

gemenerik: IIRC CF_MASS handled this in a similar way

ArisMorgens: For `CF_MASS` it feels ok to say that the inserted value is in **milligrams (mg)**, while for the initial yaw std it might be a little awkward to say **milliradians (mrad)**. But you might be right. Now that I'm looking at it, `X1000` in the end looks ugly.

ArisMorgens: Fixed in [f95bab9](https://github.com/bitcraze/crazyflie-firmware/pull/1546/commits/f95bab91f8023fce13de586ef95d385380f21acc).

Issue #1540, opened 2025-10-08
Failed attempt to rename spec `link_gcc_c_sequence` to `nosys_link_gcc_c_sequence`


khoi-nguyen-2359:
Hello all.

I run into this error when compiling the source code. In short, the `nosys_link_gcc_c_sequence` somehow already existed when executing `nosys.specs`.

Error logs:

```zsh
➜  crazyflie-firmware git:(develop) ✗ make cf2_defconfig
...
➜  crazyflie-firmware git:(develop) ✗ make -j$(sysctl -n hw.ncpu)
  GEN     ./Makefile
scripts/kconfig/conf  --silentoldconfig Kconfig
  Using .. as source for kernel
  GEN     ./Makefile
  LD      app_api/built-in.o
  CC      src/init/main.o
 ...
  LD      src/init/built-in.o
arm-none-eabi-gcc: fatal error: /Applications/ArmGNUToolchain/14.3.rel1/arm-none-eabi/bin/../lib/gcc/arm-none-eabi/14.3.1/../../../../arm-none-eabi/lib/nosys.specs: attempt to rename spec 'link_gcc_c_sequence' to already defined spec 'nosys_link_gcc_c_sequence'
compilation terminated.
make[3]: *** [src/init/built-in.o] Error 1
make[2]: *** [src/init] Error 2
make[2]: *** Waiting for unfinished jobs....
  CC      src/drivers/esp32/src/esp_rom_bootloader.o
...
```

Environment: MacOS / Apple Silicon
ARM toolchain: 14.3.rel1 (Same issues with 12.2 and 13.3)

Workaround: Removing two-- specs in LDFLAGS (issues with `nano.specs` too) can get the compile to succeed.
```diff
-LDFLAGS += --specs=nosys.specs --specs=nano.specs $(PROCESSOR) -nostdlib
+LDFLAGS += $(PROCESSOR) -nostdlib
```
My question is, when is the `nosys_link_gcc_c_sequence` created the first time? `LDFLAGS` in `Makefile` seems to be the only place showing that the specs are specified. However, I notice that `nosys` and `nano` specs are also used in `vendor/CMSIS/CMSIS/DSP/Toolchain/GCC.cmake`.

Thank you!

PR #1557, opened 2025-11-12
Add Pixi manifest

EliaCereda:
Pixi (pixi.sh, https://arxiv.org/abs/2511.04827) is a next-gen package manager for reproducible development setups. It provides a streamlined user experience, self-contained environments and a single set of dependencies that are pinned across all platforms.

This PR provides an example `pixi.toml` manifest file to build crazyflie-firmware with Pixi. After installing Pixi (https://pixi.sh/latest/installation/), the firmware can be built and flashed with the following commands:
```
$ pixi shell
> make cf21bl_defconfig
> make all -j
> make cload
```

Tested on macOS arm64 and Linux amd64

On Linux, the Makefile required a small change due to the linker error from #1540:
```
arm-none-eabi-gcc: fatal error: /code/.pixi/envs/default/arm-none-eabi/bin/../lib/gcc/arm-none-eabi/13.3.1/../../../../arm-none-eabi/lib/nosys.specs: attempt to rename spec 'link_gcc_c_sequence' to already defined spec 'nosys_link_gcc_c_sequence'
```

Which was caused by duplicated flags in the linker invocation:
```
arm-none-eabi-gcc ... --specs=nosys.specs --specs=nano.specs -mcpu=cortex-m4 -mthumb -mfloat-abi=hard -mfpu=fpv4-sp-d16 -nostdlib --specs=nosys.specs --specs=nano.specs -mcpu=cortex-m4 -mthumb -mfloat-abi=hard -mfpu=fpv4-sp-d16 -nostdlib ...
```

This is most likely not specific to Pixi, but caused by recent versions of Make and/or the embedded GCC toolchain

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -4,7 +4,7 @@
 # Run `pixi shell` to enter a shell with the dependencies available, or `pixi run <command>` to run a command with the dependencies available.
 
 [workspace]
-authors = ["Elia Cereda <eliacereda@gmail.com>"]
+authors = ["Elia Cereda <eliacereda@gmail.com>", "Arnaud Taffanel <arnaud@bitcraze.io>"]
 channels = ["conda-forge", "Memfault"]
 name = "crazyflie-firmware"
 platforms = ["linux-64", "osx-64", "osx-arm64"]


EliaCereda:
Sounds good to me. It would look strange to supply a `pyproject.toml` for a repo that doesn't build a Python package, and it can be added if this changes in the future.

In the future, it will be nice to also explore `pixi build` (https://pixi.sh/latest/build/getting_started/) to build an entire workspace of CF projects (eg cf-firmware, cfclient, etc) in a single command. But from what I can see, at the moment it might not fit well with cross-compiling an embedded firmware and flashing it. It seems to only build packages that are installed in the Pixi workspace.

gemenerik (approved):

PR #1558, opened 2025-11-12
I2C DFU access to Color deck MCU

evoggy:
The pull-request contains the following items:

* Generic I2C DFU STM32 read/write implementation (erase page on write on or across page boundaries)
* Access to Color deck MCU flash memory though DFU
* Functionality for switching between firmware and bootloader in the Color MCU
* Moves start of VL53L0x remap address moved (still support for 5 + sensors)
* A new deck memory bit for indicating if a deck can be upgraded without power-cycling (HOT_RESTART)

### DIFF
--- a/src/deck/drivers/src/color_led_deck.c
+++ b/src/deck/drivers/src/color_led_deck.c
@@ -333,13 +333,13 @@ static uint8_t colorFlasherPropertiesQuery() {
 static void resetColorDeckToBootloader() {
   isInFirmware = false;
 
-  deckctrl_gpio_write(deck_info, GPIO_PWR_EN, false);
-  deckctrl_gpio_set_direction(deck_info, GPIO_DFU_EN, true);
-  deckctrl_gpio_write(deck_info, GPIO_DFU_EN, true);
+  deckctrl_gpio_write(deck_info, GPIO_PWR_EN, LOW);
+  deckctrl_gpio_set_direction(deck_info, GPIO_DFU_EN, OUTPUT);
+  deckctrl_gpio_write(deck_info, GPIO_DFU_EN, HIGH);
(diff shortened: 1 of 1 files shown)


src/deck/drivers/src/color_led_deck.c
```diff
@@ -226,77 +242,136 @@ static void task(void *param) {
...
+    result |= DECK_MEMORY_MASK_BOOT_LOADER_ACTIVE;
+  }
+
+  return result;
+}
+
+static void resetColorDeckToBootloader() {
+  isInFirmware = false;
+
+  deckctrl_gpio_write(deck_info, GPIO_PWR_EN, false);
```
gemenerik: We have HIGH and LOW macros for readability

src/deck/drivers/src/color_led_deck.c
```diff
@@ -226,77 +242,136 @@ static void task(void *param) {
...
+  deckctrl_gpio_write(deck_info, GPIO_DFU_EN, true);
+  vTaskDelay(M2T(10));
+  deckctrl_gpio_write(deck_info, GPIO_PWR_EN, true);
+  vTaskDelay(M2T(10));
+  deckctrl_gpio_set_direction(deck_info, GPIO_DFU_EN, false);
```
gemenerik: Similarly we have INPUT and OUTPUT macros

src/deck/drivers/src/color_led_deck.c
evoggy: Good catch, updated to those instead.

src/deck/drivers/src/color_led_deck.c
evoggy: Good catch, updated to those instead.

PR #1559, opened 2025-11-13
Update sysid values

ratheron:
As promised in #1526, I also updated the torque coefficients for all the cf2x platforms. In that process I noticed that I also need to update the thrust curve for the brushless, since without the sysid deck, way higher thrust levels could be reached.

### DIFF
--- a/tools/system_id/README.md
+++ b/tools/system_id/README.md
@@ -74,6 +74,18 @@ To verify the parameters, after adding the new values and flashing, we run the d
 
 #### Dynamic parameters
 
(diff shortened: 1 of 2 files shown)


ArisMorgens (approved):

Rather1337:
Glad to hear! We are currently updating the scripts to fit the rotor dynamics and I would like to also add those to this PR, if that's fine with you. We should be done by this week.

ArisMorgens (approved):

ArisMorgens:
Both the torque coefficients and the sysid changes seem good. @Rather1337 Thanks for the contribution!

PR #1560, opened 2025-11-19
CRTP Supervisor

ArisMorgens:
This PR adds a CRTP Supervisor module that allows direct access to the supervisor through the lib.
~~Currently, only one channel (INFO) is added and is only used for reading the states.
In the future, we should discuss how to migrate commands like the **arming request** and the **recovery request** from `platform service` to the `supervisor`.~~

**Update**: 
Two channels are added: **Info** and **Commands**. The **Info** channel is read-only and is used to get the bitfield of the supervisor states, that is later decoded in the library side. The **Commands** channel is used for sending supervisor commands (currently **arming request** and **recovery request**)

### DIFF
--- a/docs/functional-areas/crtp/crtp_localization.md
+++ b/docs/functional-areas/crtp/crtp_localization.md
@@ -49,8 +49,8 @@ streams. The format of the packet is:
 |   0  | Range stream report                        |
 |   1  | Range stream report, 16 bit floating point |
 |   2  | LPP Short packet tunnel                    |
-|   3  | Enable emergency stop                      |
-|   4  | Reset emergency stop timeout               |
+|   3  | Enable emergency stop *(deprecated, use [supervisor port](crtp_supervisor.md#emergency-stop))* |
+|   4  | Reset emergency stop timeout *(deprecated, use [supervisor port](crtp_supervisor.md#emergency-stop-watchdog))* |
 |   6  | COMM GNSS NMEA                             |
 |   7  | COMM GNSS proprietary                      |
 |   8  | External pose information                  |
--- a/docs/functional-areas/crtp/crtp_platform.md
+++ b/docs/functional-areas/crtp/crtp_platform.md
@@ -21,7 +21,8 @@ The first byte describes the command:
 | value | Command |
 |-------|---------|
 | 0     | [Set continuous wave](#set-continuous-wave) |
-| 1     | [Request arm/disarm the system](#armdisarm-system) |
+| 1     | Request arm/disarm the system *(deprecated, use [supervisor port](crtp_supervisor.md#armdisarm-system))* |
+| 2     | Recover system *(deprecated, use [supervisor port](crtp_supervisor.md#recover-system))* |
 
 ### Set continuous wave
 
@@ -38,25 +39,6 @@ freqency. The same packet is sent back to confirm the value has been set.
 This command should only be sent over USB (it disables the radio communication).
 It is used in production to test the Crazyflie radio path and should not be used outside of a lab or
 other very controlled environment. It will effectively jam local radio communication on the channel.
-
-### Arm/disarm system
-
-Arm or disarm the system if possible.
-
-Command:
-
-| Byte | Description                           |
-|------|---------------------------------------|
-| 0    | command request arm/disarm system (1) |
-| 1    | 0 = disarm, non-zero = arm the system |
-
-Answer:
-
-| Byte | Description                                          |
-|------|------------------------------------------------------|
-| 0    | command request arm/disarm system (1)                |
-| 1    | success: 1 if the requested arming state was set     |
-| 2    | isArmed: 0 = system is disarmed, 1 = system is armed |
 
 ## Version commands
 
--- a/docs/functional-areas/crtp/crtp_supervisor.md
+++ b/docs/functional-areas/crtp/crtp_supervisor.md
@@ -1,0 +1,127 @@
+---
+title: Supervisor CRTP port
+page_id: crtp_supervisor
+---
+
+This port exposes supervisor commands and state queries.
+
+## CRTP channels
+
+| Port | Channel | Function |
+|------|---------|----------|
+| 9    | 0       | [State info](#state-info) |
+| 9    | 1       | [Commands](#commands) |
+
+Packets on channel 1 (commands) are handled in a high-priority CRTP callback. Packets on channel 0 (info) are handled in a lower-priority task.
+
+For responses, byte 0 is the command ID with the high bit set (`cmd | 0x80`).
+
+## Commands
+
+The first byte of the payload is the command ID.
+
+| Value | Command |
+|-------|---------|
+| 0x01  | [Arm/disarm system](#armdisarm-system) |
+| 0x02  | [Recover system](#recover-system) |
+| 0x03  | [Emergency stop](#emergency-stop) |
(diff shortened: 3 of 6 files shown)


src/modules/interface/crtp_supervisor.h
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#ifndef _CRTP_SUPERVISOR_H_
+#define _CRTP_SUPERVISOR_H_
+
+#pragma once
```
gemenerik: The header guards (at lines 26, 27, closed at 53) and #pragma once serve the same purpose, you need only one or the other.

src/modules/src/crtp_supervisor.c
```diff
@@ -0,0 +1,118 @@
...
+
+static void crtpSupervisorCB(CRTPPacket* p);
+static void handleSupervisorInfo(CRTPPacket* p);
+static void sendSupervisorResponse(uint8_t cmd, const void* data, uint8_t len);
+
+
+void crtpSupervisorInit(void)
+{
+    // Register the callback for the supervisor port
+    crtpRegisterPortCB(CRTP_PORT_SUPERVISOR, crtpSupervisorCB);
```
gemenerik, line 68: I think we should consider a hybrid approach with both callbacks and queues for the supervisor CRTP handler. Right now the implementation only handles status queries (reading bits and sending replies), which aren't time-critical. These should probably use the queue pattern like param, log, and memory do, they all have dedicated tasks at low priority (1). Callbacks run in the CRTP RX task (priority 2) and block other CRTP packets from being processed, so we should reserve them for time-critical stuff. 

However, the supervisor will also need to handle things like emergency stop, which IS time-critical and needs callback-based processing for fast response. It has not been done yet, but a port can theoretically have both a callback AND a queue ([crtp.c:180-189](https://github.com/bitcraze/crazyflie-firmware/blob/2025.12/src/modules/src/crtp.c#L180-L189)). We can filter by channel:
`SUPERVISOR_CH_INFO`: Status queries -> ignored in callback, processed by queue task
`SUPERVISOR_CH_COMMAND` (or something like that): Time-critical commands like emergency stop -> processed in callback, ignored by queue task.

Something like:
```c
void crtpSupervisorInit(void) {
    crtpRegisterPortCB(CRTP_PORT_SUPERVISOR, crtpSupervisorCB);  // Fast path
    crtpInitTaskQueue(CRTP_PORT_SUPERVISOR);                      // Slow path
    STATIC_MEM_TASK_CREATE(supervisorTask, ...);
}

static void crtpSupervisorCB(CRTPPacket* pk) {
    // Only handle time-critical commands
    if (pk->channel == SUPERVISOR_CH_COMMAND) {
        switch (pk->data[0]) {
            case CMD_EMERGENCY_STOP:
                // Trigger emergency stop
                break;
        }
    }
    // SUPERVISOR_CH_INFO just falls through to queue
}

static void supervisorTask(void* param) {
    CRTPPacket p;
    while(1) {
        crtpReceivePacketBlock(CRTP_PORT_SUPERVISOR, &p);
        
        if (p.channel == SUPERVISOR_CH_INFO) {
            handleSupervisorInfo(&p);  // Current code
        }
    }
}
```
This way we consolidate everything supervisor-related into one port while keeping good performance for both queries and critical commands. @ataffanel - thoughts on this approach?

src/modules/src/crtp_supervisor.c
```diff
@@ -0,0 +1,118 @@
...
+static void handleSupervisorInfo(CRTPPacket* p)
+// It handles the requests to the Info channel.
+{
+    if (p->size < 1) {
+        return;
+    }
+
+    uint8_t cmd = p->data[0];
+    uint8_t value8 = 0;
+    uint32_t bitfield = supervisorMem.infoBitfield;
```
gemenerik: The bitfield you are reading from is `uint16_t`, so you can make this one `uint16_t` as well. Saves us 2 bytes of communication! :partying_face:

src/modules/src/crtp_supervisor.c
```diff
@@ -0,0 +1,118 @@
...
+
+        case CMD_GET_STATE_BITFIELD:
+            sendSupervisorResponse(cmd, &bitfield, sizeof(bitfield));
+            return;
+
+        default:
+            return;
+    }
+
+    sendSupervisorResponse(cmd, &value8, 1);
```
gemenerik: Neat!

src/modules/interface/supervisor.h
```diff
@@ -27,6 +27,38 @@
...
+  uint32_t latestLandingTick;
+
+  supervisorState_t state;
+
+  // Copy of latest conditions, for logging
+  supervisorConditionBits_t latestConditions;
+  uint8_t doinfodump;
+} SupervisorMem_t;
+
+extern SupervisorMem_t supervisorMem;
```
gemenerik: Consider moving `SupervisorMem_t` and `supervisorMem` back to where they were in `supervisor.c` and instead expose a getter function:

```c
// In supervisor.h
uint16_t supervisorGetInfoBitfield(void);

// In supervisor.c
uint16_t supervisorGetInfoBitfield(void) {
  return supervisorMem.infoBitfield;
}
```

Exposing the entire struct means anyone can (accidentally) modify internal supervisor state (like state, isTumbled, etc.) and break the state machine. A getter provides read-only access to just what's needed.

src/modules/src/crtp_supervisor.c
```diff
@@ -0,0 +1,118 @@
...
+
+static bool isInit = false;
+
+static void crtpSupervisorCB(CRTPPacket* p);
+static void handleSupervisorInfo(CRTPPacket* p);
+static void sendSupervisorResponse(uint8_t cmd, const void* data, uint8_t len);
+
+
+void crtpSupervisorInit(void)
+{
```
gemenerik, line 62: add:
```c
  if(isInit) {
    return;
  }
```

src/modules/interface/crtp_supervisor.h
ArisMorgens: Nice catch - fixed in [e38a5f0](https://github.com/bitcraze/crazyflie-firmware/pull/1560/commits/e38a5f01221d03375c9b794db4d4d7151395f23c)

src/modules/src/crtp_supervisor.c
ArisMorgens: That's sick! :partying_face: Fixed in [f0eda96](https://github.com/bitcraze/crazyflie-firmware/pull/1560/commits/f0eda968f19f226f79cea34f46c485d8df5068a0)

src/modules/interface/supervisor.h
ArisMorgens: Thanks for the explanation. Moved it back in [6c8df96](https://github.com/bitcraze/crazyflie-firmware/pull/1560/commits/6c8df967b26c405c259d0688880e537c3278b7aa)

src/modules/src/crtp_supervisor.c
ArisMorgens: Fixed in [e5c352d](https://github.com/bitcraze/crazyflie-firmware/pull/1560/commits/e5c352db447697ede3261b6cb80e273c8398b958)

gemenerik (approved):

PR #1566, opened 2025-11-28
Handle eeprom with no i2c answer and  kvd header error.

tobbeanton:
In the case of a missing, broken or corrupt EEPROM the system got stuck during boot with only the two blue LEDs lit. This PR adds checks to handle this. In case of an error this will be visible in the console output from the Crazyflie and it will go into the error state.

### DIFF
--- a/src/utils/src/kve/kve_storage.c
+++ b/src/utils/src/kve/kve_storage.c
@@ -237,7 +237,9 @@ size_t kveStorageFindNextItem(kveMemory_t *kve, size_t address)
         if (header.full_length == KVE_END_TAG) {
             return KVE_STORAGE_INVALID_ADDRESS;
         }
-
+        if (header.full_length == 0) {
+            return KVE_STORAGE_INVALID_ADDRESS;
+        }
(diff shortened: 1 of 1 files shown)


src/utils/src/kve/kve_storage.c
```diff
@@ -210,15 +218,21 @@ size_t kveStorageFindNextItem(kveMemory_t *kve, size_t address)
...
 
     // Jump over the current item
-    kve->read(currentAddress, &header, sizeof(header));
+    if (kve->read(currentAddress, &header, sizeof(header)) == 0) {
+      return KVE_STORAGE_INVALID_ADDRESS;
+    }
     if (header.full_length == KVE_END_TAG) {
         return KVE_STORAGE_INVALID_ADDRESS;
     }
+    if (header.full_length == 0) {
```
gemenerik, line 227: Should this check be repeated at line 240?

tobbeanton: It will not hurt. I'll add that

PR #1567, opened 2025-12-01
Implement flashing of DeckCtrl MCU

evoggy:
This pull request introduces support for managing DFU (Device Firmware Update) memory on DeckCtrl-managed decks. The change adds a new memory handler for DeckCtrl DFU, which enables reading and writing of DFU memory via I2C, and provides mechanisms to switch the DeckCtrl between DFU and firmware modes using nRF51 commands.

**Note** This only works with *one* deck with the DeckCtrl system at a time. If more are attached the system will try to detect this and block switching the MCU into DFU mode.

### DIFF
--- a/src/deck/interface/deckctrl_dfu_memory.h
+++ b/src/deck/interface/deckctrl_dfu_memory.h
@@ -35,7 +35,7 @@
 /**
  * @brief Initialize the DeckCtrl DFU memory interface
  * 
- * This initializes the DeCkCtrl DFU memory interface and registers a
+ * This initializes the DeckCtrl DFU memory interface and registers a
  * memory with the memory subsystem.
  * 
  */


src/deck/core/deckctrl_dfu_memory.c
```diff
@@ -0,0 +1,190 @@
...
+    if (memAddr < DECK_CTRL_MEM_OFFSET) {
+        for (unsigned int i = 0; i < writeLen; i++) {
+            if (memAddr + i == CMD_MEMORY_BYTE) {
+                if (
+                    (buffer[i] & CMD_ENTER_FIRMWARE_MODE) != 0 &&
+                    (buffer[i] & CMD_ENTER_DFU_MODE) != 0)
+                {
+                    return false;
+                } else if ((buffer[i] & CMD_ENTER_FIRMWARE_MODE) != 0) {
+                    bool shouldEnable = false;
```
tobbeanton: Is there a chance this variable will be overwritten on the stack before it is used?

src/deck/core/deckctrl_dfu_memory.c
```diff
@@ -0,0 +1,190 @@
...
+                    workerSchedule(enableDFUViaNRF, (void *)shouldEnable);
+                } else if ((buffer[i] & CMD_ENTER_DFU_MODE) != 0) {
+                    if (nbrOfDeckCtrl() <= 1) {
+                        bool shouldEnable = true;
```
tobbeanton: Same here?

tobbeanton (changes requested):

src/deck/core/deckctrl_dfu_memory.c
evoggy: The value is copied when the workerSchedule function is called since we don't actually send the pointer to the data but the data itself as the pointer.

src/deck/core/deckctrl_dfu_memory.c
evoggy: The value is copied when the workerSchedule function is called since we don't actually send the pointer to the data but the data itself as the pointer.

src/deck/core/deckctrl_dfu_memory.c
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    return status;
+}
+
+// Calling this function will cut the power to VCC including this MCU and the
+// DeckCtrl, causing a reset. The DeckCtrl will then enter DFU mode on next
+// power-up.
+static void enableDFUViaNRF(void * arg) {
+    bool enable = (bool) arg;
```
gemenerik: casting a pointer, is that safe?

src/deck/core/deckctrl_dfu_memory.c
```diff
@@ -0,0 +1,190 @@
...
+ *
+ */
+
+#include <string.h>
+#include "deck_memory.h"
+#include "deck_core.h"
+#include "mem.h"
+#include "i2cdev.h"
+#include "i2c_dfu.h"
+#include "deck_core.h"
```
gemenerik: double include

src/deck/interface/deckctrl_dfu_memory.h
```diff
@@ -0,0 +1,42 @@
...
+
+#pragma once
+
+#include <stdbool.h>
+#include <stdint.h>
+
+/**
+ * @brief Initialize the DeckCtrl DFU memory interface
+ * 
+ * This initializes the DeCkCtrl DFU memory interface and registers a
```
gemenerik:
```suggestion
 * This initializes the DeckCtrl DFU memory interface and registers a
```

evoggy: Fixed!

src/deck/core/deckctrl_dfu_memory.c
evoggy: Fixed

src/deck/core/deckctrl_dfu_memory.c
evoggy: I updated this to be more clear by using a heap allocated variable and passing it as a pointer.

src/deck/core/deckctrl_dfu_memory.c
evoggy: I updated this to be more clear by using a heap allocated variable and passing it as a pointer.

src/deck/core/deckctrl_dfu_memory.c
evoggy: I updated this to be more clear by using a heap allocated variable and passing it as a pointer.

ataffanel (approved):

gemenerik (approved):

tobbeanton (approved):

PR #1568, opened 2025-12-04
Ekf python bindings mocap

NageshEranki:
For my project, I need to dissect the EKF as I noticed some odd behavior. There is documentation on building python bindings to wrap up some C code in the firmware for individual testing. Until now, there was support only for anchors, accel. and gyro measurements. This PR attempts to expand the Kalman filter emulator in the following way:

1. Expose the kalmanCoreUpdateWithPose() function by including the source files in cffirmware.i
2. Fix a critical bug in _update_with_sample() function. Until now, this function discards sample if they dont align with the current tick, this causes certain measurements to never processed (as was my case; mocap was never processed). 
3. Let the user know explicitly (through print statements) that a particular sample was not handled.

I am attaching some example code, data sets for anyone to recreate results (it was quite an aggressive flight as you can see).

### DIFF
--- a/bindings/util/estimator_kalman_emulator.py
+++ b/bindings/util/estimator_kalman_emulator.py
@@ -1,5 +1,4 @@
 import math
-import numpy as np
 import cffirmware
 
 class EstimatorKalmanEmulator:


NageshEranki:
Forgot: the code and log data.


[data.json.txt](https://github.com/user-attachments/files/23943799/data.json.txt)
[analysisTools.py](https://github.com/user-attachments/files/23943812/analysisTools.py)
[EKF_test.ipynb](https://github.com/user-attachments/files/23943813/EKF_test.ipynb)

bindings/util/estimator_kalman_emulator.py
```diff
@@ -80,7 +83,7 @@ def _lazy_init(self, sample_time_ms):
         self.next_prediction_ms = self.now_ms + self.PREDICT_STEP_MS
 
         GRAVITY_MAGNITUDE = 9.81
-        DEG_TO_RAD = math.pi / 180.0
+        DEG_TO_RAD = np.pi / 180.0
```
gemenerik: Any reason not to use `math.pi`? Otherwise let's revert.

gemenerik (changes requested):
Thanks again. It looks good. I just have a comment on the use of numpy. There are also some white spaces on lines 107, 149, 153 and 171 that would be nice to have cleared up.

bindings/util/estimator_kalman_emulator.py
NageshEranki: Thanks, I believe the current version is up to date. Let me know if anything else is required.

PR #1570, opened 2025-12-08
build: remove duplicate collision_avoidance.o entry

SEOSUK:
### Summary
Remove a duplicated `collision_avoidance.o` entry from `src/modules/src/Makefile`.

### Details
- The object file was listed twice in `obj-y`
- No functional or binary changes
- This is a small build-system cleanup

### Impact
- No behavior change
- Improves build cleanliness and readability


ArisMorgens:
Seems good! Thanks again for the contribution!

PR #1572, opened 2025-12-09
Programmatically-readable color led deck self-test results, add I2C address pin test

gemenerik:
Enhanced the production self-test for Color LED decks with programmatically-readable test results and integrating LED position verification inside deck self test:

- Test I2C_ADDRESS pin on Color LED deck connectivity by toggling it LOW/HIGH and verifying the deck MCU reads the correct state via I2C command. Ensures proper signal routing during production testing.
- Added deckTest param group with bitmap test results for both decks (bit 0: protocol version, bit 1: LED position, bit 2: I2C address pin)
- Replaced pollLedPosition() with verifyLedPosition() that checks LED is mounted on correct PCB side (bottom/top) and fails test on mismatch
- Added LED position constants matching deck firmware values (NONE=0x00, BOTTOM=0x01, TOP=0x02)
- Removed LED position from log groups (now verified during self test only)
- Added debug output for test failures

Requires:
* bitcraze/color-led-deck-firmware/pull/4

### DIFF
--- a/src/deck/drivers/src/color_led_deck.c
+++ b/src/deck/drivers/src/color_led_deck.c
@@ -382,19 +382,18 @@ static bool verifyLedPosition(colorLedContext_t *ctx, uint8_t expectedPosition) 
   uint8_t actualPosition = response[1];
   ctx->ledPosition = actualPosition;
 
-  const char* expectedStr = (expectedPosition == COLORLED_LED_POS_BOTTOM) ? "BOTTOM" : "TOP";
-  const char* actualStr;
-  if (actualPosition == COLORLED_LED_POS_BOTTOM) {
-    actualStr = "BOTTOM";
-  } else if (actualPosition == COLORLED_LED_POS_TOP) {
-    actualStr = "TOP";
-  } else if (actualPosition == COLORLED_LED_POS_NONE) {
-    actualStr = "NONE";
-  } else {
-    actualStr = "UNKNOWN";
-  }
-
   if (actualPosition != expectedPosition) {
+    const char* expectedStr = (expectedPosition == COLORLED_LED_POS_BOTTOM) ? "BOTTOM" : "TOP";
+    const char* actualStr;
+    if (actualPosition == COLORLED_LED_POS_BOTTOM) {
+      actualStr = "BOTTOM";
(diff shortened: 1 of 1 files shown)


src/deck/drivers/src/color_led_deck.c
```diff
@@ -337,20 +361,122 @@ static bool pollLedCurrent(colorLedContext_t *ctx) {
...
+
+  if (response[0] != CMD_GET_LED_POSITION) {
+    DEBUG_PRINT("Invalid response command: 0x%02x\n", response[0]);
+    return false;
+  }
+
+  uint8_t actualPosition = response[1];
+  ctx->ledPosition = actualPosition;
+
+  const char* expectedStr = (expectedPosition == COLORLED_LED_POS_BOTTOM) ? "BOTTOM" : "TOP";
```
evoggy: This seems to be only used for debug printing, what about putting it inside an ```#ifdef DEBUG``` and inside the if so it's only used if the comparison fails?

src/deck/drivers/src/color_led_deck.c
```diff
@@ -738,3 +854,25 @@ PARAM_ADD_CORE(PARAM_UINT8 | PARAM_RONLY, bcColorLedBot, &contexts[BOTTOM_IDX].i
 PARAM_ADD_CORE(PARAM_UINT8 | PARAM_RONLY, bcColorLedTop, &contexts[TOP_IDX].isInit)
 
 PARAM_GROUP_STOP(deck)
+
+PARAM_GROUP_START(deckTest)
```
evoggy, line 854: What about turning this around so we turn on the bits which fails? This way we can easily see if decks pass/fail just comparing it with 0 without knowing which bits are what.

src/deck/drivers/src/color_led_deck.c
gemenerik: I think this is important info to print, but we only need to (and do) in case of failure, so I moved constructing the string inside the failure case

src/deck/drivers/src/color_led_deck.c
gemenerik: Done in 9fbcbd49d03b1f13a388b50c25e97039932d928d

PR #1573, opened 2025-12-10
Add production date for DeckCtrl

evoggy:
This change adds the production date for the DeckCtrl memory into the configuration page.

### DIFF
--- a/docs/functional-areas/deckctrl_protocol.md
+++ b/docs/functional-areas/deckctrl_protocol.md
@@ -114,7 +114,7 @@ The 32-byte deck information block contains:
 | 0x17   |    1 | Manufacturing month | Month of manufacture (1-12) |
 | 0x18   |    1 | Manufacturing day | Day of manufacture (1-31) |
 | 0x19   |    6 | Reserved | Reserved for future use |
-| 0x1F   |    1 | Reserved | Value to make the checksum (modulo 256) of bytes 0-31 equal to zero |
+| 0x1F   |    1 | Checksum | Value to make the checksum (modulo 256) of bytes 0-31 equal to zero |
 
 #### Magic Number
 


src/deck/backends/deck_backend_deckctrl.c
```diff
@@ -91,6 +92,10 @@ static const DeckDiscoveryBackend_t deckctrlBackend;
 #define DECKCTRL_MEM_BOARD_REV_OFFSET 6    ///< Board revision ASCII character
 #define DECKCTRL_MEM_PRODUCT_NAME_OFFSET 7 ///< Product name (null-terminated, max 14 chars)
 
+#define DECKCTRL_MEM_PRODUCTION_YEAR_OFFSET 0x16  ///< Production year offset
```
gemenerik, line 95: Is there a gap at 0x15?

src/deck/backends/deck_backend_deckctrl.c
```diff
@@ -209,27 +214,37 @@ static DeckInfo* deckctrl_getNextDeck(void)
...
                 cpu_id[4], cpu_id[5], cpu_id[6], cpu_id[7],
                 cpu_id[8], cpu_id[9], cpu_id[10], cpu_id[11]);
-    DEBUG_PRINT("Assigned new I2C address: 0x%02x\n", deck_address);
+    DECKCTRL_DEBUG_PRINT("Assigned new I2C address: 0x%02x\n", deck_address);
 
     // Read deck information from the newly assigned address at register 0x0000
-    // Format: magic(2) + versions(2) + vendor(1) + product(1) + revision(1) + name(14)
-    uint8_t deck_mem_data[21];
+    // Format: magic(2) + versions(2) + vendor(1) + product(1) + revision(1) + name(14) + production date(3) + checksum(1)
+    uint8_t deck_mem_data[0x20];
```
gemenerik, line 225: Summing the numbers in the comment the length should be 25 (0x19)

src/deck/backends/deck_backend_deckctrl.c
```diff
@@ -209,27 +214,37 @@ static DeckInfo* deckctrl_getNextDeck(void)
...
+        return NULL;
+    }
+
+    // Validate checksum (sum of all bytes modulo 256 should be 0)
+    uint8_t checksum = 0;
+    for (size_t i = 0; i < sizeof(deck_mem_data); i++) {
+        checksum += deck_mem_data[i];
+    }
+    if (checksum != 0) {
+        DEBUG_PRINT("Invalid checksum 0x%02x, expected 0x00\n", checksum);
```
gemenerik, line 247: so if I understand correctly, the config page has a checksum that makes it s.t. summing all bytes in the config page % 256 = 0

evoggy: Correct! I've updated the documentation to reflect this.

src/deck/backends/deck_backend_deckctrl.c
evoggy: The documentation has been updated to show that the field is 15 chars (including the 0-terminator).

src/deck/backends/deck_backend_deckctrl.c
evoggy: Updated the comment to reflect that the config block is actually 32 bytes.

PR #1580, opened 2025-12-15
Color LED effects app

ArisMorgens:
This app contains some effects for the Color LED decks attached to a Crazyflie that were used for the 2025 Christmas video. There's a short documentation for each effect in README,


gemenerik:
I think it is a good idea to share this, but I am not sure if this is the correct place. Anything in examples should be tested and maintained for all releases, and this example having more integration with other subsystems is cool but increases maintenance load a lot. I think it may be time that we open up a crazyflie-demos repository or something like that, where examples like this one are hosted. We could include compatibility info with the examples (this was tested with...) or even include submodules with the latest tested firmware so it is guaranteed to run. That way we can share cooler examples like this one without adding too much maintenance burden.

ArisMorgens:
Creating a crazyflie-demo repository sounds like a viable solution for the continuously increasing amount of examples. Merging this one!

Issue #832, opened 2021-08-17
Mellinger controller: Attitude control unstable beyond about 30 degrees roll/pitch


whoenig:
the SE(3) controller from Mellinger/Li should be capable of trajectory tracking with around 60 degrees roll/pitch angle without problems. Our implementation only can get to around 30 degrees. This has been observed by @whoenig and @jpreiss in a trajectory tracking scenario and by @pPatrickK in https://github.com/USC-ACTLab/crazyswarm/issues/71 in a position setpoint step scenario. This might be due to some implementation bug or because of the low thrust-to-weight ratio of Crazyflies.

Transferred from https://github.com/USC-ACTLab/crazyswarm/issues/74.

jonasdn:
What would be a good way forward for this issue?

How do we determine if this is a bug or if we have reached a limitation of thrust-to-weight ratio?

Any ideas?

whoenig:
A software verification might be possible with the Python bindings (https://github.com/bitcraze/crazyflie-firmware/issues/602) and simulation of a model with a higher t2w ratio. A hardware verification would require upgraded motors.

whoenig:
One problem I just found is the omega error: this should be both in body coordinates (equation right above (8) in the paper; Lee uses R^T R_d, see (8) in his paper). I am not sure if this can cause these kind of errors, but it can certainly explain odd behavior for yaw !=0 cases.

jpreiss:
Did anyone ever do a comparison between the INDI controller and Mellinger for aggressive attitudes? If another controller performs significantly better, that would be evidence that this is a software bug.

knmcguire:
I've compared the PID and INDI controller, as the INDI controller should be able to withstand sudden distrurbances and windgusts, but not against Mellinger... 

Is there a trajectory that forces the 30 deg roll/pitch that we can compare the both with?

whoenig:
Executing https://raw.githubusercontent.com/USC-ACTLab/crazyswarm/master/ros_ws/src/crazyswarm/scripts/figure8.csv at 0.6 timescale results in 34 deg max roll and 23 deg max pitch. I remember that this results usually in crashes with the mellinger controller.

knmcguire:
This was very long over due.... but I finally got around to test this. I tested the trajectory on PID, Mellinger and INDI. Although the Crazyflie didn't crash, it did behave more aggressive and wobbly during the maneuver. With bad motors/propellers it probably crashed for sure. 

[...]

PR #1596, opened 2026-02-04
Mellinger controller angle limitation fix

ratheron:
The other day we have observed that the Mellinger controller behaves different from our implementation in [Crazyflow](https://github.com/utiasDSL/crazyflow). We expected that when sending the attitude command via rpy angles, the desired orientation matrix gets constructed from that. After some investigation, however, we noticed that it is not correctly converted back into a rotation matrix onboard, but rather the target_thrust vector with a fixed z length of 1, effectively limiting the maximum achievable angle, since the x and y length are computed from the sin of pitch and roll. 

Here is a plot of the simulation (first plot) vs reality (second plot) for the brushless drone, where we can observe a flattening of the angle at around 45 deg (as expected from the argument above). Look for the roll and pitch command vs observed angles:
sim_correct 
real_bug

To test if line 167, 168, and 174 actually cause this, we implemented this in our sim too and now sim and real look very similar:
sim_bug

The fix simply includes computing the rotation matrix from the RPY angles. With the fix applied in the firmware, we observe the following results in real, which now look very similar to the first plot in sim. We can see that this implementation enables way more aggressive angles:
real_correct

We stumbled across #832 and could imagine that this bug causes this limitation, but are not entirely sure what the observed behavior is. Maybe @whoenig still remembers what exactly was the issue. Either way, we have flown the brushless and the legacy drone with full state commands and RPYT interface and it appears to be stable in both cases.

### DIFF
--- a/src/modules/src/controller/controller_mellinger.c
+++ b/src/modules/src/controller/controller_mellinger.c
@@ -158,23 +158,6 @@ void controllerMellinger(controllerMellinger_t* self, control_t *control, const 
   self->i_error_y += r_error.y * dt;
   self->i_error_y = clamp(self->i_error_y, -self->i_range_xy, self->i_range_xy);
 
-  // Desired thrust [F_des]
-  if (setpoint->mode.x == modeAbs) {
-    target_thrust.x = self->mass * setpoint->acceleration.x                       + self->kp_xy * r_error.x + self->kd_xy * v_error.x + self->ki_xy * self->i_error_x;
-    target_thrust.y = self->mass * setpoint->acceleration.y                       + self->kp_xy * r_error.y + self->kd_xy * v_error.y + self->ki_xy * self->i_error_y;
-    target_thrust.z = self->mass * (setpoint->acceleration.z + GRAVITY_MAGNITUDE) + self->kp_z  * r_error.z + self->kd_z  * v_error.z + self->ki_z  * self->i_error_z;
-  } else {
-    target_thrust.x = -sinf(radians(setpoint->attitude.pitch));
-    target_thrust.y = -sinf(radians(setpoint->attitude.roll));
-    // In case of a timeout, the commander tries to level, ie. x/y are disabled, but z will use the previous setting
-    // In that case we ignore the last feedforward term for acceleration
-    if (setpoint->mode.z == modeAbs) {
-      target_thrust.z = self->mass * GRAVITY_MAGNITUDE + self->kp_z  * r_error.z + self->kd_z  * v_error.z + self->ki_z  * self->i_error_z;
-    } else {
-      target_thrust.z = 1;
-    }
-  }
-
   // Rate-controlled YAW is moving YAW angle setpoint
   if (setpoint->mode.yaw == modeVelocity) {
     desiredYaw = state->attitude.yaw + setpoint->attitudeRate.yaw * dt;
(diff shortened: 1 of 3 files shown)


ratheron:
Yes, we have removed this part intentionally. We have checked and can observe the same behavior before and after our change:
- In full state control mode, the drone levels when the connection is lost, because it simply stays at the last commanded setpoint. After around 2s, the drone drops.
- In attitude control mode (RPYT), also the last setpoint is applied. This means the drone continues with the same attitude and thrust. After around 0.5s, the drone levels, but continues to drift with roughly constant velocity. After around 2s, the drone drops, too.

I don't know what the commander is doing after 500ms when the connection is lost, i.e., what the mode is set to. If we want the controller to level and hover for the RPYT interface too, what I would endorse, we first need to know what the applied logic is.

src/modules/src/controller/controller_mellinger.c
```diff
@@ -191,31 +174,42 @@ void controllerMellinger(controllerMellinger_t* self, control_t *control, const
...
-    struct vec y_yaw = vcross(mkvec(0, 0, 1), x_yaw);
-    struct mat33 R_yaw_only = mcolumns(x_yaw, y_yaw, mkvec(0, 0, 1));
-    target_thrust = mvmul(R_yaw_only, target_thrust);
-  }
-
-  // Current thrust [F]
-  current_thrust = vdot(target_thrust, z_axis);
+  current_thrust = 0.0f;
+  
+  if (setpoint->mode.x == modeAbs) {
```
stefanthorstenson, line 178: Is there somewhere a description of which sets of modes that can be applied? Are they dependent? It seems like mode.x == modeAbs is connected with mode.z == modeDisable, but it is not clear.

What I'm trying to get to is to clarify what code is used when flying e.g. in manual mode, if that is possible.

stefanthorstenson: Irrelevant for this PR

stefanthorstenson (dismissed):
Looks good!

gemenerik (changes requested):
Great fix for the angle limitation, the analysis and plots are convincing.

One concern: I think this PR may break some combinations of control modes. For instance, I tested with the cfclient hover assist mode (XY velocity + Z position hold) and it no longer produces thrust. There may be other affected combinations as well.

I think it should be possible to bring these modes back while keeping the RPY fix, since the orientation and thrust computation can be handled independently.

ratheron:
Thanks for pointing this out @gemenerik. As @stefanthorstenson was also wondering, is there any documentation where we can find out which mode combinations there are? It's hard to fix this without having the full picture. I found [this enum](https://github.com/ratheron/crazyflie-firmware-sysid/blob/master/src/modules/interface/stabilizer_types.h#L245-L249) of the modes so far.

ratheron:
Thanks for coming back so fast. Sounds good. So There is nothing to do from my side?

Fyi, I just found out that due to the allowed more aggressive behavior, the brushed cf seems less stable and crashes faster (when trying to track kinky references only) than before with our offboard position 2 attitude Mellinger controller (same weights as the onboard one). This will most likely also be the case when using the onboard position to attitude part. Have you noticed anything in that regard when flying aggressive trajectories? The brushless seems to be very well tuned now.

src/modules/src/controller/controller_mellinger.c
```diff
@@ -191,31 +174,66 @@ void controllerMellinger(controllerMellinger_t* self, control_t *control, const
...
-  // yaw correction (only if position control is not used)
-  if (setpoint->mode.x != modeAbs) {
-    struct vec x_yaw = mcolumn(R, 0);
-    x_yaw.z = 0;
-    x_yaw = vnormalize(x_yaw);
-    struct vec y_yaw = vcross(mkvec(0, 0, 1), x_yaw);
-    struct mat33 R_yaw_only = mcolumns(x_yaw, y_yaw, mkvec(0, 0, 1));
-    target_thrust = mvmul(R_yaw_only, target_thrust);
+  // Calculate desired axes and current thrust
+  if (setpoint->mode.x == modeAbs) {
```
gemenerik, line 178: Branch fires on mode.x == modeAbs but reads y/z slots without checking their modes, implicitly assumes all three are modeAbs. Flagging for the follow-up issue

gemenerik (approved):

src/modules/src/controller/controller_mellinger.c
stefanthorstenson: Good comment! Created #1628. In this PR, only the manual mode of the Mellinger controller has functional changes. All other changes are refactoring.

stefanthorstenson:
@ratheron , we are ready to merge. Thanks a lot for the nice contribution!

PR #1617, opened 2026-03-23
Introduce onboard guidance abstraction layer

gemenerik:
Introduce "onboard guidance" subsystem that generates setpoints onboard the Crazyflie; trajectory following, autonomous navigation, etc. Until now, the High-Level Commander was hard-wired into the stabilizer and commander with no clear interface boundary.

This PR introduces onboard guidance as a proper abstraction with a defined interface:                                                                                                                                                                       
- `init` / `test`: initialize the guidance implementation and verify it started correctly
- `getSetpoint`: called every stabilizer iteration (~1 kHz), produces a setpoint from the current state
- `stop`: called when a higher-priority setpoint source takes over
- `tellState`: called when guidance is re-enabled after a higher-priority source, providing a state snapshot
- `block`: called every stabilizer iteration, must suppress setpoints when motors are not allowed to run (safety-critical)
- `isDone`: reports whether the current guidance task (e.g. trajectory) has finished

The High-Level Commander is the built-in implementation. Out-of-tree implementations provide these 7 functions and are linked via the dispatch table.                                                              

Firmware changes:                                                                                                                                                                                                                                           
- Add `onboard_guidance.c` orchestrator with dispatch table and auto-select, following the controller/estimator pattern
- Add `stabilizer.guidance` parameter for runtime switching
- Add `CONFIG_ONBOARD_GUIDANCE_HLC_ENABLE` so OOT-only builds can reclaim ~12 KB flash and ~6 KB RAM by excluding the HLC
- Rename `COMMANDER_PRIORITY_HIGHLEVEL` to `COMMANDER_PRIORITY_ONBOARD_GUIDANCE`
- Add `examples/app_out_of_tree_onboard_guidance/` that delegates to the HLC

Documentation:
- Add `onboard_guidance.md` covering the interface, HLC, priority switching, and OOT contract
- Split controller and estimator config docs to their respective pages
- Update OOT docs with onboard guidance section


src/modules/src/onboard_guidance.c
```diff
@@ -0,0 +1,119 @@
...
+
+#ifdef CONFIG_ONBOARD_GUIDANCE_HLC_ENABLE
+static bool hlcTestAlwaysTrue(void) { return true; }
+static void hlcStopVoid(void) { crtpCommanderHighLevelStop(); }
+static void hlcBlockVoid(bool doBlock) { crtpCommanderBlock(doBlock); }
+#endif
+
+
+static OnboardGuidanceFcns guidanceFunctions[] = {
+  {.init = 0, .test = 0, .getSetpoint = 0, .stop = 0, .tellState = 0, .block = 0, .isDone = 0, .name = "None"},
```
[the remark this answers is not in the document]
gemenerik: The init guard rejects types whose function pointers are NULL, and that the Kconfig select ensures auto-select always has HLC compiled in; so the "None" slot is unreachable in practice, making 7 no-op functions unnecessary

src/modules/src/onboard_guidance.c
```diff
@@ -0,0 +1,123 @@
...
+
+  if (selected == OnboardGuidanceTypeAutoSelect) {
+    #if defined(CONFIG_ONBOARD_GUIDANCE_HIGHLEVEL_COMMANDER)
+      selected = OnboardGuidanceTypeHighLevelCommander;
+    #elif defined(CONFIG_ONBOARD_GUIDANCE_OOT)
+      selected = OnboardGuidanceTypeOot;
+    #elif defined(CONFIG_ONBOARD_GUIDANCE_HLC_ENABLE)
+      selected = DEFAULT_GUIDANCE;
+    #else
+      ASSERT_FAILED();
```
ataffanel: Since this is a compile-time problem I think it should be an `#error` and not an assert: this will prevent the firmware to compile at all rather than crashing at runtime.

ataffanel (commented):
Great addition! AJust a comment about compile time error Vs runtime assert.

src/modules/src/onboard_guidance.c
```diff
@@ -0,0 +1,123 @@
...
+  void (*stop)(void);
+  void (*tellState)(const state_t *state);
+  void (*block)(bool doBlock);
+  bool (*isDone)(void);
+  const char *name;
+} OnboardGuidanceFcns;
+
+#ifdef CONFIG_ONBOARD_GUIDANCE_HLC_ENABLE
+static bool hlcTestAlwaysTrue(void) { return true; }
+static void hlcStopVoid(void) { crtpCommanderHighLevelStop(); }
```
stefanthorstenson, line 30: Why is the function called hlcStop_Void_? Same on the next row. It is clear from the function declaration that it returns void.

src/modules/src/onboard_guidance.c
```diff
@@ -0,0 +1,123 @@
...
+
+
+static OnboardGuidanceFcns guidanceFunctions[OnboardGuidanceType_COUNT] = {
+  [OnboardGuidanceTypeAutoSelect] = {.name = "None"},
+#ifdef CONFIG_ONBOARD_GUIDANCE_HLC_ENABLE
+  [OnboardGuidanceTypeHighLevelCommander] = {
+    .init = crtpCommanderHighLevelInit,
+    .test = hlcTestAlwaysTrue,
+    .getSetpoint = crtpCommanderHighLevelGetSetpoint,
+    .stop = hlcStopVoid,
```
stefanthorstenson, line 42: Why is the crtp function not called directly? crtpCommanderHighLevelStop

src/modules/src/onboard_guidance.c
```diff
@@ -0,0 +1,123 @@
...
+    .tellState = crtpCommanderHighLevelTellState,
+    .block = hlcBlockVoid,
```
stefanthorstenson, line 44: Why is the crtp function not called directly? crtpCommanderBlock

src/modules/src/supervisor.c
```diff
@@ -402,6 +404,9 @@ static void updateLogData(SupervisorMem_t* this, const supervisorConditionBits_t
     this->infoBitfield |= 0x0080;
   }
 
+// HLC planner state bits (0x0100-0x0400). These are HLC-specific and will
+// be zero when using an out-of-tree onboard guidance implementation.
+#ifdef CONFIG_ONBOARD_GUIDANCE_HLC_ENABLE
   enum trajectory_state traj_state =  crtpCommanderHighLevelGetPlannerState();
```
stefanthorstenson, line 410: This PR adds a really nice abstraction layer with the guidance. However, it is broken here and bypassed by directly calling high level commander. Can we find another solution to this? It seems it is only used as log data, so that could most likely be moved to the guidance module instead.

Apart from being cleaner, it would also mean better modularization and independence, which is better for testing and interface stability.

src/modules/src/supervisor.c
```diff
@@ -46,8 +46,10 @@
 #define DEBUG_MODULE "SUP"
 #include "debug.h"
 
+#ifdef CONFIG_ONBOARD_GUIDANCE_HLC_ENABLE
```
stefanthorstenson, line 49: See comment below. If logging is moved to guidance, this ifdef section can be removed.

stefanthorstenson (changes requested):
Very nice PR! I like the added abstraction layer. Some comments, and some are just questions for my understanding...

A small comment (but maybe big when it comes to changing it 😄) is the name. onboard_guidance: To me it is obvious that the code in the firmware is onboard, so I would suggest to remove onboard and just call it guidance. I'm open for discussion :smile:

enyanil (commented):
Great documentation and addition! 
As discussed we could take this opportunity to update connected documentation as well. The following sections might need a check, especially images:
- [Stabilizer Module](https://www.bitcraze.io/documentation/repository/crazyflie-firmware/master/functional-areas/sensor-to-control/) 
- [The Commander Module](https://www.bitcraze.io/documentation/repository/crazyflie-firmware/master/functional-areas/sensor-to-control/commanders_setpoints/#the-commander-module)
- A new image to replace the [High Level Commander](https://www.bitcraze.io/documentation/repository/crazyflie-firmware/master/functional-areas/sensor-to-control/commanders_setpoints/#high-level-commander) one

src/modules/src/onboard_guidance.c
gemenerik: Good catch, fixed

PR #1618, opened 2026-03-27
Improved Landing

AlexGraefe:


ArisMorgens:
Taking over this PR from @AlexGraefe.

This PR adds a new, high-level commander command, `land_precise`, for more accurate landings.

Instead of a single straight-line descent, the Crazyflie hovers above the target with temporarily-tuned controller gains to let horizontal error settle out before the final descent.

This works well when there's accurate positioning data (Lighthouse or MoCap), enabling the Crazyflie to react to external disturbances. Good for successfully landing on Crazyflie Brushless charging docks.

### Here's a list of the main changes:

- **Controller layer**: added `enterPreciseLand()`/`exitPreciseLand()` to the existing controller dispatch table, so the planner can ask "the active controller" to temporarily swap in precise-landing gains without needing to know which controller is running. Currently, only PID and Mellinger have their own precise-landing gains (each with their own param group for tuning them);.
- **Planner**: new `plan_precise_land()` builds the 3-segment hover-then-descend trajectory via a new `TRAJECTORY_STATE_PRECISE_LANDING` state, and calls the controller to swap gains for the duration. Gains are restored if the landing is completed or interrupted.
- **High-level commander**: new CRTP command `COMMAND_LAND_PRECISE` with its own packet struct and public API (`crtpCommanderHighLevelPreciseLand()`).

PR #1619, opened 2026-03-30
Use direct to task notification to wake up sensors task

reucsan:
This PR is a minor improvement for optimizing the sensor task. Previously a binary semaphore was used to wake up the sensor task from the IMU ISR. Based on the FreeRTOS documentation, direct to task notifications have better performance and use less RAM.

> A direct to task notification is an event sent directly to a task, rather than indirectly to a task via an intermediary object such as a queue, event group or semaphore.

> Unblocking an RTOS task with a direct notification is 35% faster and uses less RAM than unblocking a task using an intermediary object such as a binary semaphore.

https://www.freertos.org/Documentation/02-Kernel/02-Kernel-features/03-Direct-to-task-notifications/01-Task-notifications


ataffanel (commented):
Looks good, thanks!

PR #1627, opened 2026-05-04
Added Reset PID Integeral for Position and Velocity Controller

ShuqinZ:
Added callback functions in Position and velocity PID controller to reset accumulated error for I-term. 
This address the problem of switching from using position/velocity setpoint to attitude setpoint then back to position/velocity setpoint, the I-term is affected by hitorical data that causes jittering problem. 
One may call cf.param.set_value("posCtlPid.resetI", "1") to reset for position controller, and cf.param.set_value("velCtlPid.resetI", "1") for velocity controller.

### DIFF
--- a/src/modules/src/controller/position_controller_pid.c
+++ b/src/modules/src/controller/position_controller_pid.c
@@ -163,6 +163,45 @@ static struct this_s this = {
 };
 #endif
 
+static uint8_t resetVelI;
+static volatile bool resetVelIRequested;
+static uint8_t resetPosI;
+static volatile bool resetPosIRequested;
+
+static void resetPidIntegral(PidObject *pid)
+{
+  pid->integ = 0.0f;
+  pid->outI = 0.0f;
+}
+
+static void resetHorizontalPositionPidIntegrals(void)
+{
+  resetPidIntegral(&this.pidX.pid);
+  resetPidIntegral(&this.pidY.pid);
+}
+
+static void resetHorizontalVelocityPidIntegrals(void)
+{
+  resetPidIntegral(&this.pidVX.pid);
+  resetPidIntegral(&this.pidVY.pid);
+}
+
+static void resetPositionPidIntegralsParamCallback(void)
+{
+  if (resetPosI) {
+    resetPosIRequested = true;
+    resetPosI = 0;
+  }
+}
+
+static void resetVelocityPidIntegralsParamCallback(void)
+{
+  if (resetVelI) {
+    resetVelIRequested = true;
+    resetVelI = 0;
+  }
+}
+
 void positionControllerInit()
 {
   pidInit(&this.pidX.pid, this.pidX.setpoint, this.pidX.pid.kp, this.pidX.pid.ki, this.pidX.pid.kd,
@@ -193,6 +232,11 @@ float state_body_x, state_body_y, state_body_vx, state_body_vy;
 void positionController(float* thrust, attitude_t *attitude, const setpoint_t *setpoint,
                                                              const state_t *state)
 {
+  if (resetPosIRequested) {
+    resetPosIRequested = false;
+    resetHorizontalPositionPidIntegrals();
+  }
+
   this.pidX.pid.outputLimit = xVelMax * velMaxOverhead;
   this.pidY.pid.outputLimit = yVelMax * velMaxOverhead;
   // The ROS landing detector will prematurely trip if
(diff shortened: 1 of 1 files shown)


stefanthorstenson:
Hi @ShuqinZ !

Thanks for opening this PR! It seems like it could be a very nice addition to the stability of the Crazyflie, and we would like to take the next step with this one. In order to make best use of your ideas and code, and to make sure it really works as intended, could you supply two examples (preferably scripts, but otherwise a brief guide with exact commands, depending on your workflow) on how to:

1. Trigger the described issue (before the suggested changes)
2. Perform the same flight trajectory, but now with suggested changes included (if there is a difference)

If you have any plots of the issue and the solution, you are welcome to supply them. If not, we can do it at a later stage if we feel that we need it.

Thanks in advance!

ShuqinZ:
Hi @stefanthorstenson, sorry for the late reply.
1. Scenario that triggers the described issue:
When calling cf.commander.send_position_setpoint(x, y, z, yaw), the cascaded PID controller start to compute.
Later my script switched to use cf.commander.send_zdistance_setpoint(, roll, pitch, yawrate, zdistance) inorder to maintain orientation. This commands sets position and velocity setpoint to be 0. The drone's position changed when this attitute command was kept issuing.
Then later when switching back to use cf.commander.send_position_setpoint(x, y, z, yaw) for position control, the accumulated error that contributes to the I-gain causes the drone to react to the position error before switching to attitude control, instead of based on the new control. (The script of mine can be found here: https://github.com/flslab/fls-cf-offboard-controller/blob/master/Interaction/interactions.py#L515).
2. With the fix, one may call  cf.param.set_value("posCtlPid.resetI", "1") and cf.param.set_value("velCtlPid.resetI", "1") to reset the accumulated error. So in the scenario described above, once switching back from attitude command to position command, the drone can follow the new issued position setpoint without being affection of the old setpoints.
Here I have a comparison. 
Without the reset, as the GIF shows, the drone flys toward the left a bit though the position setpoint was at the right side of it.
translation_2026-04-28_15-48-57

With the reset (command issued in https://github.com/flslab/fls-cf-offboard-controller/blob/master/Interaction/interactions.py#L969), the drone have minimal flying back and start to execute the next position setpoint.
Copy of translation_2026-04-30_14-55-37

stefanthorstenson:
Thanks @ShuqinZ for supplying more info and for the videos! It is very clear that you have found a bug, and your fix obviously makes it a lot better! Nice!

I'm not able to access your script though. Would you mind pasting relevant parts in this PR?

What I would like to do when moving forward with this PR are mainly two things:
1. Analyze if the bug you have found has wider reach than just this specific example.
2. Decide on where in software to put the fix.

I will dig into 1. so you don't have to think about that. About 2. though, in a perfect world the drone would fix this without the interaction of offboard. Have you thought about if this is possible? I mean, to put the fix in the CF firmware.

ShuqinZ:
So the code is :
cf.commander.send_position_setpoint(target_pos[0], target_pos[1], target_pos[2], 0)
...... 
cf.commander.send_hover_setpoint(0.0, 0.0, yawrate, 1.0)
...... 
cf.param.set_value("posCtlPid.resetI", "1")
cf.param.set_value("velCtlPid.resetI", "1")
...... 
cf.commander.send_position_setpoint(target_pos[0], target_pos[1], target_pos[2], 0)

The reset happens between switching controllers.

About 2, I have thought about it, but could not find a better way to address the problem.
To describe the problem, it is:
"A position control command was initially issued to the drone, then later a attitude control command was issued and the drone changed its position. Later to issue a position command, the reset needs to be done."

I guess we can always set a tracker in cf.commander, to keep track of the previous 2 type of command issued.
Another way to do this is to hardcode the error of position and velocity controller to be 0 when using attitude command.
Don't know if that would be a good idea.

stefanthorstenson:
Thank you for the code! And thanks for the input regarding how to solve the issue. I will start looking into this in a couple of weeks.

stefanthorstenson:
Hi @ShuqinZ !

I have been able to confirm and reproduce what you are seeing. It can also happen in the Z axis. I managed to create a solution in the Crazyflie firmware and made a PR:
https://github.com/bitcraze/crazyflie-firmware/pull/1645

There you can also read more about the issue and the proposed solution.

Can you verify that this fix works for you? And comment on the PR.

Thanks for the help!

ShuqinZ:
Thanks for the help!
I am out of town for now, will verify it once I have access to my drone.

stefanthorstenson:
Hi again @ShuqinZ ! PR #1645 is now fixed and merged. Let us know how it goes! I will thereby also close this PR.

PR #1629, opened 2026-05-07
Add note in docs for propeller configuration

ratheron:
As far as I know, there are no docs on how to flash the other versions of the drone, i.e., how to set the propellers to be legacy or to thrust upgrade kit. Personally, I always forget what I need to do, since I rarely flash those versions. Thus, I added a short note in the docs.


enyanil:
Hi Marcel, thanks for this contribution! I agree this is a bit unclear. It is described how to do this with menuconfig on the [Battery Compensation](https://www.bitcraze.io/documentation/repository/crazyflie-firmware/master/functional-areas/battery_compensation/) page. And, since this is more advanced, I am thinking that maybe we can add something about the battery compensation page in the "[Customize the firmware with kbuild (Advanced)](https://www.bitcraze.io/documentation/repository/crazyflie-firmware/master/building-and-flashing/build/#customize-the-firmware-with-kbuild-advanced)" section a little bit further down instead of in the first section. Do you think this helps clear things up? I can create a new PR for this if you want, let me know what you think!

ratheron:
Hi, you are absolutely right. I have never used the menuconfig, because it was broken in the pixi env on my end, but it's clearly the better solution. I've moved the note and added a fix to `make menuconfig` inside the provided pixi env. 

Should we also add a note that pixi is available in the first place?

ataffanel:
We have considered Pixi experimental so far since we felt it required more testing. It seems to work fine and help quite a lot so it may indeed be time to mention it in the doc.

enyanil:
Yes, I think we should add a note about pixi as well.

ratheron:
Do we want to do it with this PR? If so, where? We could add at the top of [docs/building-and-flashing/build.md](https://github.com/ratheron/crazyflie-firmware-sysid/blob/c748a74b0a2122d2b5064a188d50c74680bddb61/docs/building-and-flashing/build.md) that we can use pixi, then a brief section how to install pixi and start the environment?

ataffanel:
Yes we may do it in this PR.

ArisMorgens:
I added documentation on how to build and flash the firmware using pixi. I tried to keep it short.  @enyanil let me know what do you think.

@ratheron Since you have experience using pixi for the crazyflie-firmware, feel free to check if the instructions make sense, and if you would recommend a different approach.

docs/building-and-flashing/pixi.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Development shell
+
+To get into a development shell where required tools for building and flashing the project are present run:
+
+``` bash
+pixi shell
+```
+
+## Building with pixi
```
enyanil: The way this is laid out is a little confusing, is it meant to be a step by step, which would mean you have to do pixi shell to be able to do the run bash, but if I understand correctly you don't? Maybe it would help to flesh out the "Development shell" section a bit with some more info about what it does to make it more independent.

docs/building-and-flashing/pixi.md
```diff
@@ -0,0 +1,39 @@
...
+
+The following command will produce a clean build for the Crazyflie 2.1(+). The command can be changed to build for other targets:
```
enyanil: Add something about from where this should be run?

docs/building-and-flashing/pixi.md
```diff
@@ -0,0 +1,39 @@
+---
+title: Building using pixi
+page_id: pixi
+---
+
+The [pixi package manager](https://pixi.sh) can be used to build and flash the firmware without installing the ARM toolchain, Python or the Crazyflie client on your system.
```
ratheron: You still install it on your machine, just not system-wide. I would rephrase this as
```suggestion
The [pixi package manager](https://pixi.sh) can be used to build and flash the firmware without installing the ARM toolchain, Python or the Crazyflie client system-wide on your machine.
```

docs/building-and-flashing/pixi.md
ratheron: Personal preference: I would not show the full building and flashing pipeline here, since this is a repetition from the regular one. I would just say that inside the pixi shell, one can run the regular building and flashing commands. If you want, you can still say that `pixi run <command>` also works, if one doesn't want to activate the shell.

enyanil: Or is it "more correct" to do it inside the shell?

docs/building-and-flashing/pixi.md
ArisMorgens: I added a comment right above the Development shell section in [fc91ba2](https://github.com/bitcraze/crazyflie-firmware/pull/1629/commits/fc91ba25587d9afbb46f297602fe60d5b2c20303).

docs/building-and-flashing/pixi.md
ArisMorgens: Good point. Fixed in [fc91ba2](https://github.com/bitcraze/crazyflie-firmware/pull/1629/commits/fc91ba25587d9afbb46f297602fe60d5b2c20303).

docs/building-and-flashing/pixi.md
ArisMorgens: Makes sense to have the shell as the recommended way to use pixi. Fixed in [fc91ba2](https://github.com/bitcraze/crazyflie-firmware/pull/1629/commits/fc91ba25587d9afbb46f297602fe60d5b2c20303).

ArisMorgens: It seems "more correct" to run the make commands from inside the shell, otherwise you have to run every individual command with a `pixi run` prefix. I think I made the instructions a bit more clear in [fc91ba2](https://github.com/bitcraze/crazyflie-firmware/pull/1629/commits/fc91ba25587d9afbb46f297602fe60d5b2c20303), so let me know what you think.

enyanil (dismissed):
Looks good!

docs/building-and-flashing/pixi.md
```diff
@@ -0,0 +1,34 @@
+---
+title: Building using pixi
+page_id: pixi
+---
+
+The [pixi package manager](https://pixi.sh) can be used to build and flash the firmware without installing the ARM toolchain, Python or the Crazyflie client system-wide on your machine.
+All that is needed is to have pixi installed, following the [pixi installation instructions](https://pixi.sh/latest/#installation).
+
+> On Windows, pixi isn't supported for this project (the ARM toolchain and `ncurses` dependencies aren't built for `win-64`). Use WSL instead: [enable WSL](https://docs.microsoft.com/en-us/windows/wsl/install), then follow the instructions below from inside your WSL Ubuntu shell.
```
ataffanel, line 9: We may want to note somewhere that on Windows `make cload` still uses the windows native python.exe in order to be able to use the radio usb dongle from wsl.

So on Windows Pixi solves the compilation stage. Native python lib install is still required to flash the Crazyflie.

ataffanel (changes requested):
Looks good to me, I love that we finally get Pixi in! :-)

Just one small comment about windows: we may want to add that windows still requires native python to bootload.

docs/building-and-flashing/pixi.md
ArisMorgens: Good point. I added a comment in [7719114](https://github.com/bitcraze/crazyflie-firmware/pull/1629/commits/77191140c91d27e60907e5fef09eb6094604b961). Let me know if it looks good.

ataffanel (approved):

PR #1630, opened 2026-05-07
USB stack: drop shim task, direct-kick TX, deeper TX queue

evoggy:
## Summary

Three independent (but related) USB-stack changes that together drop ping latency by ~65× and eliminate a class of disconnects under heavy traffic.

### 1. Drop the `usblinkTask` shim

The `usbd_cf_DataOut` ISR used to push raw `USBPacket`s into a `usbDataRx` queue, where the priority-3 `usblinkTask` would dequeue them, reshape into `CRTPPacket`, and re-enqueue into `crtpPacketDelivery`. The ISR now builds the `CRTPPacket` directly into a static `rxCrtp` buffer and `xQueueSendFromISR`s straight into `crtpPacketDelivery`. The shim task and its 5-deep `USBPacket` queue are gone.

A new `usbResumeRx()` is called by `usblinkReceivePacket` after each successful dequeue, to re-arm the OUT EP if the ISR halted RX because the queue was full. (`usbGetDataBlocking` is removed — the shim was its only caller.)

### 2. Direct-kick on `usbSendData`

`usbSendData` now starts the transfer immediately when the queue was empty *and* no transfer is in flight, instead of waiting up to 1 ms for the next SOF. The SOF kicker remains as a safety net.

Implemented under `NVIC_DisableIRQ(OTG_FS_IRQn)` + `xQueueReceive(usbDataTx, &outPacket, 0)`, reusing the static `outPacket` buffer the DataIn/SOF ISRs already use. The masked-IRQ window plus `doingTransfer == false` guarantees no concurrent ISR is reading `outPacket`.

### 3. Smaller TX-path memcpy chain + deeper queue

- `usbSendData` API: `(uint32_t size, uint8_t* data)` → `(USBPacket*)`. Caller (`usblinkSendPacket`) now formats wire bytes directly into a static `USBPacket` and passes it by pointer. The intermediate `outStage` and `sendBuffer` static buffers are gone (one fewer memcpy per outbound packet, −64 B RAM).
- `usbDataTx` queue depth `1` → `8` for burst headroom under chained DataIn ISR pacing.
- `doingTransfer` / `doingVcpTransfer` made `volatile` (correctness fix — they're written from the ISR and read from a task-context critical section).

## Effect

Measured with `cfcli test link-perf` over USB:

| metric             | before    | after        |
|--------------------|----------:|-------------:|
| ping avg (n=1000)  | ~21 ms*   | **~0.3 ms**  |
| uplink (sink)      | 706 kbit/s + occasional disconnect | 800–860 kbit/s |
| downlink (source)  | 0 (disconnect) | ~410 kbit/s |
| echo (round-trip)  | 0 (disconnect) | ~410 kbit/s |
| firmware RAM       | —         | **−664 B** (queue + shim stack) |

\* The 21 ms ping was actually a host-side comm-thread artefact (single-loop read+write); the firmware-side latency improvement here is on top of a parallel host fix.

## Notes

- No changes to interface descriptors, EP layout, USB FIFO sizes, or `crtpPacketDelivery` queue depth (16). The `radiolink` path is untouched.
- ISR no longer calls `portYIELD_FROM_ISR` on the CF EP — adding it was measured to add ~0.6 ms ping RTT (woken consumer gets preempted before draining; ISR's job of re-arming the OUT EP is more time-critical). The CDC EP path didn't yield either; behaviour is preserved.

## Test plan

- [x] `cfcli test link-perf` (USB): completes cleanly; ping ~0.3 ms; uplink ~800 kbit/s; downlink/echo ~410 kbit/s.
- [x] `cfcli bootload flash --bin bcCam:qcc=...` works (heavy memory-write traffic over USB).
- [x] `cfcli test link-perf` over radio — unchanged behaviour.
- [x] No regression in normal radio operation (ping, log, param subscribe/get).

### DIFF
--- a/src/hal/src/usb.c
+++ b/src/hal/src/usb.c
@@ -55,17 +55,16 @@
 NO_DMA_CCM_SAFE_ZERO_INIT __ALIGN_BEGIN USB_OTG_CORE_HANDLE    USB_OTG_dev __ALIGN_END ;
 
 static bool isInit = false;
-static bool doingTransfer = false;
-static bool doingVcpTransfer = false;
+static volatile bool doingTransfer = false;
+static volatile bool doingVcpTransfer = false;
 static bool rxStopped = true;
 static uint16_t command = 0xFF;
 
 
-// This should probably be reduced to a CRTP packet size
-static xQueueHandle usbDataRx;
-STATIC_MEM_QUEUE_ALLOC(usbDataRx, 5, sizeof(USBPacket)); /* Buffer USB packets (max 64 bytes) */
+// Owned by usblink.c; cached at init for the DataOut ISR to write into.
+static xQueueHandle crtpRxQueue = NULL;
 static xQueueHandle usbDataTx;
-STATIC_MEM_QUEUE_ALLOC(usbDataTx, 1, sizeof(USBPacket)); /* Buffer USB packets (max 64 bytes) */
+STATIC_MEM_QUEUE_ALLOC(usbDataTx, 8, sizeof(USBPacket)); /* Buffer USB packets (max 64 bytes) */
 
 #define USB_CDC_CONFIG_DESC_SIZ     98
 
@@ -282,6 +281,7 @@ static uint8_t  usbd_cdc_EP0_RxReady(void *pdev);
 static USBPacket inPacket;
 static USBPacket outPacket;
 static USBPacket outVcpPacket;
+static CRTPPacket rxCrtp;
 
 /* CDC interface class callbacks structure */
 USBD_Class_cb_TypeDef cf_usb_cb =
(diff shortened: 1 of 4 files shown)


src/hal/src/usb.c
```diff
@@ -781,32 +807,19 @@ bool usbTest(void)
   return isInit;
 }
 
-bool usbGetDataBlocking(USBPacket *in)
+bool usbSendData(USBPacket *pkt)
```
ataffanel, line 810: A small check for pkt!=NULL and the size being within limit would be cheap and might prevent future problem here.

ataffanel (dismissed):
Looks good! Just a small non-blocking nit about adding a check on `pkt`, if implemented assume I still approve :).

PR #1631, opened 2026-05-20
Enable suport for WiFi camera deck

evoggy:
This PR implements the deck driver for the WiFi camera deck:

- Enable DMA on UART1 by default if not on a Bolt based platform
- Enable firmware update of the WiFi camera deck (QCC748)

### DIFF
--- a/src/deck/drivers/src/bccam_deck.c
+++ b/src/deck/drivers/src/bccam_deck.c
@@ -1,0 +1,502 @@
+/**
+ * ,---------,       ____  _ __
+ * |  ,-^-,  |      / __ )(_) /_______________ _____  ___
+ * | (  O  ) |     / __  / / __/ ___/ ___/ __ `/_  / / _ \
+ * | / ,--´  |    / /_/ / / /_/ /__/ /  / /_/ / / /_/  __/
+ *    +------`   /_____/_/\__/\___/_/   \__,_/ /___/\___/
+ *
+ * Crazyflie control firmware
+ *
+ * Copyright (C) 2026 Bitcraze AB
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, in version 3.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ *
+ * bccam_deck.c - Deck driver for the bcCam deck (QCC744)
+ */
+
+#include <string.h>
+
+#include "FreeRTOS.h"
+#include "task.h"
+
+#include "system.h"
+#include "deck.h"
+#include "param.h"
+#include "log.h"
+#include "i2cdev.h"
+#include "deckctrl_gpio.h"
+#include "uart1.h"
+#include "console.h"
+#include "bccam_deck.h"
+
+#define DEBUG_MODULE "BCCAM"
+#include "debug.h"
+
+#define BCCAM_UART_BAUDRATE       2000000
+#define BCCAM_ISP_BAUDRATE        500000
+#define BCCAM_TASK_STACKSIZE  configMINIMAL_STACK_SIZE
+#define BCCAM_TASK_PRI        tskIDLE_PRIORITY
+
+// Deck controller GPIO pin mapping (STM32 deck-ctrl MCU)
+#define GPIO_UART1_EN   DECKCTRL_GPIO_PIN_0   // PA0 - UART1 enable
+#define GPIO_UART2_EN   DECKCTRL_GPIO_PIN_1   // PA1 - UART2 enable
+#define GPIO_QCC_EN     DECKCTRL_GPIO_PIN_2   // PA2 - QCC744 chip enable
+#define GPIO_QCC_BOOT   DECKCTRL_GPIO_PIN_7   // PA7 - QCC744 boot select
+#define GPIO_PWR_EN     DECKCTRL_GPIO_PIN_12  // PC15 - Power enable
+
+// QCC744 UART ISP protocol constants
+#define ISP_HANDSHAKE_BYTE      0x55
+#define ISP_HANDSHAKE_COUNT     32
+#define ISP_HANDSHAKE_RESP_LEN  16   // "Boot2 ISP Ready\0"
+#define ISP_HANDSHAKE_TIMEOUT   M2T(8000)
+#define ISP_CMD_TIMEOUT         M2T(5000)
+
+// ISP command IDs
+#define ISP_CMD_GET_BOOTINFO    0x10
+#define ISP_CMD_CHANGE_RATE     0x20
+#define ISP_CMD_RESET           0x21
+#define ISP_CMD_CLK_SET         0x22
+#define ISP_CMD_FLASH_ERASE     0x30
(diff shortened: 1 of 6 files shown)


src/drivers/src/uart1.c
```diff
@@ -232,10 +232,8 @@ void uart1SendDataDmaBlocking(uint32_t size, uint8_t* data)
     xSemaphoreTake(uartBusy, portMAX_DELAY);
     // Wait for DMA to be free
     while(DMA_GetCmdStatus(UART1_DMA_STREAM) != DISABLE);
-    //Copy data in DMA buffer
-    memcpy(dmaBuffer, data, size);
```
ataffanel, line 236: Since we do not guarantee a DMA safe buffer anymore I suggest editing the documentation to make the DMA-able buffer a clear requirement and maybe adding an assert so that whoever hitting this bug will have a clear indication of where the problem happened.

src/deck/drivers/src/bccam_deck.c
```diff
@@ -0,0 +1,589 @@
...
+          }
+        }
+      }
+    }
+  }
+
+  if (!gotOk) {
+    DEBUG_PRINT("ISP: Handshake timeout (no OK from ROM)\n");
+    uart1Init(BCCAM_UART_BAUDRATE);
+    goto done;
```
ataffanel: Should we make sure the deck is back into firmware mode or some known-mode if we cannot reach bootloader? Here we just leave everything dangling.

ataffanel (changes requested):

src/deck/drivers/src/bccam_deck.c
```diff
@@ -0,0 +1,589 @@
...
+  uint8_t dummy;
+  TickType_t end = xTaskGetTickCount() + M2T(timeoutMs);
+  while (xTaskGetTickCount() < end) {
+    if (!uart1GetDataWithTimeout(&dummy, M2T(5))) {
+      break;
+    }
+  }
+}
+
+static bool ispSendCmd(uint8_t cmdId, const uint8_t *payload, uint16_t payloadLen) {
```
gemenerik: Bit of a nit, but this always returns true, maybe misleading downstream to return bool?

src/deck/drivers/src/bccam_deck.c
```diff
@@ -0,0 +1,589 @@
...
+  if (!ispRecvBytes(lenBytes, 2, ISP_CMD_TIMEOUT)) {
+    return false;
+  }
+
+  uint16_t respLen = (uint16_t)lenBytes[0] | ((uint16_t)lenBytes[1] << 8);
+  if (respLen != len) {
+    DEBUG_PRINT("ISP: Read length mismatch (%u vs %u)\n", respLen, len);
+    // Drain the announced payload so the queue isn't left desynced for the
+    // next command.
+    ispDrainRx(100);
```
gemenerik, line 241: How fragile is draining by time?

src/deck/drivers/src/bccam_deck.c
```diff
@@ -0,0 +1,589 @@
...
+  // Reset write-session state — protects against an interrupted previous
+  // flash leaving stale counters.
+  flashErased = false;
+  bytesWritten = 0;
+  writeBufUsed = 0;
+
+  uint8_t handshake[ISP_HANDSHAKE_COUNT];
+  memset(handshake, ISP_HANDSHAKE_BYTE, ISP_HANDSHAKE_COUNT);
+
+  ispDrainRx(50);
```
gemenerik, line 348: Similarly, is time based drain safe enough?

src/deck/drivers/src/bccam_deck.c
```diff
@@ -0,0 +1,589 @@
...
+  // Disable QCC748
+  deckctrl_gpio_write(deckInfoG, GPIO_QCC_EN, LOW);
+  vTaskDelay(M2T(50));
+
+  // Select ROM bootloader (BOOT pin HIGH) before re-enabling
+  deckctrl_gpio_write(deckInfoG, GPIO_QCC_BOOT, HIGH);
+  vTaskDelay(M2T(10));
+
+  // Switch UART to ROM bootloader baudrate (500000)
+  uart1Init(BCCAM_ISP_BAUDRATE);
```
gemenerik: Repeated uart1Init is possibly fragile? Would potentially be safer to add a baudrate switch function.

src/deck/drivers/src/bccam_deck.c
gemenerik: There's more of these so I'll not rewrite this comment everywhere

src/deck/drivers/src/bccam_deck.c
```diff
@@ -0,0 +1,589 @@
...
+    return false;
+  }
+
+  DEBUG_PRINT("bcCam deck test passed\n");
+  return true;
+}
+
+static const DeckDriver bccam_deck = {
+  .vid = 0xBC,
+  .pid = 0x21,
```
gemenerik: We could continue the table with 0x15
https://www.bitcraze.io/documentation/system/platform/cf2-expansiondecks/#deck-info

src/deck/drivers/src/bccam_deck.c
```diff
@@ -0,0 +1,589 @@
...
+  if (!gpioSetup(info, GPIO_UART1_EN, OUTPUT, HIGH)) {
+    DEBUG_PRINT("Failed to enable UART1\n");
+    return;
+  }
+
+  // Initialize UART1 for QCC748 communication
+  uart1Init(BCCAM_UART_BAUDRATE);
+
+  xTaskCreate(bcCamWorkerTask, "bcCam", configMINIMAL_STACK_SIZE * 4,
+              NULL, tskIDLE_PRIORITY, NULL);
```
gemenerik, line 561: The idle priority may become an issue when we start communicating with the deck in fw mode, but fine for now. Just a heads up.

gemenerik (changes requested):
Nice work! Couple of small suggestions/questions.

src/drivers/src/uart1.c
```diff
@@ -232,10 +231,8 @@ void uart1SendDataDmaBlocking(uint32_t size, uint8_t* data)
...
+    DMA_InitStructureShare.DMA_Memory0BaseAddr = (uint32_t)data;
     DMA_InitStructureShare.DMA_BufferSize = size;
-    initialDMACount = size;
     // Init new DMA stream
     DMA_Init(UART1_DMA_STREAM, &DMA_InitStructureShare);
```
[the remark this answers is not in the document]
evoggy: Fixed with comment and assert

src/deck/drivers/src/bccam_deck.c
evoggy: Fixed, jumps back to firmware

src/deck/drivers/src/bccam_deck.c
evoggy: Fixed

src/deck/drivers/src/bccam_deck.c
evoggy: Added documentation for this

src/deck/drivers/src/bccam_deck.c
evoggy: Fixed with new function

src/deck/drivers/src/bccam_deck.c
evoggy: Moved to 0x15

src/deck/drivers/src/bccam_deck.c
evoggy: Documented now

gemenerik (approved):

src/deck/drivers/src/bccam_deck.c
ataffanel: For the time being all communication will be configuration or low priority state update so idle should be ok for now. Good point thought.

ataffanel (approved):

Issue #1633, opened 2026-05-21
Setting persistent parameters through app api


thewired-nl:
`param_logic.h` has functions for getting/setting parameters from within the code (like apps). However, this does not update persistent parameters in EEPROM. The `paramPersistentStore` function that handles this when done through CRTP is explicitly marked for internal use only (in the paramTask).

For now I've hacked around this limitation by replacing the CRTP link so I can inject packets, but that's obviously not ideal

ArisMorgens:
Hello @thewired-nl!

Thanks for pointing this out. I pushed a branch in crazyflie-firmware that should allow this functionality:
https://github.com/bitcraze/crazyflie-firmware/tree/Aris/persistent_param

It basically adds `paramPersistentStoreByVarId(paramVarId_t varid)` to the public API in **param_logic.h**, following the same pattern as `paramSetInt/Float`. Check it out and let me know what you think!

thewired-nl:
This looks great! Unfortunately I forgot about another limiting factor in my usecase, where this is running in an ISR, so the normal semaphore functions don't work. I will move the handling code out of the interrupt and test again, thanks!

thewired-nl:
Outside the ISR this patch works as expected

ArisMorgens:
Thank you for testing! I will create a PR with these changes.
Do you think that this issue is then resolved, or do we need to further investigate it?

thewired-nl:
Yeah the PR resolves this issue, commented there as well wrt example code

PR #1634, opened 2026-05-27
Set persistent param from apps

ArisMorgens:
Based on #1633.
This pull request adds support for persisting parameter values directly from the firmware. 
It basically introduces `paramPersistentStoreByVarId`, and updates the internal CRTP handler `paramPersistentStore` to use this new function.

It also adds a simple example of how to use it. The example can be removed, but it helps with testing this PR.


thewired-nl:
I've tested this and it works well :) 
Would also be useful to add it to the app_api example code next the other param functions https://github.com/bitcraze/crazyflie-firmware/blob/master/app_api/src/app_main.c#L115

examples/app_persistent_param/src/app_persistent_param.c
```diff
@@ -0,0 +1,39 @@
...
+#define TARGET_THRESHOLD 25.0f
+
+void appMain()
+{
+  vTaskDelay(M2T(2000));
+  paramVarId_t idThreshold = paramGetVarId("health", "propTestThreshold");
+  DEBUG_PRINT("health.propTestThreshold = %.1f\n", (double)paramGetFloat(idThreshold));
+  vTaskDelay(M2T(2000));
+
+  if (!PARAM_VARID_IS_VALID(idThreshold)) {
```
gemenerik, line 47: this should probably run before line 22?

examples/app_persistent_param/src/app_persistent_param.c
gemenerik, line 1: Missing license header, no README (but maybe acceptable for such a simple self-documenting app)

src/modules/src/param_logic.c
```diff
@@ -697,14 +697,25 @@ static void generateStorageKey(const uint16_t index, char key[KEY_LEN])
   strcat(key, name);
 }
 
+bool paramPersistentStoreByVarId(paramVarId_t varid)
+{
+  ASSERT(PARAM_VARID_IS_VALID(varid));
+
+  if (!(params[varid.index].extended_type & (PARAM_PERSISTENT >> 8))) {
+    return false;
```
gemenerik, line 710: This makes PARAM_PERSISTENT enforced rather than advisory. Before, any param could be stored, and only clients (potentially) checked the flag. Good change, but two things: is anything relying on the old behavior (not cflib; it refuses non-persistent params client-side), and persistentParamFromStorage still restores any stored key at boot, so a param stored under old firmware keeps being applied while a new store of it is now refused. If we keep the check, enforce it there too.

gemenerik (changes requested):
Good addition

This is very important and unit testing architecture is in place, please add unit tests

examples/app_persistent_param/src/app_persistent_param.c
ArisMorgens: Good catch. Fixed in [26a5048](https://github.com/bitcraze/crazyflie-firmware/pull/1634/commits/26a50488f2d6768bf8c0068a1cff663dc19f77cb).

examples/app_persistent_param/src/app_persistent_param.c
ArisMorgens: Added licence header and a short README in [26a5048](https://github.com/bitcraze/crazyflie-firmware/pull/1634/commits/26a50488f2d6768bf8c0068a1cff663dc19f77cb).

src/modules/src/param_logic.c
ArisMorgens: I added the same `paramIsPersistent` check at boot restore in [79f4088](https://github.com/bitcraze/crazyflie-firmware/pull/1634/commits/79f4088baa8e8d376fac4fed97461c7a932a77bb) and it seems that nothing is relying on the old behavior where we didn't check if the parameter is indeed persistent.

gemenerik (approved):

PR #1635, opened 2026-05-27
add state streamer example for nanocockpit

hmllr:
in accordance with https://github.com/bitcraze/aideck-gap8-examples/pull/156 this pull request adds an example to stream the state to the NINA module, such that it can be displayed and logged together with the camera images.


gemenerik (approved):

gemenerik:
We may choose to move this elsewhere as part of broader reorganizing, but good to merge

PR #1637, opened 2026-05-29
Increase bcCam firmware reset timing

ataffanel:
## Summary

The QCC748 firmware boot path needs a little more time after reset before the UART link is used. This PR replaces the short inline waits with named timing constants and increases the firmware boot wait to make reset-to-firmware more reliable.

## Changes

- Add named timing constants for the bcCam firmware reset sequence
- Increase QCC reset hold/setup waits from 10 ms to 50 ms
- Increase firmware boot wait from 100 ms to 500 ms
- Increase RX drain after firmware boot from 50 ms to 100 ms

## Test Plan

- Verified with git diff --check against origin/master
- Hardware validation: reset bcCam/QCC748 back to firmware mode and confirm it reliably boots/responds

### DIFF
--- a/src/deck/drivers/src/bccam_deck.c
+++ b/src/deck/drivers/src/bccam_deck.c
@@ -45,6 +45,11 @@
 #define BCCAM_UART_BAUDRATE       2000000
 #define BCCAM_ISP_BAUDRATE        500000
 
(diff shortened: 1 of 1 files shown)


evoggy (approved):
Looks good!

gemenerik (approved):

PR #1639, opened 2026-06-03
Add a deck supervisor for runtime deck health monitoring

evoggy:
Adds a **deck supervisor** that lets a deck expose a runtime hardware status
probe, and (optionally) acts on a failing deck through the system supervisor —
blocking arming and cutting motors in flight. Driven by the deck framework and
gated by a Kconfig switch.

## Mechanism
- `DeckDriver` gains `uint8_t (*status)(void)` — returns **0 when healthy, a
  non-zero error code otherwise**; `NULL` means no probe. The deck supervisor
  pulls the pointer from the deck registry (`deckCount()`/`deckInfo()`) and
  calls it; decks don't register themselves. The specific error code is ignored
  for now (any non-zero = fault), leaving room for richer codes later.
- A deck that failed to init returns non-zero immediately (its state can't be
  trusted to probe). Decks without a probe are skipped.

## Probes
- **Lighthouse** (`lighthouseCoreDeckStatus`): the FPGA emits a sync frame ~1 Hz
  even with no basestations, so status is 0 while a frame arrived within 1500 ms.
  The frame-age check reads the timestamp before `now` and uses an unsigned diff
  (underflow- and systick-wrap-safe).
- **Loco** (`locoDeckStatus`): after 2000 ms of UWB silence, re-reads the
  `CHAN_CTRL` config register and compares to the value captured at init —
  catches a power-glitched chip that still answers `DEV_ID` but lost its config;
  "no anchors" stays healthy. Bounded `uwbTask` wait so the probe runs in TDoA2
  too (its `onEvent` returns `portMAX_DELAY`).

## Logging — always available
Per-deck status is published in a `deckStatus` log group: `deckStatus.bcLoco` /
`deckStatus.bcLighthouse4`, **0 = ok, non-zero = error** (mirrors the probe).
Independent of the Kconfig switch, so health is always observable from an app or
the client. (Note: this is the inverse polarity of an earlier "alive" iteration.)

## Propagation switch
`CONFIG_DECK_SUPERVISOR` (default y) gates whether a deck fault reaches the
system supervisor:
- **On:** raises `SUPERVISOR_CB_DECK_FAULT` → blocks arming, and from `Flying`
  *or* `WarningLevelOut` goes to `ExceptFreeFall → Locked` (motor cut).
- **Off:** the supervisor ignores deck faults; handle them manually via the
  `deckStatus.*` log variables.

## Supervisor
New condition `SUPERVISOR_CB_DECK_FAULT` blocks the pre-flight→ready transitions
and triggers the in-flight cut (from both `Flying` and `WarningLevelOut`).
`supervisor.info` bit 11 = deck fault.


src/modules/src/supervisor_state_machine.c
ataffanel: Should we also add a state transition when `SUPERVISOR_CB_DECK_FAULT` triggers when we are leveling as well? Even though leveling is supposed to be only temporary, it is alrealy stopped by all the other fault that would cause the motor to stop so I think this one should also be added to the list.

src/modules/src/lighthouse/lighthouse_core.c
```diff
@@ -136,7 +141,28 @@ static pulseProcessorProcessPulse_t pulseProcessorProcessPulse = pulseProcessorV
...
+  // If the deck never flashed/booted we can't trust its state to probe it
+  if (!deckIsFlashed) {
+    return false;
+  }
+
+  const uint32_t now = T2M(xTaskGetTickCount());
+  const uint32_t lastFrame = lastFrameTs;
+  // A concurrent update can make lastFrame "newer" than now; that means a frame
+  // just arrived, so treat as alive (and avoid the unsigned subtraction underflowing).
+  if (now < lastFrame) {
```
ataffanel: Very minor: then systick wrap around, for one step, this will always be true. Will quickly resolve so might not require a wrap-protected difference though.

ataffanel (changes requested):
We are missing a transition from `WarningLevelOut`. Otherwise this looks good!

src/deck/interface/deck_core.h
```diff
@@ -113,6 +113,23 @@ typedef struct deck_driver {
...
+   * mutexes, no slow bus transactions in the call itself - cache results from
+   * the deck task instead).
+   *
+   * IMPORTANT: when CONFIG_DECK_SUPERVISOR is enabled a false return blocks
+   * arming and, while flying, immediately cuts the motors. The result must
+   * therefore be debounced in the driver - a single transient false reading
+   * will trigger the cut, so do not report short, recoverable glitches as not
+   * alive. Return false if the deck failed to initialize, since its state
+   * cannot be trusted to be probed. */
+  bool (*isAlive)(void);
```
tobbeanton: Would it be interesting to be able to return more error codes?

ataffanel (dismissed):
All good!

tobbeanton (approved):
Looks good!

PR #1640, opened 2026-06-03
Add params to toggle LH and Loco measurements into the estimator

evoggy:
## Summary

Adds two runtime parameters to enable/disable feeding lighthouse and Loco
positioning measurements into the state estimator (Kalman filter), without
otherwise shutting down those subsystems.

- `lighthouse.enableEst` — gates lighthouse crossing-beam position, sweep-angle, and yaw-error measurements
- `loco.enableEst` — gates Loco TWR/TDoA distance, TDoA, and absolute-height measurements

Both default to `1` (enabled, so behavior is unchanged out of the box) and are
`PARAM_PERSISTENT`, so a chosen setting survives reboot.

## Motivation

Allows comparing/combining positioning sources at runtime — e.g. fly on
lighthouse while keeping the Loco deck powered for logging, or temporarily cut
one source to diagnose estimator behavior — without reflashing or unplugging
hardware.

## Implementation

The measurements are gated at the source, right before each
`estimatorEnqueue*()` call, mirroring the existing compile-time
`CONFIG_DECK_LIGHTHOUSE_AS_GROUNDTRUTH` pattern:

- **Lighthouse** (`lighthouse_position_est.c`): new file-static `enableEstimator`
  guards the position, sweep-angle (×4) and yaw-error enqueues.
- **Loco** (`locodeck.c` + `locodeck.h`): new global `locoEnableEstimator`,
  declared `extern` in the header and checked by the three ranging algorithms
  (`lpsTwrTag`, `lpsTdoa2Tag`, `lpsTdoa3Tag`).

### DIFF
--- a/src/deck/drivers/src/locodeck.c
+++ b/src/deck/drivers/src/locodeck.c
@@ -126,6 +126,9 @@ static uwbAlgorithm_t *algorithm = &uwbTwrTagAlgorithm;
 #endif
 
 static bool isInit = false;
+
+uint8_t locoEnableEstimator = 1;
+
 static TaskHandle_t uwbTaskHandle = 0;
 static SemaphoreHandle_t algoSemaphore;
 
@@ -740,5 +743,14 @@ PARAM_GROUP_START(loco)
  * |   3   | TDoA 3 |\n
  */
 PARAM_ADD_CORE(PARAM_UINT8, mode, &algoOptions.userRequestedMode)
+
+/**
+ * @brief Feed loco measurements into the state estimator (default: 1)
+ *
+ * Set to 0 to stop pushing loco (TWR/TDoA distance, TDoA and absolute height)
+ * measurements into the Kalman filter, while keeping the loco system otherwise
+ * running.
+ */
+PARAM_ADD_CORE(PARAM_UINT8 | PARAM_PERSISTENT, enableEst, &locoEnableEstimator)
 
 PARAM_GROUP_STOP(loco)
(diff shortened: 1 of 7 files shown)


ataffanel (dismissed):

src/deck/drivers/src/locodeck.c
```diff
@@ -741,4 +744,13 @@ PARAM_GROUP_START(loco)
...
 PARAM_ADD_CORE(PARAM_UINT8, mode, &algoOptions.userRequestedMode)
 
+/**
+ * @brief Feed loco measurements into the state estimator (default: 1)
+ *
+ * Set to 0 to stop pushing loco (TWR/TDoA distance, TDoA and absolute height)
+ * measurements into the Kalman filter, while keeping the loco system otherwise
+ * running.
+ */
+PARAM_ADD_CORE(PARAM_UINT8 | PARAM_PERSISTENT, enableEst, &locoEnableEstimator)
```
tobbeanton: Should we have it persistent? Can't come up with a use case that requires this.

src/deck/drivers/src/locodeck.c
```diff
@@ -126,6 +126,9 @@ static uwbAlgorithm_t *algorithm = &uwbTwrTagAlgorithm;
 #endif
 
 static bool isInit = false;
+
+uint8_t locoEnableEstimator = 1;
```
tobbeanton, line 130: What about calling everything enableToEstimator?

tobbeanton (changes requested):

src/deck/drivers/src/locodeck.c
evoggy: What about fwdToEstimator?

src/deck/drivers/src/locodeck.c
evoggy: No, I don't really have a strong use case for this so let's remove it.

tobbeanton (approved):
Looks good!

PR #1642, opened 2026-06-09
app_hello_rs: Make rust console print run within loop and update docs

stefanthorstenson:
- Put the console print within the loop
- Add docs on how to run the example


ataffanel (approved):
Looks good!

PR #1643, opened 2026-06-16
Fix typo in Kalman filter initial yaw std configuration

ntamas:
This PR fixes a typo in the name of the macro that was supposed to contain the initial yaw std of the Kalman filter. The current code in `master` looks at a non-existent config variable and thus completely ignores the configured inital yaw std from Kconfig.


stefanthorstenson:
Hi @ntamas , and thanks for the PR! Nice catch! I will look into this; I just want to confirm that everything works as it should on my side.

stefanthorstenson (approved):
Looks good! Thanks for the contribution!

PR #1644, opened 2026-06-16
Added simple crash detection to supervisor

tobbeanton:
This pull request introduces a configurable crash detection mechanism based on acceleration measurements to the supervisor module. The main changes add a new configuration option for crash detection sensitivity, update the supervisor logic to use this setting, and expose relevant parameters and logging for monitoring and tuning.

**Crash Detection Feature:**

* Added a new configuration option `SUPERVISOR_CRASH_DETECTION_G` in `Kconfig` to set the maximum allowed deviation from 1 G for crash/tumble detection. Setting this to 0 disables the feature.
* Implemented crash detection logic in `isTumbledCheck` in `supervisor.c`, which calculates the acceleration norm and triggers a crash if the deviation from 1 G exceeds the configured threshold.

**Parameter and Logging Integration:**

* Exposed the crash detection threshold as a persistent parameter `crashDetectGs` for runtime tuning.
* Added logging for the current acceleration norm (`accNorm`) used in crash/tumble detection, enabling monitoring via the logging subsystem.
* Declared and initialized variables for the crash detection threshold and acceleration norm in `supervisor.c`.

### DIFF
--- a/src/modules/src/Kconfig
+++ b/src/modules/src/Kconfig
@@ -293,6 +293,18 @@ config LOG_MOTOR_CAP_WARNING
 
 endmenu
 
+menu "Supervisor configuration"
+
+config SUPERVISOR_CRASH_DETECTION_G
+    int "Crash detection (Gs)"
+    range 0 20
+    default 0
+    help
+        Maximum allowed deviation from 1 G when evaluating tumble/crash
+        detection. A value of 0 disables the detection.
+
+endmenu
+
 menu "Parameter subsystem"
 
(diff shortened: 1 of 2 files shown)


ArisMorgens:
Currently, the check will return true if a single acceleration sample exceeds the limits. Would it make sense to require the deviation to persist for some milliseconds before triggering?

ArisMorgens (approved):
One small comment/question. Otherwise it looks really good!

tobbeanton:
The accelerometer is heavily filtered so one sample is enough.

PR #1645, opened 2026-06-26
Reset pid controller when switching commander

stefanthorstenson:
# Issue Summary: PID Integral Corruption on Commander Handover

## Background

@ShuqinZ, among others, have sometimes noticed a large transient when switching from the low-level commander (LLC) to the high-level commander (HLC). Furthermore, @ShuqinZ discovered and verified that resetting the integral parts of the PID controllers when changing to HLC solved the issue. See #1627 for more information.

## Issue Analysis

The Crazyflie PID position controller runs two cascaded loops at 100 Hz: an outer position PID producing a velocity setpoint, and an inner velocity PID producing attitude/thrust commands. When the high-level commander (HLC) is active, both loops run and their outputs are used. When the low-level commander (LLC) is active (e.g. raw RPYT packets), the velocity PID outputs are silently overridden by the raw setpoint values — but the PIDs still run and their integrals keep accumulating.

The attitude controller runs at 500 Hz.

The velocity PIDs (`pidVX`, `pidVY`, `pidVZ`) run unconditionally every tick regardless of the active setpoint mode. When LLC sends raw attitude commands, the mode flags are `modeDisable`, causing the velocity PID outputs to be discarded (`controller_pid.c:99–105`). Meanwhile the PIDs accumulate error against actual body velocity, which has no reason to match the zero velocity setpoint. With `ki_Z = 15.0`, even one second of 0.15 m/s vertical motion adds 2.25 to `VZi` immediately on resuming position control.

When the LLC stops and the HLC resumes, the corrupted integrals produce a large, sudden attitude/thrust correction — causing a visible jitter or lurch at the transition. This can happen in x, y or z direction.

A timing subtlety: if there is a pause between `notify_setpoint_stop` and the first HLC command, the stopped planner issues a `nullSetpoint` (all modes `modeDisable`, thrust = 0), which triggers the existing `thrust == 0` safety block and resets all PIDs. So the corruption is only visible when the `go_to` command arrives fast enough to prevent that reset — specifically with `--no-pause-before-hlc`.

## Proposed Solution

Reset all position and attitude PIDs whenever the setpoint mode changes. The reason for resetting all controllers on any mode change (compared to supervising each controller and its mode) is that many of the controllers are coupled. This means that a change in a mode for one controller would need to reset another controller. Resetting everything is a simple solution, and a good start. If customers experience issues due to this strategy, we can do an overhaul.

A static `prevSetpoint.mode` is compared each tick at the top of `controllerPid`. On any change, `positionControllerResetAllPID` and `attitudeControllerResetAllPID` are called, and `attitudeDesired.roll/pitch` is snapped to the current state so the attitude controller holds level during the up-to-10 ms gap before the next 100 Hz position controller tick. The existing per-tick `modeVelocity` attitude resets are removed since the mode-change reset subsumes them.

This approach is decoupled from the commander: the controller detects the transition itself from the setpoint it receives, making behaviour predictable regardless of timing between `notify_setpoint_stop` and the first HLC command.

## Results

The tests try to reproduce and demonstrate the results of the fix, using the Z axis as example. The Z has a larger coefficient for the I part, which means less time spent using the LLC to visualize the effect.

Each test runs three phases:

- **Phase 1** — HLC hover settle: the drone takes off and holds position under HLC control.
- **Phase 2** — LLC takes over: integrals build up for controllers which output is unused.
- **Phase 3** — HLC switch-back: `notify_setpoint_stop` is called and HLC resumes with a `go_to` command to the current position. The effect of the corrupted integrals is observable here.

Three scenarios have been run:

- **Scenario A** — LLC takes over at altitude 1.5 m and descends to around 0.5 m, where command is handed over to HLC. The drone is still descending at handover.
- **Scenario B** — LLC only hovers.
- **Scenario C** — LLC takes over at altitude 1.5 m and descends to around 0.5 m, and reaches hovering before command is handed over to HLC.

#### Before fix — integral build up with no reset

Velocity Z integral `posCtl.VZi` builds up during the LLC in phase 2 and corrupts the first position-control thrust command, causing a visible altitude lurch and an audible thrust boost when entering phase 3 (HLC).

image

#### After fix — PID reset at mode switch

On mode change the integrals are reset. The transition is smooth. There is a substantial undershoot though, due to integrals being reset while the the drone is moving.

image

#### After fix

In this setup, the LLC is hovering, which gives the cleanest handover to HLC (phase 3).

There is a feedforward term, `thrustBase`, which should represent the thrust needed to hover. This does most of the initial compensation. This also means that a handover in hover state should be smooth. In the case of the Brushless, which this is, `thrustBase` is about 10 % lower than needed, which gives the initial dip. The dip is a lot less than in the dynamic handover case above, however.

image

### Scenario C

In this example, integrals build up during LLC descent. The LLC then tries to reach hover (z velocity close to 0), before handing over to HLC. 

#### Before fix

image

#### After fix

image

#### Comparison

The image is zoomed in; only last part of phase 2 and the first part of phase 3 are seen.

To the left: Before the fix. This creates a control bump of approximately 20 cm.

To the right: After the fix. Undershoot of about 7 cm.

image

# Conclusions

Resetting all PID controllers is a good place to start. This should solve the problems for all axes, and for all cases when switching commander. Even though the steady-state handover in the simplest case going from HLC to LLC to HLC is probably better with the old strategy (since the I part should be exactly correct after HLC has found it after its first run), the dip seen with the new strategy is acceptable and avoids the worst cases. Also, the new strategy gives an honest and clear state of the controllers, and make it easier to improve if needed.

### DIFF
--- a/src/modules/src/controller/controller_pid.c
+++ b/src/modules/src/controller/controller_pid.c
@@ -40,6 +40,15 @@ bool controllerPidTest(void)
   return pass;
 }
 
+static void controllerPidReinitialize(const state_t *state)
+{
+  positionControllerResetAllPID(state->position.x, state->position.y, state->position.z);
+  positionControllerResetAllfilters();
+  attitudeControllerResetAllPID(state->attitude.roll, state->attitude.pitch, state->attitude.yaw);
+  attitudeDesired.roll  = state->attitude.roll;
+  attitudeDesired.pitch = state->attitude.pitch;
+}
+
 static float capAngle(float angle) {
   float result = angle;
 
@@ -52,15 +61,6 @@ static float capAngle(float angle) {
   }
 
   return result;
-}
-
-static void controllerPidReinitialize(const state_t *state)
-{
-  positionControllerResetAllPID(state->position.x, state->position.y, state->position.z);
-  positionControllerResetAllfilters();
-  attitudeControllerResetAllPID(state->attitude.roll, state->attitude.pitch, state->attitude.yaw);
-  attitudeDesired.roll  = state->attitude.roll;
-  attitudeDesired.pitch = state->attitude.pitch;
 }
 
 static bool setpointModeChanged(const setpoint_t *setpoint)
--- a/src/modules/src/controller/controller_pid.c
+++ b/src/modules/src/controller/controller_pid.c
@@ -79,7 +79,7 @@ void controllerPid(control_t *control, const setpoint_t *setpoint,
   control->controlMode = controlModeLegacy;
 
   if (setpointModeChanged(setpoint)) {
-    controllerPidReinitialize(state);
+    controllerPidReinitialize(state); // To prevent control bump
   }
 
   if (RATE_DO_EXECUTE(ATTITUDE_RATE, stabilizerStep)) {
--- a/src/modules/src/controller/controller_pid.c
+++ b/src/modules/src/controller/controller_pid.c
@@ -57,6 +57,7 @@ static float capAngle(float angle) {
 static void controllerResetAllPids(const state_t *state)
 {
   positionControllerResetAllPID(state->position.x, state->position.y, state->position.z);
+  positionControllerResetAllfilters();
   attitudeControllerResetAllPID(state->attitude.roll, state->attitude.pitch, state->attitude.yaw);
   attitudeDesired.roll  = state->attitude.roll;
   attitudeDesired.pitch = state->attitude.pitch;
--- a/src/modules/src/controller/controller_pid.c
+++ b/src/modules/src/controller/controller_pid.c
@@ -1,3 +1,4 @@
+#include <string.h>
 
 #include "stabilizer_types.h"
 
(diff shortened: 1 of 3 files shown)


src/modules/src/controller/controller_pid.c
```diff
@@ -53,13 +54,34 @@ static float capAngle(float angle) {
...
+  positionControllerResetAllPID(state->position.x, state->position.y, state->position.z);
+  positionControllerResetAllfilters();
+  attitudeControllerResetAllPID(state->attitude.roll, state->attitude.pitch, state->attitude.yaw);
+  attitudeDesired.roll  = state->attitude.roll;
+  attitudeDesired.pitch = state->attitude.pitch;
+}
+
+static bool setpointModeChanged(const setpoint_t *setpoint)
+{
+  static setpoint_t prev; // only .mode is used; zero-init triggers reset on first tick
```
gemenerik: It would be cheaper on static RAM to only store .mode

src/modules/src/controller/controller_pid.c
```diff
@@ -53,13 +54,34 @@ static float capAngle(float angle) {
   return result;
 }
 
+static void controllerResetAllPids(const state_t *state)
```
gemenerik: Name could potentially be better, does more than just reset PIDs

gemenerik (commented):
Resetting on any mode change seems fair. Approach looks fair. Would have been nice to have the boundedness demonstrated in a worse scenario (fast motion, big integral), but theoretically this is sound.

src/modules/src/controller/controller_pid.c
stefanthorstenson: Good catch! Seems reasonable to only store what is needed.

stefanthorstenson: Fixed. I created a named typedef for the anonymous struct that was the setpoint mode.

src/modules/src/controller/controller_pid.c
stefanthorstenson: Fixed.

stefanthorstenson:
> Resetting on any mode change seems fair. Approach looks fair. Would have been nice to have the boundedness demonstrated in a worse scenario (fast motion, big integral), but theoretically this is sound.

Updated PR description with Scenario C.

gemenerik (approved):
Nice work!

ArisMorgens (approved):
Tested it and it looks good!

PR #1647, opened 2026-06-30
Fix FP32 to FP16 conversion

nellbru:
`single2half()` joined the exponent and rounded mantissa with `|`. However, when rounding overflows the mantissa to `0x400`, the carry bit lands on the exponent's lowest bit. If that bit is `1`, the `|` operator drops the carry bit, resulting in a value 2× too small (which showed up as position spikes in my telemetry logs). 

This only happens for values where the upper 11 bits of the mantissa of the FP32 are `1` and where the exponent is odd.

**Fix:**
Use `+` instead of `|` so the carry bit properly propagates to the exponent.

Before the fix:
`single2half(0.49999f)` -> `0.25`

After the fix:
`single2half(0.49999f)` -> `0.5`

### DIFF
--- a/src/utils/src/num.c
+++ b/src/utils/src/num.c
@@ -58,7 +58,7 @@ uint16_t single2half(float number)
     if (e<(127-15))
         return 0; //Do not handle generating subnormalised representation
 
-    return (s<<15) | ((e-127+15)<<10) + (((num>>13)&0x3FF)+((num>>12)&0x01));
+    return (s<<15) | (((e-127+15)<<10) + (((num>>13)&0x3FF)+((num>>12)&0x01)));
 }
 
 float half2single(uint16_t number)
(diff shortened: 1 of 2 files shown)


nellbru:
Here are plotted position logs (values logged as FP16) when making one drone fly at an altitude of 0.5 m, before and after applying the fix:

#### Before applying the new firmware

Before flash

#### After applying the new firmware

After flash

Note that in both cases the drone started at a position around (-0.6, 0.6, 0) to go to (0, 0, 0.5).

stefanthorstenson:
Nice catch! Thanks for finding it and fixing it. I added some unit tests to confirm your finding:
`testThatSingle2halfHandlesRoundingCarryIntoExponent`
Pre-fix: fails
Post-fix: passes

ataffanel:
Nice catch indeed! Looks good to merge for me.

stefanthorstenson (approved):

stefanthorstenson:
Merging! Thanks a lot!

PR #1650, opened 2026-07-03
Improve Loco TDoA3 pairing

stefanthorstenson:
- Reject loco TDoA3 measurement pairs with bad geometry, based on distance difference ratio.
- Since matchRandomAnchor now calculates the distance difference in order to determine if geometry is good, output that calculated value to only calculate it once.

See comments below for verification and explanation of the chosen algorithm.

### DIFF
--- a/src/modules/src/estimator/estimator_kalman.c
+++ b/src/modules/src/estimator/estimator_kalman.c
@@ -524,6 +524,8 @@ LOG_GROUP_STOP(kalman)
 
 LOG_GROUP_START(outlierf)
   LOG_ADD(LOG_INT32, lhWin, &sweepOutlierFilterState.openingWindowMs)
+  LOG_ADD(LOG_FLOAT, tdoaIntegrator, &outlierFilterTdoaState.integrator)
+  LOG_ADD(LOG_UINT8, tdoaFilterOpen, &outlierFilterTdoaState.isFilterOpen)
 LOG_GROUP_STOP(outlierf)
 
 /**
--- a/src/modules/src/estimator/estimator_kalman.c
+++ b/src/modules/src/estimator/estimator_kalman.c
@@ -524,8 +524,6 @@ LOG_GROUP_STOP(kalman)
 
 LOG_GROUP_START(outlierf)
   LOG_ADD(LOG_INT32, lhWin, &sweepOutlierFilterState.openingWindowMs)
-  LOG_ADD(LOG_FLOAT, tdoaIntegrator, &outlierFilterTdoaState.integrator)
-  LOG_ADD(LOG_UINT8, tdoaFilterOpen, &outlierFilterTdoaState.isFilterOpen)
 LOG_GROUP_STOP(outlierf)
 
 /**
--- a/src/modules/src/kalman_core/mm_tdoa.c
+++ b/src/modules/src/kalman_core/mm_tdoa.c
@@ -25,6 +25,9 @@
 
 #include "mm_tdoa.h"
 #include "test_support.h"
+#include "log.h"
+
+static float tdoaInnovation = 0.0f;
 
 #if CONFIG_ESTIMATOR_KALMAN_TDOA_OUTLIERFILTER_FALLBACK
 #include "outlierFilterTdoaSteps.h"
@@ -60,6 +63,7 @@ void kalmanCoreUpdateWithTdoa(kalmanCoreData_t* this, tdoaMeasurement_t *tdoa, c
 
   float predicted = d1 - d0;
   float error = measurement - predicted;
+  tdoaInnovation = error;
 
   float h[KC_STATE_DIM] = {0};
   arm_matrix_instance_f32 H = {1, KC_STATE_DIM, h};
@@ -92,3 +96,7 @@ void kalmanCoreUpdateWithTdoa(kalmanCoreData_t* this, tdoaMeasurement_t *tdoa, c
     }
   }
 }
+
+LOG_GROUP_START(tdoa)
+  LOG_ADD(LOG_FLOAT, innovation, &tdoaInnovation)
+LOG_GROUP_STOP(tdoa)
--- a/src/modules/src/kalman_core/mm_tdoa.c
+++ b/src/modules/src/kalman_core/mm_tdoa.c
@@ -25,9 +25,6 @@
 
 #include "mm_tdoa.h"
 #include "test_support.h"
-#include "log.h"
-
-static float tdoaInnovation = 0.0f;
 
 #if CONFIG_ESTIMATOR_KALMAN_TDOA_OUTLIERFILTER_FALLBACK
 #include "outlierFilterTdoaSteps.h"
@@ -63,7 +60,6 @@ void kalmanCoreUpdateWithTdoa(kalmanCoreData_t* this, tdoaMeasurement_t *tdoa, c
 
   float predicted = d1 - d0;
   float error = measurement - predicted;
-  tdoaInnovation = error;
 
   float h[KC_STATE_DIM] = {0};
   arm_matrix_instance_f32 H = {1, KC_STATE_DIM, h};
@@ -96,7 +92,3 @@ void kalmanCoreUpdateWithTdoa(kalmanCoreData_t* this, tdoaMeasurement_t *tdoa, c
     }
   }
 }
-
-LOG_GROUP_START(tdoa)
-  LOG_ADD(LOG_FLOAT, innovation, &tdoaInnovation)
-LOG_GROUP_STOP(tdoa)
--- a/src/modules/src/lighthouse/lighthouse_throttle.c
+++ b/src/modules/src/lighthouse/lighthouse_throttle.c
@@ -33,7 +33,7 @@
 #include "log.h"
 
 static const uint32_t evaluationIntervalMs = 100;
-static uint16_t maxRate = 50;  // Samples / second
+static uint16_t maxRate = 10;  // Samples / second
 static float discardProbability = 0.0f;
 
 bool throttleLh2Samples(const uint32_t nowMs) {
--- a/src/modules/src/lighthouse/lighthouse_throttle.c
+++ b/src/modules/src/lighthouse/lighthouse_throttle.c
@@ -33,7 +33,7 @@
 #include "log.h"
 
 static const uint32_t evaluationIntervalMs = 100;
-static uint16_t maxRate = 10;  // Samples / second
+static uint16_t maxRate = 50;  // Samples / second
 static float discardProbability = 0.0f;
 
 bool throttleLh2Samples(const uint32_t nowMs) {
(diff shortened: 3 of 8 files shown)


stefanthorstenson:
# Verification

"Fix FW" is dcd2a97c496606d1779e24d4b5f5faf1518671b5. This commit also contains a setup for using Lighthouse as a reference system, and added logging. These additions have since been removed.

Anchors according to `ff-geo.yaml`:

```yaml
0:
  x: 1.6800000000000002
  y: 0.02

  z: 0.0
1:
  x: 0.0
  y: 0.0
  z: 0.0
2:
  x: 0.0
  y: 1.12
  z: 0.0
3:
  x: -1.0
  y: 0.0
  z: 0.0
10:
  x: 1.68
  y: -0.33
  z: 1.67
```

Crazyflie stationary in (0.88, 0, 0.20).

This gives a measurement-to-baseline for anchor pair 1 and 3 a ratio of 0.99.

image

## Setup performance test

Anchor pair 1 and 3 should affect X position estimation.

Runs:

1. Anchors 0,1,2,10. Release FW.
2. Anchors 0,1,2,3,10. Release FW.
3. Anchors 0,1,2,3,10. Fix FW.

Compare:
Standard deviation for X position.
Maximum error for X position.

### Results

True X position: 0.88 m.

Data source: `Run2-60s` logs (60 s per case).

| Case | Anchors | FW      | N    | Mean X (m) | Std dev X (m) | Max error X (m) |
|------|---------|---------|------|------------|---------------|------------------|
| 1    | 0,1,2,10    | Release | 1530 | 0.6833     | 0.0972        | 0.8044           |
| 2    | 0,1,2,3,10  | Release | 1538 | 0.6983     | 0.0558        | 0.3413           |
| 3    | 0,1,2,3,10  | Fix     | 1505 | 0.6799     | 0.0844        | 0.4532           |

### Summary

No obvious increase in performance. Rather the contrary.

## Setup functional test

1. Anchors 0,1,2,10.
2. Anchors 1,3.
3. Anchors 0,1,2,3,10.

Log `tdoaEngine.stGeometryTEST`, which shows how many measurements have been discarded
due to geometry when matching pairs.

### Results

#### Case 1

```
LogData { timestamp: 2075131, data: {"tdoaEngine.stGeometryTEST": F32(0.0)} }
LogData { timestamp: 2075231, data: {"tdoaEngine.stGeometryTEST": F32(0.0)} }
LogData { timestamp: 2075331, data: {"tdoaEngine.stGeometryTEST": F32(0.0)} }
LogData { timestamp: 2075431, data: {"tdoaEngine.stGeometryTEST": F32(0.0)} }
LogData { timestamp: 2075531, data: {"tdoaEngine.stGeometryTEST": F32(0.0)} }
LogData { timestamp: 2075631, data: {"tdoaEngine.stGeometryTEST": F32(0.0)} }
LogData { timestamp: 2075731, data: {"tdoaEngine.stGeometryTEST": F32(1.6666666)} }
LogData { timestamp: 2075831, data: {"tdoaEngine.stGeometryTEST": F32(1.6666666)} }
LogData { timestamp: 2075931, data: {"tdoaEngine.stGeometryTEST": F32(1.6666666)} }
LogData { timestamp: 2076031, data: {"tdoaEngine.stGeometryTEST": F32(1.6666666)} }
LogData { timestamp: 2076131, data: {"tdoaEngine.stGeometryTEST": F32(1.6666666)} }
LogData { timestamp: 2076231, data: {"tdoaEngine.stGeometryTEST": F32(1.6666666)} }
LogData { timestamp: 2076331, data: {"tdoaEngine.stGeometryTEST": F32(0.0)} }
LogData { timestamp: 2076431, data: {"tdoaEngine.stGeometryTEST": F32(0.0)} }
LogData { timestamp: 2076531, data: {"tdoaEngine.stGeometryTEST": F32(0.0)} }
LogData { timestamp: 2076631, data: {"tdoaEngine.stGeometryTEST": F32(0.0)} }
LogData { timestamp: 2076731, data: {"tdoaEngine.stGeometryTEST": F32(0.0)} }
LogData { timestamp: 2076831, data: {"tdoaEngine.stGeometryTEST": F32(0.0)} }
LogData { timestamp: 2076931, data: {"tdoaEngine.stGeometryTEST": F32(1.6666666)} }
```

#### Case 2

```
LogData { timestamp: 1809323, data: {"tdoaEngine.stGeometryTEST": F32(20.0)} }
LogData { timestamp: 1809423, data: {"tdoaEngine.stGeometryTEST": F32(20.0)} }
LogData { timestamp: 1809523, data: {"tdoaEngine.stGeometryTEST": F32(38.333332)} }
LogData { timestamp: 1809623, data: {"tdoaEngine.stGeometryTEST": F32(38.333332)} }
LogData { timestamp: 1809723, data: {"tdoaEngine.stGeometryTEST": F32(38.333332)} }
LogData { timestamp: 1809823, data: {"tdoaEngine.stGeometryTEST": F32(38.333332)} }
LogData { timestamp: 1809923, data: {"tdoaEngine.stGeometryTEST": F32(38.333332)} }
LogData { timestamp: 1810023, data: {"tdoaEngine.stGeometryTEST": F32(38.333332)} }
LogData { timestamp: 1810123, data: {"tdoaEngine.stGeometryTEST": F32(33.333332)} }
LogData { timestamp: 1810223, data: {"tdoaEngine.stGeometryTEST": F32(33.333332)} }
LogData { timestamp: 1810323, data: {"tdoaEngine.stGeometryTEST": F32(33.333332)} }
LogData { timestamp: 1810423, data: {"tdoaEngine.stGeometryTEST": F32(33.333332)} }
LogData { timestamp: 1810523, data: {"tdoaEngine.stGeometryTEST": F32(33.333332)} }
LogData { timestamp: 1810623, data: {"tdoaEngine.stGeometryTEST": F32(33.333332)} }
LogData { timestamp: 1810723, data: {"tdoaEngine.stGeometryTEST": F32(14.999999)} }
LogData { timestamp: 1810823, data: {"tdoaEngine.stGeometryTEST": F32(14.999999)} }
```

#### Case 3

```
LogData { timestamp: 2146777, data: {"tdoaEngine.stGeometryTEST": F32(0.0)} }
LogData { timestamp: 2146877, data: {"tdoaEngine.stGeometryTEST": F32(0.0)} }
LogData { timestamp: 2146977, data: {"tdoaEngine.stGeometryTEST": F32(0.0)} }
LogData { timestamp: 2147077, data: {"tdoaEngine.stGeometryTEST": F32(0.0)} }
LogData { timestamp: 2147177, data: {"tdoaEngine.stGeometryTEST": F32(0.0)} }
LogData { timestamp: 2147277, data: {"tdoaEngine.stGeometryTEST": F32(3.3333333)} }
LogData { timestamp: 2147377, data: {"tdoaEngine.stGeometryTEST": F32(3.3333333)} }
LogData { timestamp: 2147477, data: {"tdoaEngine.stGeometryTEST": F32(3.3333333)} }
LogData { timestamp: 2147577, data: {"tdoaEngine.stGeometryTEST": F32(3.3333333)} }
LogData { timestamp: 2147677, data: {"tdoaEngine.stGeometryTEST": F32(3.3333333)} }
LogData { timestamp: 2147777, data: {"tdoaEngine.stGeometryTEST": F32(3.3333333)} }
LogData { timestamp: 2147877, data: {"tdoaEngine.stGeometryTEST": F32(0.0)} }
LogData { timestamp: 2147977, data: {"tdoaEngine.stGeometryTEST": F32(0.0)} }
LogData { timestamp: 2148077, data: {"tdoaEngine.stGeometryTEST": F32(0.0)} }
LogData { timestamp: 2148177, data: {"tdoaEngine.stGeometryTEST": F32(0.0)} }
```

#### Summary

Case 1 and 3 are quite similar. No obvious difference.

For case 2, it is clear that the function works as intended.

It would be worth investigating why the 1,3 pair shows up so rarely.

stefanthorstenson:
# Sensitivity approximation

The measurement is h(p) = |p - A1| - |p - A2|. Its gradient with respect to the tag position p is
grad(h) = u1 - u2, where u1, u2 are the unit vectors from each anchor towards the tag (i.e.
the lines of sight). The true measurement sensitivity is:
|grad(h)| = |u1 - u2| = 2 * sin(theta / 2)
with theta the angle between those two lines of sight (subtended angle).

The measurement sensitivity is a relative representation of how sensitive a measurement it to noise, where 1 is best and 0 is worst. As an example, a geometry with sensitivity 0.5 will be affected twice as much by the same noise as a geometry with sensitivity 1.

An example of a geometry which shows the subtended angle and the measurement sensitivity:

image

The subtended angle to measurement sensitivity graph:

image

The true measurement sensitivity can only be calculated if the target position is known. Position is only available downstream from pairing in the software stack. At pairing, the distance difference ratio can be used as an approximation.

Below are a few examples of how the distance difference ratio approximates the true sensitivity. A sensitivity of 0.3 is used in the plots as an example to get a feeling of how the distance difference ratio will filter out good pairs.

The implemented algorithm will discard all pairs with a higher distance difference ratio than the limit (contrary to the measurement sensitivity, where a high number is good).

Note that the distance difference ratio cannot catch bad geometry due to large distance away from the anchors.

## The suggested value (0.85)

image

Note that no analysis of why 0.85 would be a good value has been done.

## The suggested lower boundary (0.60)

image

## The suggested upper boundary (0.98)

image

gemenerik (commented):
In a typical 8-node volume the only way to trigger this, assuming you stay within the volume enclosed by said anchors, is by flying close to one of the anchors (the rejection region wraps each anchor). This is precisely where the sensitivity is above average. This filter would then reject this above-average pair in favor of a likely more average (and thus worse) one. 

For any set-up where you can expect to fly somewhere where two nodes line up on one side of the drone (typical for a >8 node volume), I can definitely see this helping.

Should we add some way to accept at least one pair? Right now it can theoretically reject all pairs.

I think this needs more evidence before it lives on master. Perhaps through comparing the same log through bindings with this enabled/disabled? What about merging to experimental-show in the meantime?

stefanthorstenson:
> In a typical 8-node volume the only way to trigger this, assuming you stay within the volume enclosed by said anchors, is by flying close to one of the anchors (the rejection region wraps each anchor). This is precisely where the sensitivity is above average. This filter would then reject this above-average pair in favor of a likely more average (and thus worse) one.
> 
> For any set-up where you can expect to fly somewhere where two nodes line up on one side of the drone (typical for a >8 node volume), I can definitely see this helping.

You're right! Also, this might not be needed if the measurement model and noise model in the Kalman filter would be better.

> Should we add some way to accept at least one pair? Right now it can theoretically reject all pairs.
> 

It's a good point. If the Kalman filter would handle Loco measurements better, then this early filtering could be used as a way to find the best pairs (instead of rejecting pairs) and it makes sense to always let pairs through. As of now, the state estimate could be really off if we let bad measurements through, but then rejecting all measurements feels more like a patch than a real solution. Let's discuss!

> I think this needs more evidence before it lives on master. Perhaps through comparing the same log through bindings with this enabled/disabled? What about merging to experimental-show in the meantime?

Sounds good!

tobbeanton:
I might have accidentally merged this to master when I tried to just merge it to my local branch...

PR #1651, opened 2026-07-08
Add bcCam UART service integration

ataffanel:
## Summary
- Add the bcCam UART frame, CRC, and link layer with discovery, credit tracking, and raw frame dispatch.
- Add UART1 ISR callbacks and a bcCam RX collector to route framed UART input from interrupt context.
- Integrate the bcCam UART runtime, firmware and bootloader clients, service state machine, deck controller, and build configuration.
- Update the unit-test helper for Ruby 4 compatibility.

## Testing
- `make unit FILES="test/deck/drivers/src/test_bccam_uart_link_host.c test/deck/drivers/src/test_bccam_uart_rx_collector_host.c test/drivers/src/test_uart1_isr_host.c test/deck/drivers/src/test_bccam_bootloader_uart_client_host.c test/deck/drivers/src/test_bccam_deck_controller_host.c test/deck/drivers/src/test_bccam_firmware_uart_client_host.c test/deck/drivers/src/test_bccam_uart_runtime_host.c test/deck/drivers/src/test_bccam_uart_service_host.c"`
  - 157 tests, 0 failures
- `make -j$(nproc)`
  - cf2 firmware build passed, produced `cf2.bin` and `cf2.hex`


src/deck/drivers/src/bccam_bootloader_uart_client.c
```diff
@@ -0,0 +1,545 @@
...
+      write_len > sizeof(client->write_buffer) - client->write_buffer_used) {
+    return false;
+  }
+
+  memcpy(&client->write_buffer[client->write_buffer_used], buffer, write_len);
+  client->write_buffer_used += write_len;
+  client->bytes_written += write_len;
+
+  const bool is_last =
+    (new_fw_size > 0 && client->bytes_written >= new_fw_size);
```
gemenerik, line 504: `bytes_written` counts how many bytes were accepted, not which bytes were written. If a chunk arrives twice, or two chunks overlap, the count runs ahead of the actual progress and `is_last` fires before the whole image is there. The client tears down the session while the host still has data to send, and the flash can be reported as successful when it isn't.

Could we track coverage instead? Keep how far the image is contiguously filled from `image_base`, reject a write that skips ahead and leaves a gap, and finish when that reaches `image_size`.

The memory docs say writes should go in ascending address order, but that's only a "should", and it doesn't rule out overlapping chunks - 0-3 then 2-5 is ascending and still double-counts. cflib itself resends a chunk on timeout when safelink is off. Better not to depend on clients getting this right.

src/deck/drivers/src/bccam_bootloader_uart_client.c
```diff
@@ -0,0 +1,545 @@
...
+  client->write_buffer_used = 0;
+  client->write_buffer_base = 0;
+  client->bytes_written = 0;
+  client->image_base = 0;
+  client->image_size = 0;
+  client->flash_erased = false;
+  client->image_complete = false;
+}
+
+bool bccam_bootloader_uart_client_write_flash(
```
gemenerik, line 435: I think this fails silently if client fails to set new_fw_size.

src/deck/drivers/src/bccam_uart_service.c
```diff
@@ -0,0 +1,1888 @@
...
+#endif
+
+static bool handle_enter_bootloader_request(void) {
+  if (service_state == BCCAM_UART_SERVICE_STATE_ISP_COMPLETE) {
+    reset_flash_session();
+    service_state = BCCAM_UART_SERVICE_STATE_ISP_ENTERING;
+    return true;
+  }
+
+  if (!is_firmware_state(service_state)) {
```
gemenerik, line 872: If a flash is aborted mid-session (user cancels, client crashes), the service is left in ISP_READY with stale flash_erased/bytes_written in the bootloader client, and nothing times it out. A retried ENTER_BOOTLOADER is dropped here, but ISP_READY is exactly the state flash_write_allowed() accepts writes in, so the host's new image streams in against the old session's state: erase skipped, bytes_written carried over, so the image latches complete early.

tobbeanton:
The coding style differs from ours. Should be move towards this? To do this I think a discussion is needed first.

src/drivers/src/uart1_isr.h
```diff
@@ -0,0 +1,55 @@
...
+         ((status & USART_FLAG_FE) != 0) ||
+         ((status & USART_FLAG_PE) != 0);
+}
+
+static inline uart1IsrRxAction_t uart1IsrGetRxAction(uint32_t status) {
+  const bool has_rx_data = (status & USART_FLAG_RXNE) != 0;
+  const bool has_error = uart1IsrStatusHasError(status);
+
+  if (has_error && has_rx_data) {
+    return UART1_ISR_RX_DISCARD_BYTE_AND_REPORT_ERROR;
```
gemenerik, line 43: If we hit an ORE, this discards the previous, intact byte sitting in DR. Per RM0090 an overrun destroys the incoming byte in the shift register, not the one already in DR. So we drop two adjacent bytes instead of one, which in the worst case straddles a frame boundary and costs us the preceding frame too. We can treat it as a special case and have a `UART1_ISR_RX_DELIVER_BYTE_AND_REPORT_ERROR`.

src/deck/drivers/src/bccam_uart_service.c
```diff
@@ -0,0 +1,1888 @@
...
+  if (!is_rx_firmware_state(service_state)) {
+    return false;
+  }
+
+  const int result =
+    bccam_firmware_uart_client_on_rx_fault(&firmware_client, fault);
+  if (result != BCCAM_UART_OK ||
+      bccam_uart_runtime_get_state(&firmware_client.runtime) ==
+        BCCAM_UART_LINK_FAULT) {
+    enter_fw_resetting();
```
gemenerik, line 1236: If I understand correctly, this is triggered for CRC error. Would be nicer if we could use something like BCCAM_UART_LINK_OP_REINIT_REQUEST, rather than having a fw reset.

Also this has no backoff, so this could start a boot loop of the deck.

src/deck/drivers/src/bccam_uart_link.c
```diff
@@ -0,0 +1,1325 @@
...
+                       default_services,
+                       (uint8_t)(sizeof(default_services) / sizeof(default_services[0])));
+        return BCCAM_UART_OK;
+
+    case BCCAM_UART_LINK_OP_DISCOVER_REPLY:
+        if (endpoint->role != BCCAM_UART_LINK_ROLE_INITIATOR ||
+            endpoint->state != BCCAM_UART_LINK_UNINITIALIZED ||
+            !endpoint->transaction_in_flight ||
+            message.transaction_id != endpoint->pending_transaction_id) {
+            return management_fault(endpoint);
```
gemenerik, line 836: A second DISCOVER_REPLY can't match (the first one cleared the pending transaction at 847-848), so it faults the link. Can we think of any scenario where we may hit the timeout and send a second DISCOVER, and then receive both replies? Even if unreachable today, is it safe to keep?

src/deck/drivers/src/bccam_uart_link.c
```diff
@@ -0,0 +1,1325 @@
...
+        return BCCAM_UART_ERR_LINK_FAULT;
+    }
+
+    if (endpoint->state != BCCAM_UART_LINK_ACTIVE) {
+        return BCCAM_UART_ERR_NOT_ACTIVE;
+    }
+
+    const int service_index = find_service_index(endpoint, service);
+    if (service_index < 0 || service == BCCAM_UART_SERVICE_LINK_MANAGEMENT ||
+        available_credit > 127u ||
```
gemenerik, line 560: As far as I understand we only advertise up to 1 credit right now, but maybe we should also cap this by max number of slots (1 right now)

src/deck/drivers/src/bccam_uart_service.c
```diff
@@ -0,0 +1,1888 @@
...
+    bccam_firmware_uart_client_poll(&firmware_client, now_ticks);
+
+  bccam_firmware_uart_client_observation_t observation;
+  observe_firmware_client(&observation);
+  if (result != BCCAM_UART_OK) {
+    log_poll_failure(result, &observation);
+  }
+
+  if (observation.link_state == BCCAM_UART_LINK_ACTIVE) {
+    service_state = BCCAM_UART_SERVICE_STATE_FW_ACTIVE;
```
gemenerik, line 1192: This fires as soon as link is up, before control probe completes. FW_ACTIVE will already be set by update_firmware_startup_progress(), so this can be removed?

src/deck/drivers/src/bccam_uart_service.c
```diff
@@ -0,0 +1,1888 @@
...
+  if (observation.startup_result !=
+      BCCAM_UART_FIRMWARE_STARTUP_INCOMPATIBLE) {
+    return false;
+  }
+
+  if (phase_message != NULL) {
+    *phase_message = startup_phase_message();
+  }
+
+  service_state = BCCAM_UART_SERVICE_STATE_FW_INCOMPATIBLE;
```
gemenerik, line 796: Should this transition also run disable_firmware_rx_callbacks() + reset_rx_queue_state() like the other terminal transitions (enter_fw_resetting, suspend_firmware_link)?

src/deck/drivers/src/bccam_firmware_uart_client.c
```diff
@@ -0,0 +1,450 @@
...
+#endif
+
+static void firmware_drain_rx(uint32_t timeout_ms) {
+#if defined(UNIT_TEST) || defined(UNIT_TEST_MODE)
+  (void)timeout_ms;
+  test_trace_append_entry(BCCAM_FIRMWARE_UART_CLIENT_TEST_UART_DRAIN);
+#else
+  uint8_t dummy;
+  const TickType_t end = xTaskGetTickCount() + M2T(timeout_ms);
+  while (xTaskGetTickCount() < end) {
```
gemenerik, line 158: We could make this wraparound safe if we want. In a few other places as well.

src/deck/drivers/src/bccam_uart_service.c
```diff
@@ -0,0 +1,1888 @@
...
+#define DEBUG_PRINT(...)
+#endif
+
+#define BCCAM_FW_STARTUP_TIMEOUT_MS 3000
+#define BCCAM_UART_REQUEST_QUEUE_LENGTH 4
+#define BCCAM_UART_RX_QUEUE_LENGTH 8
+#define BCCAM_UART_RX_FRAME_POOL_LENGTH 4
+#define BCCAM_UART_RX_DRAIN_BUDGET BCCAM_UART_RX_QUEUE_LENGTH
+#define BCCAM_UART_TASK_NAME "bcCamUart"
+#define BCCAM_UART_TASK_STACKSIZE (configMINIMAL_STACK_SIZE * 4)
```
gemenerik, line 31: To follow fw convention task name/stack size/priority should be defined in `config.h` instead of here

gemenerik (changes requested):
Looking good!

Two comments I didn't know where else to put:
- Do we want to include license headers like in our other files?
- Should we bound write_flash against actual camera flash size? Currently completely unbounded.

PR #1652, opened 2026-07-08
Links to the crazyflie-demos repository

ArisMorgens:
The examples have moved from **crazyflie-lib-python** to **crazyflie-demos**, so someone following the link wouldn't be able to find the examples.


enyanil (approved):
Looks good!

PR #1656, opened 2026-07-17
Support Ruby 4 unit test lookup

ataffanel:
## Summary

Update the unit-test helper to use `File.exist?` instead of the removed `File.exists?` alias.

This restores unit-test source lookup with Ruby 4 while preserving behavior on supported older Ruby versions.

## Validation

- Direct Ruby 4 lookup exercise passes for existing and missing files.
- Focused Crazyflie unit-test smoke suite passes: 22/22.


gemenerik (dismissed):

ataffanel:
Note for history: #1657 was accidentally based on this branch and was merged into it, so this PR now contains both the original Ruby 4 unit-test lookup fix and the complete Camera Deck Rev G/versioned UART Link work. See #1657.

PR #1658, opened 2026-07-17
Add a parameter to turn the radio off while charging

nellbru:
**Paired with: bitcraze/crazyflie2-nrf-firmware#123**

Crazyflie drones have to be powered on to charge, so drones sitting on a charging rack answer radio scans exactly like flight-ready ones. This costs time, and a charging drone can unintentionally be sent flight commands. I believe this can't be filtered client-side since the ACK that makes a drone appear in a scan is generated by the nRF51 radio itself, so a charging drone and a flight-ready one are indistinguishable at scan time.

**Fix:**
Add the parameter `disableRadioOnUsb` (persistent, default `0`, set to `1` to enable). The rule is applied by the nRF51, which owns both the radio and the USB detection. This PR only sends the preference, over a new `SYSLINK_RADIO_DISABLE_ON_USB` (`0x0C`), from the param callback.

Before the fix:
charging drone -> detected in a scan and answers like a flight-ready drone

After the fix (with both the STM32 and the nRF51 flashed):
charging drone -> same as a powered-off drone, and radio back on unplug without a reboot

Inert on its own: nothing acts on `SYSLINK_RADIO_DISABLE_ON_USB` until the nRF51 PR lands, so the two can be merged in any order.

**Also in this PR:**
The first commit fixes `persistentParamFromStorage()` calling `paramSet()` without `paramNotifyChanged()`, so a param declared with `PARAM_ADD_WITH_CALLBACK` was only notified on a client write, never when its value was restored at boot. Without it, a Crazyflie rebooting while plugged in keeps the value but never tells the nRF51.

### DIFF
--- a/src/hal/interface/radiolink.h
+++ b/src/hal/interface/radiolink.h
@@ -54,6 +54,7 @@ void radiolinkSetChannel(uint8_t channel);
 void radiolinkSetDatarate(uint8_t datarate);
 void radiolinkSetAddress(uint64_t address);
 void radiolinkSetPowerDbm(int8_t powerDbm);
+void radiolinkSendDisableOnUsb(void);
 void radiolinkSyslinkDispatch(SyslinkPacket *slp);
 struct crtpLinkOperations * radiolinkGetLink();
 bool radiolinkSendP2PPacketBroadcast(P2PPacket *p2pp);
--- a/src/hal/interface/syslink.h
+++ b/src/hal/interface/syslink.h
@@ -51,6 +51,7 @@
 #define SYSLINK_RADIO_P2P_ACK       0x09
 #define SYSLINK_RADIO_P2P_BROADCAST 0x0A
 #define SYSLINK_RADIO_READY         0x0B
+#define SYSLINK_RADIO_DISABLE_ON_USB 0x0C
 
 #define SYSLINK_PM_GROUP              0x10
 #define SYSLINK_PM_SOURCE             0x10
--- a/src/hal/src/radiolink.c
+++ b/src/hal/src/radiolink.c
@@ -44,6 +44,7 @@
 #include "queuemonitor.h"
 #include "static_mem.h"
 #include "cfassert.h"
+#include "param.h"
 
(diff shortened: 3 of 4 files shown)


nellbru:
Hello, this PR is ready for review. Can a maintainer have a look? I'm pretty sure everything is good, but I'm waiting for your feedback before I flash all of my drones.

gemenerik:
Thanks for the PR!

We need to discuss this internally, but my first thought is that if the main problem is the drone responding to flight commands while being fixed to a usb cable (a safety concern), the fix is not in removing visibility but should be sought in the supervisor. An opt-in transition condition could be added to getting into the ready-to-fly state. That works over any link, doesn't depend on the nRF firmware version, and fails safe rather than making the drone disappear.

Could you say a bit more about what costs time? I'm not sure if you mean the scanning itself, telling the charging drones from the flight-ready ones, or the charging.

In the meantime you could check `pm.state` for scanned drones if you do not want to have custom fw flashed. 

Let's treat the missing `paramNotifyChanged()` as a bug on its own. Could you split it into a separate PR? Easier to review on its own, and it doesn't need to wait for our discussion here.

nellbru:
Thank you for your reply, and sorry for only replying now.

I've split the missing `paramNotifyChanged()` into its own PR (#1678) as you suggested.

Regarding what costs time, I meant the per drone connection. My script opens a link to each available drone. Charging drones sit mixed in with the ready ones, and there is no way to tell them apart without connecting and reading `pm.state`, so I pay a connection for every drone on every test run, which adds up for a large fleet.

That is also why the supervisor idea, although a good fit for the safety side, does not help since it keeps the radio on, so my script still opens a link to every drone including the charging ones, which is exactly the cost I am trying to avoid. I might be wrong, but turning the radio off seems to be the only option where charging drones drop out at the radio level, so I never pay for them.

So I see these as two different things. The supervisor is the right call for safety, since a charging drone should not fly, while this feature is aimed at scan and connection time at scale.

gemenerik:
Thanks for the clarification! We'll discuss internally in a couple of weeks.

PR #1662, opened 2026-07-27
Fallback default for CONFIG_DECK_BACKEND_DECKCTRL_MAX_DECKS

ArisMorgens:
`CONFIG_DECK_BACKEND_DECKCTRL_MAX_DECKS` depends on `DECK_BACKEND_DECKCTRL` in Kconfig. The new `test_deck_backend_deckctrl.c` unit test compiles `deck_backend_deckctrl.c` directly (bypassing the Kbuild obj-y gate), which broke CI. This PR adds a fallback for `CONFIG_DECK_BACKEND_DECKCTRL_MAX_DECKS`


gemenerik:
While this does unblock the build, it puts a second copy of the 4 from Kconfig into firmware source, with nothing keeping the two in sync.

The dependency belongs on the test instead. `test_deck_backend_deckctrl.c` needs to declare that it requires CONFIG_DECK_BACKEND_DECKCTRL, so it's skipped in configs where the backend is off. Closing; I'll fix it there.

PR #1665, opened 2026-07-30
Fix qqmul() operand order

gemenerik:
qqmul() returned the Hamilton product `p*q` while documenting `q*p`. Correct the implementation to match the documented behaviour and add unit tests that pin the operand order.

mm_pose.c relied on the reversal: it passed its operands in the opposite order and so computed the correct body-frame attitude residual anyway. Its operands are swapped to match the corrected primitive, leaving its behaviour unchanged.

controller_brescianini.c is deliberately left untouched. The controller was imported in ead23f7a with its own quaternion multiply, which took its operands in the order qqmul() documents, and was converted to qqmul() in 4a43abd9 with the argument order carried over as-is. Its five call sites have been running with the operands the other way around ever since, so correcting qqmul() changes that controller's behaviour. 

Working through the call sites points the same way:

  - L307 composed pitch/roll with yaw in the wrong order, rotating the desired thrust direction by the yaw setpoint.
  - L310/L315 produced a world-frame attitude error that was fed to control_omega as body rates, rotating it again by the current yaw.
- L337/L349 are in the reduced/full mixing branch and I have not worked out what the change does there. That branch only runs with mixing_factor strictly between 0 and 1; the default is 1.0

Together these rotated the commanded horizontal acceleration by twice the vehicle yaw, so the controller was correct only near yaw 0 and diverged as heading grew. Both are now correct. This is verified analytically only; the controller has not been flown since the change.


test/modules/src/test_math3d.c
```diff
@@ -0,0 +1,100 @@
+// File under test math3d.h
+#include "math3d.h"
+
+#include <math.h>
+
+#include "unity.h"
+
```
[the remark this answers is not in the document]
gemenerik: The runner generator handles this

test/modules/src/test_math3d.c
```diff
@@ -0,0 +1,100 @@
...
+  TEST_ASSERT_FLOAT_WITHIN(1e-5f, expected.y, actual.y);
+  TEST_ASSERT_FLOAT_WITHIN(1e-5f, expected.z, actual.z);
+  TEST_ASSERT_FLOAT_WITHIN(1e-5f, expected.w, actual.w);
+}
+
+void testThatQqmulReturnsTheHamiltonProductInArgumentOrder() {
+  // Fixture: 90 deg about x, then 90 deg about y. Chosen because these do not
+  // commute, so a reversed implementation cannot coincidentally pass.
+  struct quat q = qaxisangle(mkvec(1, 0, 0), M_PI_2);
+  struct quat p = qaxisangle(mkvec(0, 1, 0), M_PI_2);
```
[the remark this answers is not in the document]
gemenerik: Fixed in e1dc8f2ad29e6e9e1b2d047b55453b9394fa04c5

gemenerik:
Note that math3d.h is exposed through the Python bindings (`%include "math3d.h"` in bindings/cffirmware.i), so qqmul() is part of the public API and out-of-tree users may be calling it. Same for the Brescianini controller, which simulators run through these bindings.

src/modules/src/kalman_core/mm_pose.c
```diff
@@ -40,7 +40,10 @@ void kalmanCoreUpdateWithPose(kalmanCoreData_t* this, poseMeasurement_t *pose)
   // compute orientation error
   struct quat const q_ekf = mkquat(this->q[1], this->q[2], this->q[3], this->q[0]);
   struct quat const q_measured = mkquat(pose->quat.x, pose->quat.y, pose->quat.z, pose->quat.w);
-  struct quat const q_residual = qqmul(q_measured, qinv(q_ekf));
+  // The error state D0..D2 is a body-frame perturbation: kalmanCoreFinalize()
```
stefanthorstenson: A little chatty compared to the general commenting style. I would maybe argue for removing these lines.

test/modules/src/test_math3d.c
```diff
@@ -0,0 +1,78 @@
+// File under test math3d.h
+#include "math3d.h"
+
+#include "unity.h"
+
+// These tests pin the operand order of qqmul(): it returns the Hamilton
```
stefanthorstenson: Very focused on the order. Can we dial this focus down in comments?

test/modules/src/test_math3d.c
```diff
@@ -0,0 +1,78 @@
...
+// product q*p, and callers such as mm_pose.c depend on that order to get a
```
stefanthorstenson: Remove reference to callers.

test/modules/src/test_math3d.c
```diff
@@ -0,0 +1,78 @@
...
+
+  // Test
+  struct quat actual = qqmul(q, qinv(q));
+
+  // Assert
+  assertQuatEqual(qeye(), actual);
+}
+
+void testThatQqmulAppliesBodyFramePerturbationOnTheRight() {
+  // Fixture: this is the property mm_pose.c depends on. With q_ekf mapping
```
stefanthorstenson: Remove references to callers.

test/modules/src/test_math3d.c
```diff
@@ -0,0 +1,78 @@
...
+  // body to world, a body-frame perturbation composes on the right, so the
+  // body-frame residual is q_ekf^-1 * q_measured.
+  // q_measured is written out by hand rather than composed with qqmul(), so
+  // that building the fixture and unwinding it cannot cancel a reversed order.
+  struct quat qEkf = qaxisangle(mkvec(0, 0, 1), M_PI_2_F);
```
stefanthorstenson: Variable naming implies who the user is. Ideally make naming more general, but that's not really necessary...

stefanthorstenson (changes requested):
Looks good! Generally, the AI generated documentation is a bit biased towards this exact fix (operand order), so I had some comments on that.

Also, we might want to think about the breaking API change, but we can take it in person.

test/modules/src/test_math3d.c
gemenerik: totally agree, generalized

stefanthorstenson (approved):
Nice!

PR #1666, opened 2026-08-04
CI kbuild-targets softening of CCMRAM test and stricter randconfig runs

stefanthorstenson:
## Part 1 - soften CCMRAM overflow requirement

With recent additions to the firmware, enabling all build options ("allyesconfig") overflows the CCMRAM. This configuration is most likely never going to be used, and it doesn't have to stop merging or releasing. CI always fails at the moment, which introduces a high risk of missing other failures than the CCMRAM.

This change softens the requirement on kbuild allyesconfig, so that CI passes even though CCMRAM overflows.

Changes:
- Create separate kbuild_target CI script.
- Add flag `tolerate-ccmram-overflow`, which checks if a build fails, and if it fails due to CCMRAM overflow, make the test pass anyways.
- For allyesconfig, pass the `tolerate-ccmram-overflow` flag

## Part 2 - stricter randconfig jobs

Also, randconfig gets run every time, with a new seed for each run, making the result random for each CI run. If this fails for a PR, it is probably not connected to the changes made in the PR, and the PR creator can simply rerun CI to make it pass. That's not good.

Note: This change has been included in this PR, as the solution depends on the changes in the PR.

Changes:
- Create separate kbuild-target-random job, which triggers only on schedule and manual runs and runs randconfig build job.
- Derive KCONFIG_SEED (seed for randconfig) from commit SHA, to make seed deterministic. Put in CI.yml, in order to easily change to something else if needed (run_id being a candidate).
- Add parameter RANDOM_BUILD_COUNT to configure number of randconfig build jobs to run, with individual seeds. This can account for the loss of coverage when randconfig is not run on every PR and master commit.


stefanthorstenson:
One could also argue for creating an elf file even though CCMRAM is exceeded, creating memory statistics, and then performing the check.

.github/workflows/CI.yml
```diff
@@ -124,7 +124,6 @@ jobs:
         target:
         - allyesconfig
         - allnoconfig
```
gemenerik, line 126: Could we do something like this so we don't need to have an additional job?:

```
strategy:
  fail-fast: false
  matrix:
    target: ${{ fromJSON(github.event_name == 'schedule'
      && '["allyesconfig","allnoconfig","randconfig"]'
      || '["allyesconfig","allnoconfig"]') }}
```

And perhaps also not only on "schedule" but also on manual run (workflow_dispatch)

gemenerik:
Feels fair enough to decouple randconfig from the PR but the coverage shrinks a lot now especially since the seed is now a pure function of the commit sha. So an idle master retests an identical config.

Overflow path seems to skip check_elf

Should we handle randconfigs that exceed CCRAM?

Is this supposed to replace tools/build/build? Also I think the [toolbelt](https://github.com/bitcraze/toolbelt) looks for that file

tools/build/kbuild_target
stefanthorstenson, line 1: Do we need to create a new file? Check if it makes sense to keep this in build

stefanthorstenson:
> One could also argue for creating an elf file even though CCMRAM is exceeded, creating memory statistics, and then performing the check.

Team discussion: Keep build logic as it is.

stefanthorstenson:
> Feels fair enough to decouple randconfig from the PR but the coverage shrinks a lot now especially since the seed is now a pure function of the commit sha. So an idle master retests an identical config.

Investigate to add structure for setting a number of randconfigs to run

tools/build/kbuild_target
stefanthorstenson: toolbelt reads files dynamically, so for that purpose it is okay to have a new, separate file

.github/workflows/CI.yml
stefanthorstenson: The jobs are for sure very related, but triggers and run commands differ. I would argue for keeping randconfig as a separate job for clarity and ease of maintenance. Separate jobs will make it clearer when the job is run or skipped, and the yaml file will be easier to read.

stefanthorstenson:
> Should we handle randconfigs that exceed CCRAM?

Added build flag `tolerate-ccmram-overflow`, so this can easily be added to the randconfig job if we want it.

> Is this supposed to replace tools/build/build? Also I think the [toolbelt](https://github.com/bitcraze/toolbelt) looks for that file

Put new logic in `build`, and removed the new file `kbuild_target`

stefanthorstenson:
Run randconfig on master commits?

stefanthorstenson:
Verified randconfig loop.

image

image

See run https://github.com/bitcraze/crazyflie-firmware/actions/runs/31704782223 for randconfig job. After this commit, it is only triggered on schedule or manual dispatch.

gemenerik (approved):
Nice!

Issue #1663, opened 2026-07-27
Blank 0W memories not writable with newer firmware versions


vinzenzm:
I created custom decks with DS28E05 on the PCB. With older firmware versions i am able to write the deck memories with the script in crazyflie_demos repo:
https://github.com/bitcraze/crazyflie-demos/tree/main/demos/scripts/cflib/memory/write_ow

However I now upgraded to a newer firmware version and I am now unable to do so. 
When I run the script it cannot find any deck memories to write to. 

I use crazyflie2.1. Latest known version to work: 2025.02. Version it doesn't work with: 2026.04.

My guess is that it broke with version 2025.12, where the deck system was almost entirely rewritten.

ArisMorgens:
Hello @vinzenzm!

I tested both the [read_ow](https://github.com/bitcraze/crazyflie-demos/tree/main/demos/scripts/cflib/memory/read_ow) and the [write_ow](https://github.com/bitcraze/crazyflie-demos/tree/main/demos/scripts/cflib/memory/write_ow) scripts in crazyflie-demos using the latest release firmware (2026.04) and they worked fine for me.

Could you try using the script with just one of the Bitcraze decks attached to the Crazyflie? This way we can rule out the possibility that the issue is related to your custom deck. If this fails, could you send us the terminal output?

vinzenzm:
Hi Aris, 

sorry, I wasn't clear enough in my first post.
Flashing any previously flashed 0-wire memory works with the newer version. (Custom as well as Bitcraze-Decks). 
The issue only arises for a new, never before flashed 0-Wire memory (DS28E05).
I'll have some new PCB's ready soon then I will be able to send the terminal output.

ArisMorgens:
Sorry, I also missed the blank in the title. I opened PR #1669 which might fix this problem. Please give it a try and let me know how it goes.

vinzenzm:
Just had the time to check it out. I was able to flash a freshly soldered 0-wire memory with the PR #1669 firmware. 
Thanks. 

You can close this issue, or wait until the PR is merged.

ArisMorgens:
@vinzenzm Thanks for testing! The PR is merged, so I'm also closing this issue.

PR #1669, opened 2026-08-05
Writing blank OW-memory 

ArisMorgens:
Based on #1663, writing blank OW-memories didn't work with the recent crazyflie-firmware versions. 

Fixed `owBackendGetNextDeck()` by moving `memoryRegisterHandler` right after a successful `owRead()`, before `infoDecode()` is even attempted. Previously, it returned early on decode failure - which always happens for a blank OW-memory, skipping the registration entirely.

I tested the existing [read_ow/write_ow scripts](https://github.com/bitcraze/crazyflie-demos/tree/main/demos/scripts/cflib/memory) for an already flashed OW-memory and everything works as expected. I haven't tested with a blank OW yet.

### DIFF
--- a/src/deck/backends/deck_backend_onewire.c
+++ b/src/deck/backends/deck_backend_onewire.c
@@ -229,6 +229,20 @@ static DeckInfo* owBackendGetNextDeck(void) {
     // Store backend reference
     deckBuffer.discoveryBackend = NULL; // Backend reference will be set after copying in enumeration code
 
+    // Register the memory handler for this OW device now, independently of whether
+    // the deck info below decodes successfully. A blank, never-flashed memory will
+    // always fail infoDecode(), but it must still be reachable over CRTP so that it
+    // can be written to for the first time.
+    ow_memory_handlers[currentDeck] = (MemoryHandlerDef_t) {
+        .type = MEM_TYPE_OW,
+        .getSize = owMemorySize,
+        .read = owMemoryRead,
+        .write = owMemoryWrite,
+        .getSerialNbr = owMemorySerialNbr,
+        .internal_id = currentDeck,
+    };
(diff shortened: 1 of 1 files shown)


src/deck/backends/deck_backend_onewire.c
stefanthorstenson: Maybe update debug print to something along the lines "...has corrupt or blank memory..."

stefanthorstenson (dismissed):
Looks good! Small comment on debug print.

Could be worth analysing if `deck_backend_deckctrl.c` should get the same treatment.

ArisMorgens:
Thanks for checking! Regarding `deck_backend_deckctrl.c`, it's a bit different than the OW backend. 
For OW, `memoryRegisterHandler()` is the only access path to that chip. 
For DeckCtrl, there are two independent handlers over the same flash: the discovery-time one (read-only, gated behind magic/checksum/production-date checks) and the DFU one (`MEM_TYPE_DECKCTRL_DFU`, read+write, registered unconditionally at init in `deckctrl_dfu_memory.c`, covering the full flash). So even when the discovery-time handler is skipped for a deck with invalid/blank config data, the memory is still reachable through the DFU handler. 

So, I don't think we need to touch `deck_backend_deckctrl.c` for now.

src/deck/backends/deck_backend_onewire.c
ArisMorgens: Good point. Fixed in [986617e](https://github.com/bitcraze/crazyflie-firmware/pull/1669/commits/986617ef48475d9e29f3c13c765d015a61ace39b).

stefanthorstenson (approved):
Nice!

ArisMorgens:
In #1663 vinzenzm confirmed that this PR works for writing to blank OW-memories.

stefanthorstenson (approved):

PR #1670, opened 2026-08-05
Precise landing v2

ArisMorgens:
This PR is a continuation of #1618 and introduces a "precise landing" feature, which allows the flight controller to temporarily adjust control gains for improved landing accuracy. Instead of a single straight-line descent, the Crazyflie hovers above the target with temporarily-tuned controller gains to let horizontal error settle out before the final descent.

This works well when there's accurate positioning data (Lighthouse or MoCap), enabling the Crazyflie to react to external disturbances. Good for successfully landing on Crazyflie Brushless charging docks.

Also, this PR adds new high-level CRTP commands for precise landing, with and without explicit yaw.

Currently, the PID and Mellinger controllers support this feature.

An older version of this PR has been tested in the [fishtank demo](https://github.com/bitcraze/crazyflie-firmware-experimental/tree/icra-2026) at ICRA 2026.


stefanthorstenson (changes requested):
To me, this implementation feels more like a proof-of-concept than a ready-to-merge change.

## Analysis

This change introduces a new structure for changing the chosen controller mid-flight. It introduces new interfaces for all controllers. The ExitPreciseLand command introduces code in the planner specific for this functionality, which justification is not obvious and introduces a decrease in abstraction that is not present before on this level (why does the planner have to call a function when plan is finished?). EnterPreciseLand is called on another abstraction layer, however.

The introduced architecture suggests that all controllers need to implement their own version of this use case-specific behaviour. I don't see that all controllers will create their own form of the precise landing. The architecture also suggests that other types of use case-specific behaviour may exist, which would then need their own interface (e.g. "EnterTakeoff", "EnterCloseproximity"). That would make a cluttered interface, and would suggest packaging the new functions more generically.

## Conclusions

To me, the landing controller is a new controller. I think it would make more sense to keep it as a separate, single controller. Customer interest lies before landing is commanded. A single controller will make it more accessible, and easier to maintain and improve. This is just a suggestion, however.

I think the added functionality is interesting and seems to improve the performance, so it is worth pursuing!

I would suggest a face-to-face rework of the architecture of this fix, both how to activate it and how to systemize it in onboard software.

ArisMorgens:
Thanks for the feedback!

At first I thought that this architecture would be clean, but when I started working on it, it started spilling in multiple places of the firmware. 

I agree that the planner should be cleaned up - maybe by having `crtp_commander_high_level.c` to detect completion itself and call exit from there.

Having a separate landing controller would make sense, so the precise landing is detached from the existing controllers. However, in the current implementation, we don't change controllers mid-flight - we just modify the gains of the already used one. This means, that if we decide to create a new landing controller, we need to cleanly support the mid-flight controller switch, which I assume is possible. 

From the user perspective, I think having a new high-level command is straightforward. Let's find a way to smoothly integrate the low-level as well.

PR #1673, opened 2026-08-12
docs: update Python bindings test instructions

Justine0211:
This updates the Python bindings documentation to recommend `make test_python`
instead of installing the generated bindings with `python3 setup.py install --user`.

The `test_python` target already builds the bindings and runs the tests with
`PYTHONPATH=build`, which matches the intended test workflow.

Tested:
- Documentation-only change; checked the edited section in the GitHub editor.

### DIFF
--- a/docs/building-and-flashing/build.md
+++ b/docs/building-and-flashing/build.md
@@ -129,12 +129,18 @@ First make sure that you have [SWIG](https://swig.org/) installed on your system
 
 ```
 $ make cf2_defconfig
-$ make bindings_python
-$ cd build
-$ python3 setup.py install --user
+$ make test_python
 ```
 
+The `test_python` target builds the bindings and runs the Python tests with the
+build directory on `PYTHONPATH`. If you want to import the generated bindings
+manually without installing them, use:
+
+```
+$ PYTHONPATH=build python3 -c "import cffirmware"
+```
 ## Make targets
+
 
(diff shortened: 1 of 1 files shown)


enyanil:
Hi @Justine0211, thank you for your contribution! This is a very nice update, I was just thinking that we mention swig is needed, but could we also take the opportunity to add something about the required python packages (numpy, pytest, pyyaml)? Thank you!

Justine0211:
Thanks for the suggestion! I’ll update the docs to mention swig and the required Python packages (numpy, pytest, pyyaml).

docs/building-and-flashing/build.md
```diff
@@ -129,13 +129,19 @@ First make sure that you have [SWIG](https://swig.org/) installed on your system
...
+$ make test_python
 ```
 
+The `test_python` target builds the bindings and runs the Python tests with the
+build directory on `PYTHONPATH`. If you want to import the generated bindings
+manually without installing them, use:
+
+```
+$ PYTHONPATH=build python3 -c "import cffirmware"
+```
```
ArisMorgens, line 141: A blank line needs to  be added before `## Make targets`.

ArisMorgens (changes requested):
This is a nice update alongside #1677. I tested `make test_python` and it worked well.

PR #1674, opened 2026-08-12
usd: support three-digit log file indexes

### LINKED ISSUE #988: SD card leads to assert error for 101st log file
Hi,

I have observed a high accumulation of these errors today:
SYS: Assert failed at crazyflie-firmware/vendor/FreeRTOS/tasks.c:1856

More precisely, they occurred every time I tried to switch the usd.logging param to 0 (I'm using the "enable on startup" flag). 
At first I thought it's just another assert error with SD cards (see our discussions [here](https://forum.bitcraze.io/viewtopic.php?t=5221)). However, I have then realized that my SD card contains logs from 0 to 99 now. When deleting the file with suffix 99, this error vanishes for one logging procedure, and then arises again for the next.

So in conclusion it seems this error occurs when the system cannot write any 2-digit suffix to the log filename anymore.

Is allowing for only 100 log files (aka using only 2-digit suffixes) a dedicated, desired choice from your side or could this be changed? Either way, I thought having this error documented could be useful, given that the error message is not very straightforwardly pointing to this cause.

Best,
Dominik

Daniiiil1:
## What changed

- extend uSD log file indexes from two to three digits (`000`-`999`)
- carry across every index digit instead of incrementing beyond `9`
- stop logging cleanly when all 1000 filenames are occupied

## Why

After `log99`, the old two-digit incrementer advanced the tens character past `9`, producing an invalid next filename and ultimately an assertion when logging stopped. The new bounded incrementer keeps all suffix characters numeric and handles exhaustion without corrupting the filename.

Closes #988.

## Validation

- `git diff --check`
- `make cf2_defconfig`
- firmware build started successfully, but could not complete locally because `arm-none-eabi-gcc` is not installed in this environment

### DIFF
--- a/src/deck/drivers/src/usddeck.c
+++ b/src/deck/drivers/src/usddeck.c
@@ -94,6 +94,7 @@
 #define MAX_USD_LOG_EVENTS                (20)
 #define FIXED_FREQUENCY_EVENT_ID          (0xFFFF)
 #define FIXED_FREQUENCY_EVENT_NAME        "fixedFrequency"
+#define USD_LOG_FILE_INDEX_DIGITS          (3)
 
 
 /* set to true when graceful shutdown is triggered */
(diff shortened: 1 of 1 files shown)


gemenerik:
Please complete a firmware build locally and test the work on hardware (Crazyflie + Micro SD card deck, with actual logging to a uSD card)

PR #1677, opened 2026-08-15
docs: mention required Python packages for bindings

Justine0211:
Update Python bindings docs to mention SWIG and required Python packages (numpy, pytest, pyyaml).

### DIFF
--- a/docs/building-and-flashing/build.md
+++ b/docs/building-and-flashing/build.md
@@ -125,7 +125,7 @@ Please go to [these instructions](/docs/development/kbuild.md) to learn how to u
 
 There are certain functions, like the high level commander and controllers, that have been wrapped to python bindings. These can be used to easily test these functions on a computer or use it in a simulation.
 
-First make sure that you have [SWIG](https://swig.org/) installed on your system. Then execute the following commands in the terminal
+First make sure that you have [SWIG](https://swig.org/) installed on your system and that the required Python packages are available: `numpy`, `pytest`, and `pyyaml`. Then execute the following commands in the terminal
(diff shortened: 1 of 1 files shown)


docs/building-and-flashing/build.md
```diff
@@ -125,7 +125,7 @@ Please go to [these instructions](/docs/development/kbuild.md) to learn how to u
 
 There are certain functions, like the high level commander and controllers, that have been wrapped to python bindings. These can be used to easily test these functions on a computer or use it in a simulation.
 
-First make sure that you have [SWIG](https://swig.org/) installed on your system. Then execute the following commands in the terminal
+First make sure that you have [SWIG](https://swig.org/) installed on your system and that the required Python packages are available: `numpy`, `pytest`, and `pyyaml`. Then execute the following commands in the terminal
```
ArisMorgens, line 128: You could add `setuptools` as well.

In [bindings/setup.py](https://github.com/bitcraze/crazyflie-firmware/blob/master/bindings/setup.py) `distutils` is used, which was removed from the standard python library in python 3.12.

Installing `setuptools` is a workaround for now, as it has its own copy of `distutils`. The proper fix is to update [bindings/setup.py](https://github.com/bitcraze/crazyflie-firmware/blob/master/bindings/setup.py) to use setuptools imports, which will also require setuptools as a dependency.

ArisMorgens (changes requested):
Really nice update. When the change is made, this PR can be merged.

I'll create a separate PR for updating bindigs/setup.py to use setuptools.

PR #1678, opened 2026-08-17
Fix persistent param callbacks not called at boot

nellbru:
**Split out from #1658 at @gemenerik's request.**

When a param is written from the client, the param system does `paramSet()` then `paramNotifyChanged()`, so the callback runs. When a persistent value is restored at boot, `persistentParamFromStorage()` only calls `paramSet()`, the callback is never run.

So a param declared with `PARAM_ADD_WITH_CALLBACK` is notified on a client write, but not when its stored value is loaded at startup.

**Fix:**
Call `paramNotifyChanged()` after restoring the value, matching the client-write path.

No param is currently both persistent and has a callback, so this is a no-op on the current firmware, it just makes the two paths consistent. #1658 adds the first such param and relies on it.

### DIFF
--- a/src/modules/src/param_logic.c
+++ b/src/modules/src/param_logic.c
@@ -882,6 +882,7 @@ static bool persistentParamFromStorage(const char *key, void *buffer, size_t len
 
   if (PARAM_VARID_IS_VALID(varId) && paramIsPersistent(varId.index)) {
     paramSet(varId.index, buffer);
+    paramNotifyChanged(varId.index);
   }
 
   return true;


src/modules/src/param_logic.c
```diff
@@ -882,6 +882,7 @@ static bool persistentParamFromStorage(const char *key, void *buffer, size_t len
 
   if (PARAM_VARID_IS_VALID(varId) && paramIsPersistent(varId.index)) {
     paramSet(varId.index, buffer);
+    paramNotifyChanged(varId.index);
```
gemenerik, line 885: This runs inside storageMutex. Any callback that touches storage could deadlock. Nastier because this happens before watchdog starts.

You could instead; record the restored indices here, and run `paramNotifyChanged()` from `paramLogicStorageInit` below, or at any later point (see the general review comment).

gemenerik (changes requested):
`paramLogicStorageInit()` runs inside `commInit()` which is before most subsystems have initialized. I see two ways out; either move the notify pass (see other comment) to a later hook after module init, or document on PARAM_ADD_WITH_CALLBACK that callbacks must tolerate being called at boot before their module's init(). Opinions @ataffanel?

ataffanel (commented):
This PR is short-circuiting one of the main assumption of the param subsystem so I do not think this is the right way forward.

For your comment @gemenerik, calling callbacks before `systemStart` would potentially add complexity everywhere in the firmware and break the assumption that everything `systemStart` is only initialization and setup. Callbacks, only happens when the system has started. The main reason is that we cannot guarantee any inter-module init state before the system has started so we cannot implement generic mechanisms at that point.

For the more over-arching architecture part, the *param* subsystem is currently designed so that parameter values are considered stable and initialized at `systemStart`. This means that any change in param values before that is part of the system initialization and should not matter or be notified.

I think a better implementation for this use-case is that the module should wait on systemStart and should read the parameter value at this point to get its initial value. This also means that the flow will be the same if the value is set by stored parameter or by a kConfig flag (for the radio/usb use case, one might want to build a firmware with it enabled by default for example).

PR #1679, opened 2026-08-17
Removed examples that have been moved in crazyflie-demos

ArisMorgens:
The non-basic examples should be hosted in [crazyflie-demos](https://github.com/bitcraze/crazyflie-demos). 
This PR removes some of them that were added in crazyflie-demos in this PR: https://github.com/bitcraze/crazyflie-demos/pull/7


gemenerik:
I think maybe the appchannel one is core API enough to warrant living here instead of the demos repo. What do you think @ataffanel

ataffanel:
> I think maybe the appchannel one is core API enough to warrant living here instead of the demos repo. What do you think @ataffanel

Yes I agree, this is part of the core functionality so if we want to keep examples for core basic functionality there, at least one simple appchannel example should stay.

gemenerik (approved):

PR #1683, opened 2026-08-20
Add rate limiter to Loco measurements

stefanthorstenson:
Loco measurements can arrive at a combined rate of up to 400 Hz. Put a limit on this rate, in order to free up CPU load on the Crazyflie.

- Add a rate limiter for queuing measurement to the Kalman filter, by not queuing measurements within a certain time period since the last queued measurement. The time period is the inverse of the frequency.
- All measurements are still used for clock correction and are put in the stack of measurements
- Add a kconfig parameter for the frequency
- Add a kconfig parameter for related code

## Notes

There is no intelligence to which measurements are discarded.

Microseconds are used for calculations, since milliseconds make for relatively large deviations from requested rate at frequencies larger than 150 Hz. E.g. 150 Hz becomes 167 Hz.

This solution was chosen due to the ease of use and that configuration is easy to understand.

## Verification

tdoaEngine.stFound gives the rate for enqueued TDOA messages.
Compare this value to tdoaEngine.maxRateHz

maxRateHz = 0 (unlimited):

image

maxRateHz = 150:

image

maxRateHz = 75:

image

No analysis has been to see if the filtering is spread evenly across anchors.

### DIFF
--- a/src/deck/drivers/src/Kconfig
+++ b/src/deck/drivers/src/Kconfig
@@ -345,6 +345,14 @@ config DECK_LOCO_LONGER_RANGE
     help
         Note that anchors need to be built with support for this as well
 
+config DECK_LOCO_TDOA_RATE_LIMIT
+    bool "Support rate limiting of TDoA measurements forwarded to the estimator (experimental)"
+    depends on DECK_LOCO
+    default y
+    help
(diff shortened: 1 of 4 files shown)


gemenerik (approved):

tobbeanton (approved):
Looks good!

PR #1689, opened 2026-08-24
Add sourced console service

ataffanel:
## Summary

This PR lets clients discover and enable console output from the Camera Deck through CRTP. It makes Camera Deck startup and runtime messages available for diagnosis under the source name `deck:bcCam`.

The Crazyflie discovers the deck's `bitcraze.console` UART service and forwards its bytes to source-tagged CRTP packets. Clients discover sources through a boot-lifetime catalog and choose which sources to enable. A successful disable response orders earlier queued console packets before the response and prevents new packets for that source until it is enabled again.

The Crazyflie retains accepted Camera Deck frames, including partially forwarded frames, across disable and startup recovery. CRTP congestion holds later output in the deck's spool, subject to its capacity. A radio activity timeout stops log blocks without clearing console packets already in the shared CRTP transmit queue. Queueing a packet does not establish delivery to a client.

The CRTP document describes the client protocol. A separate Camera Deck document explains its UART binding, buffering, and recovery behavior. The `deck:bcCam` path identifies the Camera Deck's console source; the catalog does not identify individual deck instances.

## Verification

On commit `607470a3`:

- Clean `cf2_defconfig` firmware build passed.
- Affected unit suites passed: **84 tests** (console 11, Common Link 31, UART runtime 13, UART service 28, deck memory 1).
- `git diff --check` passed.

An earlier full `make unit UNIT_TEST_STYLE=min` sweep was incomplete: 170 tests passed before compilation stopped in the unchanged lighthouse test on a missing `emptySequence` type and an incompatible UART mock callback type. The full sweep was not rerun on this revision.

Earlier PR hardware testing exercised retained startup output and disable behavior on a Crazyflie over radio channel 22. The accepted-frame recovery fix, queue-retention change, and this revision have not been hardware-tested.

## Memory

Clean `cf2_defconfig` build on this revision:

- Flash: 307084 / 1032192 bytes
- RAM: 102644 / 131072 bytes
- CCM: 55608 / 65536 bytes (9928 bytes free)


docs/functional-areas/crtp/crtp_console.md
gemenerik, line 1: As far as the reader of this doc is concerned there never was a change, so we should clean out things like: "Channel 0 is the existing Crazyflie-local stream.", "retains its legacy format", "it is unchanged"

docs/functional-areas/crtp/crtp_console.md
```diff
@@ -3,31 +3,107 @@ title: Console
...
+Disabling a source does not discard an already accepted upstream frame; it
+withholds upstream credit until the source is enabled and downstream CRTP
+capacity is available. A successful single-source or all-source disable
+response is a transmit-ordering barrier: a sourced packet that was accepted
+before disable is queued ahead of the response, and no sourced packet for the
+disabled source is queued after it until that source is enabled again.
+
+## Channel 3: source TOC
+
+The TOC uses the original 8-bit log/parameter command shape, without V1/V2 in
```
gemenerik, line 66: This reads as "design-history leakage"

docs/functional-areas/crtp/crtp_console.md
```diff
@@ -3,31 +3,107 @@ title: Console
...
+GET_INFO request:  [0x01]
+GET_INFO response: [0x01, source_count, catalog_crc_u32_le]
+
+Error response:    [command, errno]
+```
+
+The source path occupies the remainder of `GET_ITEM`; it is not NUL terminated.
+Paths are non-empty UTF-8 with non-empty colon-separated segments and are
+unique in the catalog, for example `deck:bcCam` or `cf:nRF51`. They are for
+display and filtering, not stable device identity.
```
gemenerik, line 82: From what I understand then, nothing stably links a deck to its console source, and the doc advises against using source paths as identity. That means this design doesn't support e.g. a UI that shows a console next to each detected deck. Acceptable?

src/modules/src/system.c
```diff
@@ -342,6 +342,7 @@ void systemTask(void *arg)
 /* Global system variables */
 void systemStart()
 {
+  consoleSourceFreeze();
```
gemenerik, line 345: Do we ever reach this if we fail self tests? That is; can we get consoles if any self-test fails?

docs/functional-areas/crtp/crtp_console.md
```diff
@@ -3,31 +3,107 @@ title: Console
...
+
+The catalog is frozen for the Crazyflie boot. IDs are contiguous from zero and
+remain assigned if a source becomes temporarily unavailable. The CRC is the
+firmware CRC-32 over each entry in ID order as `[id byte][source path bytes]`.
+Before freeze, a valid `GET_INFO` or `GET_ITEM` request returns `EAGAIN`; clients
+must retry after startup. An unknown item returns `ENOENT`. As on channel 2, an
+empty request is not answered, an unknown command returns `ENOSYS`, and a known
+command with an invalid payload returns `EINVAL`.
+
+## Camera Deck binding
```
gemenerik, line 92: Should we have deck-specific documentation here?

gemenerik: Also includes what reads as notes for future work. Perhaps better to move into an issue.

gemenerik (commented):
When the link drops (unexpectedly), sourced consoles stay enabled, so the bcCam adapter keeps moving deck output from the deck spool into the CRTP TX queue. That data is then lost without a marker:  either flushed by the existing crtpReset() in log.c when log blocks were active, or ignored by lib-rs on reconnect (see comment on crazyflie-lib-rs#66). Would it make sense to stop forwarding while !crtpIsConnected(), so output stays in the deck spool, which the docs describe as the intended loss boundary?

docs/functional-areas/crtp/crtp_console.md
```diff
@@ -3,31 +3,107 @@ title: Console
...
+IDs 0x00 through 0xfe identify entries in the boot-lifetime source catalog.
+Packet boundaries are adapter boundaries only. A Common Link frame larger than
+29 bytes is split across multiple channel-1 packets with the same source ID.
 
- - The output buffer (of 30 bytes) is full
- - A \"newline\" character has to be send (\\n and/or \\r)
- - The flush function has been called
+## Channel 2: runtime control
+
+`SET_ENABLED` is idempotent and lasts until changed or the Crazyflie reboots.
```
stefanthorstenson: idempotent? what's that in plain English?

src/deck/drivers/src/bccam_uart_service.c
```diff
@@ -938,6 +992,11 @@ static bool update_startup_recovery_watchdog(
     return false;
   }
 
+  if (console_diagnostics_active()) {
```
stefanthorstenson, line 995: Could use a justifying comment.

src/modules/src/console.c
```diff
@@ -77,13 +161,216 @@ void consoleInit()
...
+}
+
+static void consoleCrtpCallback(CRTPPacket *packet)
+{
+  if (packet->channel == CONSOLE_CHANNEL_CONTROL) {
+    if (packet->size == 0u) {
+      return;
+    }
+    if (packet->data[0] != CONSOLE_CMD_SET_ENABLED) {
+      consoleSendCommandError(packet, ENOSYS);
```
stefanthorstenson, line 299: ENOSYS is missing in crtp_error_numbers.md

docs/functional-areas/crtp/crtp_console.md
```diff
@@ -3,31 +3,107 @@ title: Console
...
-Crazyflie to a host using the consoleprintf function.
+CRTP port 0 carries local and dynamically bound console byte streams. Console
+content is canonically UTF-8 but transported as binary: code points may span
+packets and invalid UTF-8 may occur. Clients should use an incremental,
+loss-tolerant decoder and must not infer line or record boundaries from CRTP
+packets.
 
-## Communication protocol
+The sourced Console extension on channels 1 through 3 is available from CRTP
+protocol version 13. Channel 0 retains its legacy format.
```
stefanthorstenson, line 13: Add link to crtp_error_numbers.md, since error numbers are used throughout the document

src/modules/src/console.c
```diff
@@ -77,13 +161,216 @@ void consoleInit()
...
+}
+
+static void consoleSendCommandError(CRTPPacket *packet, uint8_t error)
+{
+  packet->size = 2u;
+  packet->data[1] = error;
+  (void)crtpSendPacketBlock(packet);
+}
+
+static void consoleCrtpCallback(CRTPPacket *packet)
```
stefanthorstenson, line 292: Do I understand it correct that a client tells error from success by looking at the packet length? If so, this should be documented, probably in crtp_console.md

docs/functional-areas/crtp/crtp_console.md
```diff
@@ -3,31 +3,107 @@ title: Console
...
+disabled source is queued after it until that source is enabled again.
+
+## Channel 3: source TOC
+
+The TOC uses the original 8-bit log/parameter command shape, without V1/V2 in
+the command names.
+
+```text
+GET_ITEM request:  [0x00, source_id]
+GET_ITEM response: [0x00, source_id, source_path ...]
```
stefanthorstenson, line 71: Add to docs: source_path is less than or equal to 28 bytes in length

docs/functional-areas/crtp/crtp_console.md
```diff
@@ -3,31 +3,107 @@ title: Console
...
+Port 0, channel 1, Crazyflie -> client
++-----------+--------------------+
+| source ID | console bytes      |
++-----------+--------------------+
+    u8            0..29
+```
 
-The current algorithm in the Crazyflie is to send a console packet is any of
-these condition are fulfilled:
+IDs 0x00 through 0xfe identify entries in the boot-lifetime source catalog.
```
stefanthorstenson, line 33: CONSOLE_SOURCE_MAX is 8 at the moment. Should be mentioned in the docs

docs/functional-areas/crtp/crtp_console.md
```diff
@@ -3,31 +3,107 @@ title: Console
...
+```
+
+The source path occupies the remainder of `GET_ITEM`; it is not NUL terminated.
+Paths are non-empty UTF-8 with non-empty colon-separated segments and are
+unique in the catalog, for example `deck:bcCam` or `cf:nRF51`. They are for
+display and filtering, not stable device identity.
+
+The catalog is frozen for the Crazyflie boot. IDs are contiguous from zero and
+remain assigned if a source becomes temporarily unavailable. The CRC is the
+firmware CRC-32 over each entry in ID order as `[id byte][source path bytes]`.
```
stefanthorstenson, line 86: Specify which CRC, to be able to calculate it.

docs/functional-areas/crtp/crtp_console.md
stefanthorstenson: Is there by the way a reason for 8 being max?

docs/functional-areas/crtp/crtp_console.md
ataffanel: I documented deck:bcCam as the Camera Deck’s console source. The catalog still has no deck-instance identifier; I’d leave a general instance mapping for a future use case that needs it. I indeed think this wording was too strong: I do intend to have the name link to the deck but I did not want to lock the semantic right now.

src/modules/src/system.c
ataffanel: We cannot get other console but we can get the Crazyflie console.

A failed Crazyflie self-test prevents the deck service from starting. The Camera Deck MCU’s later startup is separate: its failure does not automatically fail the Crazyflie self-test, and its console can report a problem if the deck brings up its communication service.

I think this model will be used by future decks with MCU: the power-on-self-test of the Crazyflie only proves that the Crazyflie firmware could start, decks have their own that is accessible after the Crazyflie firmware can be trusted (so that it's POST has passed)

docs/functional-areas/crtp/crtp_console.md
ataffanel: Fixed

docs/functional-areas/crtp/crtp_console.md
ataffanel: I agree, I removed unecessary history.

docs/functional-areas/crtp/crtp_console.md
ataffanel: Moved camera deck details to a separate document.

src/modules/src/console.c
```diff
@@ -77,13 +161,216 @@ void consoleInit()
...
+{
+  if (packet->channel == CONSOLE_CHANNEL_CONTROL) {
```
stefanthorstenson, line 294: I would have liked the entire consoleCrtpCallback to be structured more like this:

```
if (channel == 0) // Start with the lowest
  return;
else if (channel == 1)
  ...
else if (channel == 3)
  if data[0] == SET_ENABLED
     // Set source to enabled
  else // Catches non-covered cases
    sendError()
```
It is very hard to read at the moment, it is difficult to understand which case you are reading. Also, I worry that maintenance will be more difficult if we add more cases.

src/modules/src/console.c
```diff
@@ -77,13 +161,216 @@ void consoleInit()
...
+    packet->size = 4u;
+    packet->data[3] = result;
+    (void)crtpSendPacketBlock(packet);
+    return;
+  }
+
+  if (packet->channel != CONSOLE_CHANNEL_TOC || packet->size == 0u) {
+    return;
+  }
+  if (packet->data[0] == CONSOLE_TOC_GET_INFO) {
```
stefanthorstenson, line 331: Example of difficult to read: which channel is this?

stefanthorstenson (changes requested):
Tried to keep my review to public interfaces and docs.

PR #1694, opened 2026-08-26
Color LED deck fade & blink

matejkarasek:
This PR adds 2 effects to the Color LED Deck:
1) fade into the new color
2) blinking

While it was possible to adjust the color of the deck e.g. from Python via parameters, smooth color transitions were difficult to achieve and fast blinking almost impossible due to bandwith limitations (with cflib in Python at least).

This PR implements these effects in the firmware and uses the following parameters:

Fade effect 
- tFade: time to fade into the new color, in seconds. 

Blink effect
- blinkFreq: frequency in Hz. If 0, blinking is disabled.
- blinkDutyMax: portion of the cycle time spent at max intensity
- blinkDutyMin: portion of the cycle time spent at min intensity
- blinkIntMax: maximal intensity
- blinkIntMin: minimal intensity

Tested on the bottom mounted deck only, but should also work on the top mounted one.

### DIFF
--- a/src/deck/drivers/src/color_led_deck.c
+++ b/src/deck/drivers/src/color_led_deck.c
@@ -51,7 +51,12 @@ typedef struct {
   bool isInit;
   uint8_t brightnessCorr;
   uint32_t currentWrgb8888;
+  uint32_t targetWrgb8888;
+  uint32_t initialWrgb8888;
   uint32_t wrgb8888;
+  float tFade;
+  float tFadeTarget;
+  float tFadeRemaining;
   uint8_t deckTemperature;
   uint8_t throttlePercentage;
   uint8_t ledPosition;
@@ -252,6 +257,14 @@ static void colorLedDeckInit(DeckInfo *info, colorLedContext_t *ctx, const char 
     ledDeckRegisterHandler(&colorLedDeckLedHandler);
     handlerRegistered = true;
   }
+
+  ctx->currentWrgb8888 = 0;
+  ctx->targetWrgb8888 = 0;
+  ctx->initialWrgb8888 = 0;
+
+  ctx->tFade = 0.0f;
+  ctx->tFadeTarget = 0.0f;
+  ctx->tFadeRemaining = 0.0f;
 
   ctx->isInit = true;
 }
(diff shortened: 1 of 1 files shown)


stefanthorstenson:
Hi @matejkarasek !

Cool stuff! I would love to see it in reality, not just code ;) But we can wait with that. A few questions before I or we dig into to reviewing the actual code:

1. What is the intended interface? A small use case would be nice.
2. Would it make sense to put this logic on the LED deck, instead of in the CF stm32?

matejkarasek:
Hi @stefanthorstenson , thanks, to your questions

1) The interface is via the parameters framework, similar to the LED ring. The parameters live in the "colorLedTop", "colorLedBottom" groups, so to test this just connect with cflient and adjust the parameters. Their description is in the code and should auto-propagate to the documentation and cfclient hints once merged.

If the code is mergeable in this form (let's see about point 2), I can also write a demo similar to this one:
https://github.com/bitcraze/crazyflie-demos/blob/main/demos/scripts/cflib/led-ring/led_param_set/led_param_set.py

2) I think the parameters need to stay in the Crazyflie firmware. I guess it may be possible to move part of the code responsible for the actual fading/blinking to the deck's STM. Should you find that necessary, I could use help from someone who can port the code there.

matejkarasek:
A demo Python script:
https://github.com/flapper-drones/crazyflie-demos/blob/PR_colorLEDeffects/demos/scripts/cflib/color_led_deck/color_led_effects/color_led_effects.py

A typical use case would be to use this in a swarm/drone show setting for light effects

src/deck/drivers/src/color_led_deck.c
```diff
@@ -51,7 +51,19 @@ typedef struct {
...
+  float tFade;
+  float tFadeTarget;
+  float tFadeRemaining;
+  float blinkFreq;
+  uint8_t blinkDutyMax;
+  uint8_t blinkDutyMin;
+  uint8_t blinkIntMax;
+  uint8_t blinkIntMin;
+  float blinkCycleTime;
+  uint8_t blinkIntensity;
```
stefanthorstenson: blinkIntensity does not have to be part of this struct. It is set in every execution path where it is needed. I would suggest to remove it from the struct to keep the interface clean.

src/deck/drivers/src/color_led_deck.c
```diff
@@ -51,7 +51,19 @@ typedef struct {
...
   uint32_t wrgb8888;
+  float tFade;
+  float tFadeTarget;
+  float tFadeRemaining;
+  float blinkFreq;
+  uint8_t blinkDutyMax;
+  uint8_t blinkDutyMin;
+  uint8_t blinkIntMax;
+  uint8_t blinkIntMin;
+  float blinkCycleTime;
```
stefanthorstenson: Suggest to rename to `blinkPhase` or similar, since it is not a time, but rather a fraction of the cycle completed.

src/deck/drivers/src/color_led_deck.c
```diff
@@ -502,19 +530,69 @@ static void task(void *param) {
...
-            .b = ctx->currentWrgb8888 & 0xFF
-        };
+        } else {
+          ctx->blinkCycleTime = 0.0f;
+          ctx->blinkIntensity = 255;
+        }
 
+        wrgb_t input;
+        if (ctx->tFadeRemaining > 0.0f) {
+          float alpha = ctx->tFadeRemaining / ctx->tFadeTarget;
```
stefanthorstenson: Very unlikely, but let's guard against zero-division

src/deck/drivers/src/color_led_deck.c
```diff
@@ -502,19 +530,69 @@ static void task(void *param) {
...
+          input.w = ctx->blinkIntensity / 255.0f * (alpha * ((ctx->initialWrgb8888 >> 24) & 0xFF) + (1-alpha) * ((ctx->targetWrgb8888 >> 24) & 0xFF));
+          input.r = ctx->blinkIntensity / 255.0f * (alpha * ((ctx->initialWrgb8888 >> 16) & 0xFF) + (1-alpha) * ((ctx->targetWrgb8888 >> 16) & 0xFF));
+          input.g = ctx->blinkIntensity / 255.0f * (alpha * ((ctx->initialWrgb8888 >> 8) & 0xFF) + (1-alpha) * ((ctx->targetWrgb8888 >> 8) & 0xFF));
+          input.b = ctx->blinkIntensity / 255.0f * (alpha * (ctx->initialWrgb8888 & 0xFF) + (1-alpha) * (ctx->targetWrgb8888 & 0xFF));
+          
+          ctx->tFadeRemaining -= loopInterval/1000.0f;
+          
+        } else {
+        
+          ctx->tFadeRemaining = 0.0f;
```
stefanthorstenson: This could be moved to the end of the "true" branch, to avoid the assignment as default and separate the functionality

ArisMorgens:
Hello @matejkarasek!

Based on this PR, I started moving part of the logic on the color LED deck, along with the fade effect.

- [crazyflie-firmware PR](https://github.com/bitcraze/crazyflie-firmware/pull/1709)
- [color-led-deck-firmware PR](https://github.com/bitcraze/color-led-deck-firmware/pull/9)

A limitation of this approach is that we can't have long-running effects on the deck, as they quickly start to drift. Instead, these effects should be hosted on the Crazyflie side. We want to keep the deck driver clean, so the recommended way to add more effects would be to implement them in an app, similar to [these examples](https://github.com/bitcraze/crazyflie-demos/tree/main/demos/firmware).

matejkarasek:
Hi Aris, thanks for your help, nice that the fade is now on the deck itself!

I must say for a drone show use case, I would be in favour of keeping a set of (basic) effects in the firmware. An app is an extra layer that needs to be maintaned separately and managed in the fleet. But I understand your reasons. 

Thanks, I will close this PR

PR #1695, opened 2026-08-26
Geofence

matejkarasek:
This PR adds geofence function to the supervisor, which ensures that:

1) Setpoints can only be set within the geofenced volume
2) If the crazyflie flies (too far) out of the geofence, it gets locked

The geofence function is disabled by default and can be turned on by SUPERVISOR_GEOFENCE_CHECK_ENABLE or via a (persistent) parameter. The geofence volume limits can be set via parameters.

A new state "WarningReturnToGeofence" is introduced when the crazyflie is outside the geofence volume but still within the "geofenceWarningZone" distance from it. This is to allow the crazyflie to overshoot e.g. due to a disturbance. And also for situations when we are in modeVelocity, here we switch to position control until we are back within the geofence.

Flight tested on the crazyflies and Flappers BEFORE the following was introduced, so some testing is still required (am currently out of office, but will test later):
- SUPERVISOR_CB_DECK_FAULT (but a pretty much identical implementation was used to detect Lighthouse deck communication timouts in the flight tests)
- SUPERVISOR_CB_MOTORS_NOT_RESPONDING

### DIFF
--- a/src/modules/interface/supervisor.h
+++ b/src/modules/interface/supervisor.h
@@ -39,9 +39,10 @@ uint16_t supervisorGetInfoBitfield(void);
  *
  * @param sensors        Latest sensor data
  * @param setpoint       Current setpoint
+ * @param state          Latest state estimate
  * @param stabilizerStep Stabilizer step for rate control
  */
-void supervisorUpdate(const sensorData_t *sensors, const setpoint_t* setpoint, stabilizerStep_t stabilizerStep);
+void supervisorUpdate(const sensorData_t *sensors, const setpoint_t *setpoint, const state_t *state, stabilizerStep_t stabilizerStep);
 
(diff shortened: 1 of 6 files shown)


stefanthorstenson:
Hi @matejkarasek !

Nice! Let us know when you have done the testing you want, and we can look into it!

BR
Stefan

matejkarasek:
Hi @stefanthorstenson, thanks!
So the code is thoroughly flight tested, but needed to merge it with some recent updates to make the PR. It should work, but will report back to confirm.

matejkarasek:
Hi again @stefanthorstenson, so there indeed was a bug introduced during the rebase, now fixed and flight tested on cf21bl. Looking forward to your feedback!

PR #1696, opened 2026-08-27
Disable prop and battery tests (Flapper platform)

matejkarasek:
This PR adds 2 defines allowing to disable the prop and battery tests for platforms, where this functionality is not implemented.

The defines are added to the Flapper platform defaults, no other platforms should be affected.


ArisMorgens (approved):
Tested it and works well!

matejkarasek:
Thanks for the quick merge!

PR #1698, opened 2026-08-31
Add zsh and bash tab completion scripts for make

stefanthorstenson:
Add tab completion for make targets for zsh and bash. 

Tab-completion scripts for `make` are available for bash and zsh in `tools/completion/`. They complete build targets (e.g. `menuconfig`, `clean`) as well as the `configs/*_defconfig` files, and only take effect inside a checkout of this repository.

image

image

It has to be manually enabled by sourcing the file in respective rc file. See documentation in `kbuild.md` for details.


ataffanel (approved):
Great QOL improvement!

PR #1699, opened 2026-08-31
Updated release to use softprops/action-gh-release

### LINKED ISSUE #1675: Replace deprecated CI actions create-release and upload-release-asset
## Problem

release.yml still uses actions/create-release@v1 and actions/upload-release-asset@v1, both archived/unmaintained by GitHub with no newer major version to bump to. As a result, every release run shows ##[warning]Node.js 20 is deprecated annotations, since these actions are forced onto Node 24 by the runner but still target Node 20 internally.

## Suggested solution

Replace both actions with [softprops/action-gh-release](https://github.com/softprops/action-gh-release), which covers both release creation and asset upload in a single maintained action, actively supporting current Node runtimes. Will require reworking the release/upload job steps in release.yml and verifying draft-release and multi-platform binary upload behavior still matches current output.

stefanthorstenson:
## Summary

Closes #1675.

`.github/workflows/release.yml` used `actions/create-release@v1` and
`actions/upload-release-asset@v1`, both archived by GitHub with no
newer major version. They run on the Node.js 20 runtime, which GitHub
Actions no longer ships, so every `release` run logged a
`Node.js 20 is deprecated` warning annotation on the `Create Release on
Github` job and on each per-platform `Upload Binaries to release`
matrix job.

This PR replaces both with
[`softprops/action-gh-release`](https://github.com/softprops/action-gh-release)
`v3.0.3` (pinned by commit SHA, matching this file's existing pinning
convention), which creates the release and uploads assets to it in one
maintained action, and removes the `release_url.txt` artifact
upload/download steps that only existed to pass `create_release`'s
`upload_url` output between jobs — `action-gh-release` doesn't need it,
since it can be invoked multiple times against the same tag (creates
on first call, adds files on later calls).

## Why a third-party action

GitHub has not published an in-house successor to `actions/create-release`/
`actions/upload-release-asset`. The alternative is scripting release
creation/upload with the first-party `gh` CLI directly in `run:` steps,
which avoids granting a third-party action `contents: write` at the
cost of more shell and manual idempotency handling.
`softprops/action-gh-release` was chosen instead, checked 2026-08-31:

- Referenced in ~267,000 workflow files on GitHub (`gh api search/code`),
  the de facto standard replacement for the archived `actions/*-release*`
  actions.
- 5,752 stars, 638 forks, not archived, latest release `v3.0.3`
  published the day before this check with a steady release cadence.

## Behavior preserved / changed

- `tag_name: ${{ github.ref }}` is passed on both jobs, matching the
  old actions exactly, so tag/release naming is unchanged.
- Asset renaming to `<platform>-<version>.bin` is preserved by copying
  the built binary to the desired filename before upload (into a new
  `release_assets/` directory — see below), since `action-gh-release`
  uploads under the basename of the given path rather than exposing a
  rename input.
- `draft: true` / `prerelease: true` are now passed explicitly on
  *both* the `release` and `upload` steps. `action-gh-release` updates
  the existing release in place and defaults `draft` to `false`, so
  without repeating it on `upload`, the release was found to go public
  the moment the first matrix job uploaded — before all platforms
  finished. This is a real behavior fix, not a no-op.
- The version-extraction step now sanitizes `/` out of
  `release_version` (`tr '/' '-'`). The old `asset_name:` input was
  just a label, so an embedded `/` (possible when this
  `workflow_dispatch`-only workflow is run against a branch ref
  instead of a tag) was harmless; the new step does a literal `cp` to
  a path built from that string, so an unsanitized `/` was read as a
  directory separator and crashed the copy. No-op for the intended
  tag-dispatch case.
- The renamed binary is now copied into a new `release_assets/`
  directory instead of alongside the build output in `build/`. The
  `Build` step runs inside a root Docker container that leaves `build/`
  root-owned on the runner host; the old workflow only ever *read* from
  `build/`, but the new `cp` step needs to *write* into it, which the
  non-root `runner` user can't do.

## Verification

Verified with four real
`workflow_dispatch` runs against scratch tags on this repo (each
behavior change above was found and fixed via one such run). Final run
([`33489172268`](https://github.com/bitcraze/crazyflie-firmware/actions/runs/33489172268),
tag `test-release-action-v4`) passed every check:

Run: [`33489172268`](https://github.com/bitcraze/crazyflie-firmware/actions/runs/33489172268)
(scratch tag `test-release-action-v4`)
Draft release: https://github.com/bitcraze/crazyflie-firmware/releases/tag/untagged-04c7082a2df3c3ea4bf5

| Test | Result |
| --- | --- |
| No `Node.js 20 is deprecated` annotation on any of the 6 jobs | Pass |
| Release created, tag matches, only one release object | Pass |
| Release stays `draft: true` / `prerelease: true` throughout (`published_at: null`) | Pass |
| Exactly 5 assets, one per platform (`cf2`, `bolt`, `tag`, `flapper`, `cf21bl`) | Pass |
| Assets correctly named `<platform>-test-release-action-v4.bin` | Pass |
| Assets non-empty, plausible size (297,540–306,656 bytes) | Pass |
| No `Save/Load Release URL File` plumbing steps in any job | Pass |
| Both jobs request only `contents: write` | Pass |
| Idempotency: re-running a matrix job updates its asset in place (no duplicate) | Pass |

The scratch tag/release used for the final verification run should be
deleted before or after merge.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -52,7 +52,7 @@ jobs:
       id: get_release_version
       env:
         GITHUB_REF : ${{ github.ref }}
-      run: echo "release_version=${GITHUB_REF/refs\/tags\//}" >> $GITHUB_OUTPUT
+      run: echo "release_version=$(echo "${GITHUB_REF/refs\/tags\//}" | tr '/' '-')" >> $GITHUB_OUTPUT
 
     - name: Rename binary for upload
       run: cp build/${{ matrix.platform }}.bin build/${{ matrix.platform }}-${{ steps.get_release_version.outputs.release_version }}.bin
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -55,10 +55,12 @@ jobs:
       run: echo "release_version=$(echo "${GITHUB_REF/refs\/tags\//}" | tr '/' '-')" >> $GITHUB_OUTPUT
 
     - name: Rename binary for upload
-      run: cp build/${{ matrix.platform }}.bin build/${{ matrix.platform }}-${{ steps.get_release_version.outputs.release_version }}.bin
+      run: |
+        mkdir -p release_assets
+        cp build/${{ matrix.platform }}.bin release_assets/${{ matrix.platform }}-${{ steps.get_release_version.outputs.release_version }}.bin
 
     - name: Upload ${{ matrix.platform }} bin
       uses: softprops/action-gh-release@e598afbe1493e6b1bafb1f389cabb956eab91231 # v3.0.3
       with:
         tag_name: ${{ github.ref }}
-        files: build/${{ matrix.platform }}-${{ steps.get_release_version.outputs.release_version }}.bin
+        files: release_assets/${{ matrix.platform }}-${{ steps.get_release_version.outputs.release_version }}.bin
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -63,4 +63,6 @@ jobs:
       uses: softprops/action-gh-release@e598afbe1493e6b1bafb1f389cabb956eab91231 # v3.0.3
       with:
         tag_name: ${{ github.ref }}
+        draft: true
+        prerelease: true
         files: release_assets/${{ matrix.platform }}-${{ steps.get_release_version.outputs.release_version }}.bin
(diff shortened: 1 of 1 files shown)


ataffanel (approved):
Great work!

PR #1705, opened 2026-09-04
Reject TDoA candidates whose remote packet is older than one 32 bit time stamp wrap

gemenerik:
## Problem

TDoA3 logs from a 16 anchor system showed gross distance difference outliers of 25 to 100 m in about 8% of candidates. For each transmitting anchor the error was always the same step (for instance −93.5 m for anchor 6 and −41.5 m for anchor 2),  sometimes twice that step, and it only appeared when the remote anchor's packet was 67 ms or more old.

## Cause

A TDoA candidate pairs the packet just received from the transmitting anchor (An) with an earlier packet from a remote anchor (Ar). `calcTDoA()` computes two differences:
- `(txAn - rxAr_by_An)`: how long An waited between receiving Ar's packet and transmitting its own, in An's clock. Both time stamps are in An's packet.
- `(rxAn_by_T - rxAr_by_T)`: how long it was between the tag receiving Ar's packet and receiving An's packet, in the tag's clock.

It adds the Ar→An time of flight to the first difference, converts the result to tag ticks by multiplying it with the clock correction (the two clocks run a few ppm apart), and subtracts it from the second. What remains is how much longer An's packet took to reach the tag than Ar's packet, which is the TDoA.

Anchors only put 32 bit time stamps on air, so both differences are truncated to 32 bits and wrap every 2^32 ticks (67 ms). A candidate is only used if An and the tag hold the same latest packet from Ar (matching sequence numbers). That shared packet gets older with every packet from Ar that both miss, for instance due to packet collisions. Depending on the anchors' transmit rate, a single packet missed by both can be enough to make it more than 67 ms old, and both differences then come out 2^32 ticks too small.

If it were not for the clock correction, the two errors would cancel. But (txAn - rxAr_by_An) is multiplied by the clock correction before it is subtracted, so its error becomes 2^32 · clockCorrection ticks instead of 2^32. The two errors then differ by 2^32 · (clockCorrection − 1) ticks, and that is what ends up in the TDoA.

Light travels about 4.69 mm per tick (the tick rate is 499.2 MHz · 128), so one ppm of clock offset gives an error of about 20 m. This matches what the logs show. The step is fixed per transmitting anchor because its clock offset is, two missed wraps give twice the step, and the observed steps correspond to ordinary clock offsets of 0.6 to 4.6 ppm.

The existing checks do not catch this:
- The sequence number check passes, since An and the tag really do hold the same packet.
- The 30 ms remote data validity looks like it should reject old packets, but it limits how old An's report is, not how old Ar's packet is. If An keeps reporting the same packet from Ar, the entry stays valid even when that packet is more than 67 ms old.
- The clock correction check only verifies that An's clock correction is reliable, based on An's own consecutive packets. In this failure the clock correction is correct, and the check never looks at how old Ar's packet is.
- The (new) geometry check and the outlier filter's physical bound remove the largest errors. But a 12 to 25 m error from an anchor with a small clock offset on a 30 m anchor pair stays below the geometry limit (0.85 × 30 m = 25.5 m), and passes when the outlier filter is open.

## Fix

The tag keeps 40 bit time stamps, so `(rxAn_by_T - rxAr_by_T)` only wraps every 17.2 s and can be used to detect this. A candidate is skipped, the same way a candidate with bad geometry is skipped, when this difference is one 32 bit wrap minus a margin or more. In practice: when the tag received An's packet 67.2 ms or more after the Ar packet it would be paired with.

The margin is needed because if An's clock runs faster than the tag's, An's count can reach one wrap slightly before the tag's does. The margin, 2^20 ticks (16 µs), is well above the difference caused by clock offset, 1.34 µs at the 20 ppm clock offset limit in `clockCorrectionEngine.c`.

The check is applied in both `matchRandomAnchor()` and `matchYoungestAnchor()`. The rejection rate is logged as `tdoaEngine.stRxWrap`, next to `stGeometry`.

## Verification
- Unit tests for the check: a difference of one wrap is rejected, a difference just below the limit is accepted, the difference is still correct when the tag's 40 bit counter rolls over between the two receptions, and a wrapped candidate is skipped in favor of the next one.
- On a Crazyflie at rest in a Loco system, `stRxWrap` fires at about 2 per second out of 148 received packets per second, with the position estimate steady.
- In the logged flight, every candidate with an error of 25 to 100 m had a remote packet older than one wrap, so the check removes that whole band. Wrap errors can also be smaller (about 12 m for the anchor with the smallest clock offset) and are removed too, but below 25 m there are also outliers with other causes, which this PR does not address.

### DIFF
--- a/src/utils/src/tdoa/tdoaEngine.c
+++ b/src/utils/src/tdoa/tdoaEngine.c
@@ -136,6 +136,22 @@ static double calcDistanceDiff(const tdoaAnchorContext_t* otherAnchorCtx, const 
   return SPEED_OF_LIGHT * tdoa / locodeckTsFreq;
 }
 
+// Anchors only put 32 bit time stamps on air, so the anchor side gap (txAn - rxAr_by_An) in calcTDoA()
+// is only valid if the anchor transmitted less than one 32 bit wrap (67 ms) after it received the
+// remote anchor's packet. If the remote packet is older than that (the remote anchor went silent,
+// for instance due to packet collisions), the truncated gaps on the tag and anchor side no longer
+// cancel once the anchor side is scaled by the clock correction. The resulting error is
+// 2^32 * (clockCorrection - 1) ticks, that is about 20 m per ppm of clock offset.
+//
+// The tag has 40 bit time stamps, so the tag side gap is unambiguous and can be used to detect
+// the condition. The margin accounts for the difference in flight times seen by the tag and the anchor.
+#define TDOA_ENGINE_REMOTE_RX_MAX_AGE_MARGIN 0x00100000
+static bool isRemoteRxTimeWithinAnchorTsRange(const tdoaAnchorContext_t* otherAnchorCtx, const int64_t rxAn_by_T_in_cl_T) {
+  const int64_t rxAr_by_T_in_cl_T = tdoaStorageGetRxTime(otherAnchorCtx);
+  const uint64_t tagSideGap = tdoaEngineTruncateToTagTimeStamp(rxAn_by_T_in_cl_T - rxAr_by_T_in_cl_T);
+  return tagSideGap < (TDOA_ENGINE_TRUNCATE_TO_ANCHOR_TS_BITMAP - TDOA_ENGINE_REMOTE_RX_MAX_AGE_MARGIN);
+}
+
 static float sq(float a) { return a * a; }
 
 static float distanceBetweenAnchorsSquared(const point_t* a, const point_t* b) {
@@ -197,6 +213,11 @@ TESTABLE_STATIC bool matchRandomAnchor(tdoaEngineState_t* engineState,
     if (!doExcludeId || (excludedId != candidateAnchorId)) {
       if (tdoaStorageGetCreateAnchorCtx(engineState->anchorInfoArray, candidateAnchorId, now_ms, otherAnchorCtx)) {
         if (engineState->matching.seqNr[index] == tdoaStorageGetSeqNr(otherAnchorCtx) && tdoaStorageGetRemoteTimeOfFlight(anchorCtx, candidateAnchorId)) {
+          if (!isRemoteRxTimeWithinAnchorTsRange(otherAnchorCtx, rxAn_by_T_in_cl_T)) {
+            STATS_CNT_RATE_EVENT(&engineState->stats.remoteRxWrapRejected);
+            continue;
+          }
+
           const double candidateDistanceDiff = calcDistanceDiff(otherAnchorCtx, anchorCtx, txAn_in_cl_An, rxAn_by_T_in_cl_T, locodeckTsFreq);
           if (!isGeometryGoodEnough(anchorCtx, otherAnchorCtx, candidateDistanceDiff)) {
             STATS_CNT_RATE_EVENT(&engineState->stats.geometryRejected);
@@ -238,6 +259,11 @@ static bool matchYoungestAnchor(tdoaEngineState_t* engineState,
             uint32_t updateTime = tdoaStorageGetLastUpdateTime(otherAnchorCtx);
             if (updateTime > youngestUpdateTime) {
               if (engineState->matching.seqNr[index] == tdoaStorageGetSeqNr(otherAnchorCtx)) {
+                if (!isRemoteRxTimeWithinAnchorTsRange(otherAnchorCtx, rxAn_by_T_in_cl_T)) {
+                  STATS_CNT_RATE_EVENT(&engineState->stats.remoteRxWrapRejected);
+                  continue;
+                }
+
                 youngestUpdateTime = updateTime;
                 bestId = candidateAnchorId;
               }
--- a/src/modules/src/tdoaEngineInstance.c
+++ b/src/modules/src/tdoaEngineInstance.c
@@ -67,6 +67,12 @@ STATS_CNT_RATE_LOG_ADD(stFound, &tdoaEngineState.stats.suitableDataFound)
 STATS_CNT_RATE_LOG_ADD(stGeometry, &tdoaEngineState.stats.geometryRejected)
 
 /**
+ * @brief Rate of measurement anchor pair candidates that were rejected because the remote
+ * anchor's packet is older than one 32 bit time stamp wrap (67 ms) [1/s]
+ */
+STATS_CNT_RATE_LOG_ADD(stRxWrap, &tdoaEngineState.stats.remoteRxWrapRejected)
+
+/**
  * @brief Rate of packets where the time stamp is used to update the clock correction factor for an anchor [1/s]
  */
 STATS_CNT_RATE_LOG_ADD(stCc, &tdoaEngineState.stats.clockCorrectionCount)
--- a/src/utils/interface/tdoa/tdoaEngine.h
+++ b/src/utils/interface/tdoa/tdoaEngine.h
@@ -48,4 +48,10 @@ static inline uint64_t tdoaEngineTruncateToAnchorTimeStamp(uint64_t fullTimeStam
   return fullTimeStamp & TDOA_ENGINE_TRUNCATE_TO_ANCHOR_TS_BITMAP;
 }
 
+// The tag (DW1000) keeps 40 bit time stamps
+#define TDOA_ENGINE_TRUNCATE_TO_TAG_TS_BITMAP 0xFFFFFFFFFF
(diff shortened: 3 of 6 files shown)


tobbeanton (dismissed):
The PR looks good and I'm counting on that it has been tested and verified 😉

stefanthorstenson:
I logged the innovation error (which is calculated before the Loco outlier filter), before and after fix.

Before wrap fix (fig 1):
image

After wrap fix (fig 2):
image

Frequency of wrap rejections, `tdoaEngine.stRxWrap` (fig 3):
image

Fig 1 and fig 2 show that the amount of large outliers (more than 5 m) are less. The effect is most likely greater in large spaces which have larger distances between the anchors (which can lead to larger plausible errors).

Fig 3 shows that outliers are now caught and filtered out earlier in the chain (not visible in the innovation plots), which is good! The worst outliers in this test were probably discarded by the geometry check - it could have been interesting to try without it.

src/utils/src/tdoa/tdoaEngine.c
```diff
@@ -136,6 +136,22 @@ static double calcDistanceDiff(const tdoaAnchorContext_t* otherAnchorCtx, const
...
+// cancel once the anchor side is scaled by the clock correction. The resulting error is
+// 2^32 * (clockCorrection - 1) ticks, that is about 20 m per ppm of clock offset.
+//
+// The tag has 40 bit time stamps, so the tag side gap is unambiguous and can be used to detect
+// the condition. The margin accounts for the difference in flight times seen by the tag and the anchor.
+#define TDOA_ENGINE_REMOTE_RX_MAX_AGE_MARGIN 0x00100000
+static bool isRemoteRxTimeWithinAnchorTsRange(const tdoaAnchorContext_t* otherAnchorCtx, const int64_t rxAn_by_T_in_cl_T) {
+  const int64_t rxAr_by_T_in_cl_T = tdoaStorageGetRxTime(otherAnchorCtx);
+  const uint64_t tagSideGap = tdoaEngineTruncateToTagTimeStamp(rxAn_by_T_in_cl_T - rxAr_by_T_in_cl_T);
+  return tagSideGap < (TDOA_ENGINE_TRUNCATE_TO_ANCHOR_TS_BITMAP - TDOA_ENGINE_REMOTE_RX_MAX_AGE_MARGIN);
```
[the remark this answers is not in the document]
stefanthorstenson: Very unlikely to happen

stefanthorstenson: I don't understand the naming of TDOA_ENGINE_REMOTE_RX_MAX_AGE_MARGIN

src/utils/src/tdoa/tdoaEngine.c
```diff
@@ -136,6 +136,22 @@ static double calcDistanceDiff(const tdoaAnchorContext_t* otherAnchorCtx, const
   return SPEED_OF_LIGHT * tdoa / locodeckTsFreq;
 }
 
+// Anchors only put 32 bit time stamps on air, so the anchor side gap (txAn - rxAr_by_An) in calcTDoA()
+// is only valid if the anchor transmitted less than one 32 bit wrap (67 ms) after it received the
+// remote anchor's packet. If the remote packet is older than that (the remote anchor went silent,
+// for instance due to packet collisions), the truncated gaps on the tag and anchor side no longer
+// cancel once the anchor side is scaled by the clock correction. The resulting error is
+// 2^32 * (clockCorrection - 1) ticks, that is about 20 m per ppm of clock offset.
```
stefanthorstenson: What does clockCorrection have to do with it?

stefanthorstenson (commented):
Great! I just have a few questions, for my understanding. We can take it in person.

src/utils/src/tdoa/tdoaEngine.c
stefanthorstenson: Now I understand :smile:

stefanthorstenson:
As discussed, we might want to improve documentation either in the PR or code. Otherwise, it is good to merge from my view! Really awesome finding!

stefanthorstenson:
It seems likely that we can make the same fix in the anchor software - 40-bit internal counter with a 32-bit representation broadcast. So that we only populate data from anchor B if the data is younger than 67 ms, to avoid ambiguity.

It still makes sense to have this fix on the Crazyflie side though, since the issue will exist in "in the future old" anchors without this fix.

Created issue https://github.com/bitcraze/lps-node-firmware/issues/74

stefanthorstenson:
Instead of rejecting measurements where the remote counter has wrapped, we could calculate the number of wraps and use that to get a correct measurement. Either by adding the wrapped time to the remote time, or by applying the clock correction a correct amount of times (e.g. 2 times if the timer has wrapped 1 time).

However, this depends on if https://github.com/bitcraze/lps-node-firmware/issues/74 gets fixed or not. Also, it might not be interesting to use measurements older than 67 ms.

stefanthorstenson (dismissed):

stefanthorstenson (dismissed):
Nice! Great fix!

src/utils/src/tdoa/tdoaEngine.c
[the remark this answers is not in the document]
gemenerik: Very unlikely, as @stefanthorstenson also noted. For this to happen, both the anchor and the tag would have to miss every packet from the remote anchor for 17.2 s, while the anchor keeps reporting that 17 second old packet. I think anchors stop reporting remote anchors they haven't heard for a while, which would make this impossible. The hybrid mode code in this repo only reports anchors heard within the last 2 s, but I'd have to check the anchor firmware to be sure.

Either way it isn't a regression. Before this PR such a candidate was used as well, together with every candidate older than one 32 bit wrap. This PR only removes cases, it doesn't add any. The comment above the check now mentions that the tag's time stamps wrap every 17.2 s.

src/utils/src/tdoa/tdoaEngine.c
```diff
@@ -238,6 +259,11 @@ static bool matchYoungestAnchor(tdoaEngineState_t* engineState,
             uint32_t updateTime = tdoaStorageGetLastUpdateTime(otherAnchorCtx);
             if (updateTime > youngestUpdateTime) {
               if (engineState->matching.seqNr[index] == tdoaStorageGetSeqNr(otherAnchorCtx)) {
+                if (!isRemoteRxTimeWithinAnchorTsRange(otherAnchorCtx, rxAn_by_T_in_cl_T)) {
+                  STATS_CNT_RATE_EVENT(&engineState->stats.remoteRxWrapRejected);
+                  continue;
```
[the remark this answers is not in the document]
gemenerik: Good point, added testThatAWrappedCandidateIsSkippedInFavorOfAnOlderCandidateByTheYoungestMatcher in 64b9d57ac2ad3515e2437a267294ee1c16668daa. matchYoungestAnchor() is now TESTABLE_STATIC so it can be called from the test.

src/utils/src/tdoa/tdoaEngine.c
gemenerik: The name referred to a maximum age that didn't exist as its own constant. In 522f2f3d it's split into two:
  - TDOA_ENGINE_MAX_REMOTE_RX_AGE: the limit. A candidate is rejected if the tag received the remote anchor's packet this long or longer before the current packet, which is one 32 bit wrap (67 ms) minus the margin.
  - TDOA_ENGINE_REMOTE_RX_AGE_MARGIN: 16 µs. If the anchor's clock runs faster than the tag's, the anchor's time stamps can wrap slightly before the tag's do, and the margin makes sure those candidates are still caught.
  
The comment above the check now explains this as well.

src/utils/src/tdoa/tdoaEngine.c
gemenerik: Thanks for sticking with it, the PR text and the comment above the check didn't make this easy to follow. Both are rewritten now.

gemenerik:
Thanks for the reviews. The explanation wasn't good enough, so I updated the following:

- **PR description and the comment above the check**: rewritten to explain step by step why the wrap causes an error and why the clock correction matters.
- **Margin**: the comment now explains why it is needed, and the constants are renamed to `TDOA_ENGINE_MAX_REMOTE_RX_AGE` and `TDOA_ENGINE_REMOTE_RX_AGE_MARGIN` (522f2f3d).
- **Tests**: added a test for the check in `matchYoungestAnchor()`, which TDoA2 uses (64b9d57a).

PR #1708, opened 2026-09-09
Fix thrust limits in RPYT commander

ratheron:
The RPYT commander uses some arbitrary upper thrust limit. The user expects the thrust to be linear between PWM_MIN and PWM_MAX (7000, 65535), so there shouldn't be an upper limit. The lower limit was 1000 before, fixed it to 7000. I've fixed the commander accordingly and added some documentation.

### DIFF
--- a/docs/functional-areas/crtp/crtp_commander.md
+++ b/docs/functional-areas/crtp/crtp_commander.md
@@ -39,12 +39,9 @@ cycle, which is not linear in thrust and dependent on battery voltage.
   is disabled, non-zero thrust is ignored until a packet with thrust 0
   has been received. This prevents the motors from starting on a stale
   or unintended set-point.
-* **Lower bound.** Thrust values below `MIN_THRUST` (7000) are treated as
-  zero. The value is `THRUST_MIN / THRUST_MAX * UINT16_MAX` from
-  `platform_defaults_cf2.h` and marks the point below which the measured
-  thrust curves are too noisy to be useful. When
-  `CONFIG_ENABLE_THRUST_BAT_COMPENSATED` is enabled (default), the same limit is
-  applied again per motor in the battery compensation.
+* **Lower bound.** Thrust values below `MIN_THRUST` are treated as zero. 
+  If `CONFIG_ENABLE_THRUST_BAT_COMPENSATED` is enabled (default), the limit 
+  is computed by `THRUST_MIN / THRUST_MAX * UINT16_MAX`.
 * **Upper bound.** There is no clipping in the commander. The per-motor
   output is capped to the PWM range in the power distribution.
 
@@ -63,16 +60,7 @@ interpreted.
 
 ## Example
 
-Using the python library, send thrust 0 once to unlock, then stream
-set-points:
-
-```python
-cf.commander.send_setpoint(0, 0, 0, 0)
-for _ in range(100):
-    cf.commander.send_setpoint(0.0, 0.0, 0.0, 30000)
-    time.sleep(0.01)
-cf.commander.send_setpoint(0, 0, 0, 0)
-```
+**[ramp.py](https://github.com/bitcraze/crazyflie-demos/tree/main/demos/scripts/cflib/motors/ramp)**: Sends low-level roll/pitch/yaw/thrust set-points directly using the Commander class, ramping thrust up and down.
 
 Set-points must be sent continuously, see the watchdog timeouts in the
 [commander framework](/docs/functional-areas/sensor-to-control/commanders_setpoints.md).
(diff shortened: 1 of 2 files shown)


docs/functional-areas/crtp/crtp_commander.md
```diff
@@ -17,7 +18,61 @@ and the values are valid until the next packet is received.
...
+  applied again per motor in the battery compensation.
+* **Upper bound.** There is no clipping in the commander. The per-motor
+  output is capped to the PWM range in the power distribution.
+
+## Flight modes
+
+The `flightmode` parameter group changes how the four fields are
+interpreted.
+
+| Parameter | Effect |
```
stefanthorstenson: Is this not described anywhere else??

docs/functional-areas/crtp/crtp_commander.md
```diff
@@ -17,7 +18,61 @@ and the values are valid until the next packet is received.
...
+| `posSet` | Roll, pitch and thrust become absolute Y, X (m) and Z (thrust / 1000 m) positions, yaw becomes an absolute angle. |
+| `yawMode` | Frame used for roll/pitch in yaw rate mode: 1 plus-mode, 2 x-mode (default). |
+
+## Example
+
+Using the python library, send thrust 0 once to unlock, then stream
+set-points:
+
+```python
+cf.commander.send_setpoint(0, 0, 0, 0)
```
stefanthorstenson: Suggest to remove code and link to an example instead:

```
**[ramp.py](https://github.com/bitcraze/crazyflie-demos/tree/main/demos/scripts/cflib/motors/ramp)**: Sends low-level roll/pitch/yaw/thrust set-points directly using the Commander class, ramping thrust up and down.
```

src/modules/src/crtp_commander_rpyt.c
```diff
@@ -35,8 +35,11 @@
 #include "num.h"
 #include "position_controller.h"
 
-#define MIN_THRUST  1000
-#define MAX_THRUST  60000
+// Thrust below this is treated as zero. 7000 = THRUST_MIN / THRUST_MAX * UINT16_MAX
+// (platform_defaults_cf2.h); below it the measured thrust curves are too noisy.
+// With CONFIG_ENABLE_THRUST_BAT_COMPENSATED, motorsCompensateBatteryVoltage()
+// clips per motor at THRUST_MIN anyway. Literal so it also works without it.
+#define MIN_THRUST  7000
```
stefanthorstenson: 7000 = THRUST_MIN / THRUST_MAX * UINT16_MAX is only true for the default configuration, and only when those macros are defined. We could do:

```
#include "platform_defaults.h"

#ifdef CONFIG_ENABLE_THRUST_BAT_COMPENSATED
  #define MIN_THRUST ((uint16_t)(THRUST_MIN / THRUST_MAX * UINT16_MAX))
#else
  #define MIN_THRUST 7000
#endif
```

That would make the code more dynamic.

Also, we might want to keep 1000 as default, since 7000 is from battery compensation range if I understand it correctly.

src/modules/src/crtp_commander_rpyt.c
```diff
@@ -35,8 +35,11 @@
 #include "num.h"
 #include "position_controller.h"
 
-#define MIN_THRUST  1000
-#define MAX_THRUST  60000
+// Thrust below this is treated as zero. 7000 = THRUST_MIN / THRUST_MAX * UINT16_MAX
```
stefanthorstenson: If we go with the if statement it is clear that the value is due to battery compensation. The battery compensation module should own the reason for the value. We can delete all comments in this case.

docs/functional-areas/crtp/crtp_commander.md
```diff
@@ -17,7 +18,61 @@ and the values are valid until the next packet is received.
...
+motor commanded with value `t` produces `t / UINT16_MAX * THRUST_MAX`
+newtons (see `platform_defaults_cf2.h`), independent of battery voltage.
+Without battery compensation the value is used directly as PWM duty
+cycle, which is not linear in thrust and dependent on battery voltage.
+
+* **Thrust lock.** After connecting, and whenever the commander priority
+  is disabled, non-zero thrust is ignored until a packet with thrust 0
+  has been received. This prevents the motors from starting on a stale
+  or unintended set-point.
+* **Lower bound.** Thrust values below `MIN_THRUST` (7000) are treated as
```
stefanthorstenson: Update documentation according to changes suggested by other comments (if applicable)

stefanthorstenson (commented):
Nice! And finally a documention of `flightmode` 😄 In general the updates look really good. Let's see if we can make it a little more dynamic, and tune down some of the comments - see comments.

ratheron:
Thank you for your fast response! I agree, computing the values when battery compensation is enable makes more sense. In the other case, we can leave it at 1000. I've also adjusted the docs accordingly.

stefanthorstenson:
Awesome! Thanks for the contribution!

A note: with default settings, with battery compensation, request has a lower limit of 7000 which equals around 8 m/s^2. This is almost free fall and a reasonable lower limit.

PR #1709, opened 2026-09-11
Move Brightness Correction and Effects to the Color LED deck

ArisMorgens:
This PR matches: https://github.com/bitcraze/color-led-deck-firmware/pull/9

- Brightness correction moved on the Color LED deck - The driver no longer applies gamma/perceptual
  correction itself; it sends raw colors and lets the Color LED deck correct them.
- **Color fade** - The  `SET_COLOR` command includes both `wrgb8888` and `tfade`, for slowly fading to a specific color.
- **Blink effect** - New `blinkFreq`/`blinkIntMax`/`blinkIntMin`/`blinkDutyMax`/`blinkDutyMin`
  params, sent to the deck as one atomic `SET_BLINK` command whenever any of them change.
- Protocol version bumped to match the deck firmware.


tobbeanton:
This is great but it is also none trivial due to the CF FW dependency that can cause confusion. That is why I'm a bit hesitant about effects. The color correction makes a lot of sense, the color fade makes sense but the blink effect I'm not so sure. This is mainly because we have to bump the protocol version every time we would make a new effect. So why not implement the effect on the STM32? Also running the same effect on bottom and top LED could make them un-synchronized after a while due to drifting clocks on the decks.

ArisMorgens:
What you're saying about bumping the protocol makes sense.

After some testing, long effects like the blinking, start drifting over time. This starts to become visible quiet early (~30 seconds). So, I'll remove it from the deck side and only keep the fade.

Other effects should then be hosted on the Crazylfie side. However, we want to keep the deck driver clean, so an app would be a better place for them. We already have some [examples](https://github.com/bitcraze/crazyflie-demos/tree/main/demos/firmware) for that.

PR #1710, opened 2026-09-14
Apply include-path shadowing to out-of-tree build

stefanthorstenson:
Files that are put in the out-of-tree or app directory `overrides` will take precedence in the app build. This way, the user can override firmware files without touching the base firmware.

Functionality verified on CF2.1.

**How to test**

Copy
`platform_defaults.h`
`platform_defaults_cf2bl.h`

to `examples/app_out_of_tree_controller/overrides` (overrides is a new directory you have to create)

Change `PID_ROLL_KP` to e.g. 4.3

In `examples/app_out_of_tree_controller`, run
`make cf2bl_defconfig`
`make`

Flash to CF and read parameter `pid_attitude.roll_kp`


tobbeanton (approved):
Looks great!

PR #1711, opened 2026-09-14
Make LPS distance ratio limit a kbuild parameter

stefanthorstenson:
Make distance ratio limit a Kbuild parameter.

Set default to 0.98.

image

Add explanation and recommendations to Kbuild help section


tobbeanton (approved):
0.98 makes sense as default!

PR #1713, opened 2026-09-21
Fix for failing web CI

enyanil:
Link in docs broke the web CI, this should fix the issue with internal linking.


ArisMorgens (approved):
Looks good!

